PR #25, opened 2017-03-09
[WIP] Add Faster RCNN

yuyu2172:
### TODOs

- [x]  `bboxes` -> `bbox`
- [x] Adapt to `bbox, label` convention instead of `bbox` with shape (R, 5).
- [x] Better interface for FasterRCNN. Possibly `feature_extractor, rpn, ...`
- [x] image shape of the original image needs to be known to exclude small objects at post processing steps.
- [x] Formulate loss as sum of `rpn_loss` and `frcnn_loss`.
- [x] properly implement data preparation code. Length of the larger edge should not exceed some number (e.g. 1000 pixels)
- [ ] see if `smooth_l1_loss`  can be replaced by the main Chainer implementation.
- [x] remove the constraint that only one bounding box is used per image for training.
- [ ] Tests


yuyu2172:
## Reproducing output of Caffe

1. `bbox_transform_inv` produces different results for CPU and GPU due to the specification of `xp.exp`.
2. Convolutional layers output differently between the two frameworks on CPU mode. On cuDNN mode, the outputs are same
3. The outputs of nms for GPU and CPU in the Caffe implementation are different.

In the following case, the test outputs from the two frameworks are same

+ use `bbox_transform_inv` with numpy arrays only
+ use cuDNN convolutions
+ use CPU nms

mitmul:
I'm fixing my implementation here: https://github.com/mitmul/chainer-faster-rcnn/tree/v2
The training of RegionProposalNetwork seems to be working correctly.

yuyu2172:
@mitmul 

I did not know that you were working on improving the code as well!
Maybe we should collaborate.

Currently, I am trying to reproduce the results produced by the Caffe implementation.
I succeeded in reproducing the result in the test case with very slight modification from the original code.
For the train code, I found some differences from the caffe version.

In particular,

+ The sigma used for loss_bbox should be 1, instead of 3. The sigma for rpn_loss_bbox is fine to be 3.
+ optimizer is MomentumSGD in caffe version
+ no gradient update for layers conv_1 and conv_2

I find this visualization of Caffe Prototxt to be useful way to debug.
http://ethereon.github.io/netscope/#/gist/cb2c8750fed46cd42626483027eae24d


I took a slight glance at your updated code and it seems great. Especially, the tests you wrote for all the modules!

yuyu2172:
> The training of RegionProposalNetwork seems to be working correctly.

Was there a bug?

mitmul:
I re-organized and cleaned-up many parts of the codes because a lot of variable names were confusing and many unused codes were included in the previous one (and also in the official py-faster-rcnn), for example, "inside_weights" seems actually not to be needed anymore. Then I tried the training for only RPN at this time, I hadn't check the separated (not end2end) training could be successfully done with the previous version, though.

In my `v2` branch, I think fundamental processes are the same as the official implementation, but I couldn't get the RPN converged with lambda=10 that is a balancing parameter used in the loss function denoted as Eq. (1) in the journal version of Faster R-CNN paper (v3 in the arxiv page is [TPAMI version](https://arxiv.org/abs/1506.01497)). OTOH, I succeeded to converge RPN with lambda=1 with my v2 branch codes.

And I also baranced the number of fg/bg bboxes in AnchorTargetLayer [here](https://github.com/mitmul/chainer-faster-rcnn/blob/v2/models/anchor_target_layer.py#L167), while the original code makes the rest of bboxes except the positive ones out of batchsize (256) "negative" [here](https://github.com/rbgirshick/py-faster-rcnn/blob/master/lib/rpn/anchor_target_layer.py#L164). According to the paper (v3), the authors say

> we randomly sample 256 anchors in an image to compute the loss function of a mini-batch, where the sampled positive and negative anchors have a ratio of up to 1:1.

in section "3.1.3 Training PRNs". But the original implementation (py-faster-rcnn) seems not to perform this correctly (the numbers of bg/fg are not always equal).

I just briefly read the paper again, so please correct me if I'm wrong. :)

mitmul:
Ah, sorry, the next sentence of the above quoted one is

> If there are fewer than 128 positive samples in an image, we pad the mini-batch with negative ones.

, so they did the right thing in the original codes already. Somehow I succeeded to train RPN by padding the minibatch with -1 instead of 0, so it might not matter a lot.

yuyu2172:
> I re-organized and cleaned-up many parts of the codes because a lot of variable names were confusing and many unused codes were included in the previous one (and also in the official py-faster-rcnn), for example, "inside_weights" seems actually not to be needed anymore.

That is great. I will check your code and borrow some ideas.

> so they did the right thing in the original codes already.

It seems like so. Thanks for looking into the issue in detail.


Unlike your code, mine trains the model using end2end protocol. I decided to do this because of simplicity. It seems that there is negligible performance difference between the end2end model and alternative optimization model.

FYI, in the current code, training seems to work and the trained model outputs reasonably. However, I still need to work on quantitative evaluation of the model.

yuyu2172:
Deprecated by #154 #153

PR #44, opened 2017-03-13
PCA lighting

Hakuyume:
I added a PCA-based data augmentation method.
This is used in AlexNet [NIPS 2012].

### DIFF
--- a/chainercv/transforms/image/pca_lighting.py
+++ b/chainercv/transforms/image/pca_lighting.py
@@ -2,9 +2,12 @@ import numpy
 
 
 def pca_lighting(img, sigma, eigen_value=None, eigen_vector=None):
-    """Alter the intensities of input image using PCA.
+    """AlexNet style color augmentation
 
-    This is used in training of AlexNet [1].
+    This method adds a noise vector drawn from a Gaussian. The direction of
+    the Gaussian is same as that of the principal components of the dataset.
+
+    This method is used in training of AlexNet [1].
 
     .. [1] Alex Krizhevsky, Ilya Sutskever, Geoffrey E. Hinton. \
     ImageNet Classification with Deep Convolutional Neural Networks. \
--- a/chainercv/transforms/image/pca_lighting.py
+++ b/chainercv/transforms/image/pca_lighting.py
@@ -2,10 +2,10 @@ import numpy
 
 
 def pca_lighting(image, sigma, eigen_value=None, eigen_vector=None):
-    """
-    Alter the intensities of input image using PCA.
-    This is used in "ImageNet Classification with Deep Convolutional
-    Neural Networks" [NIPS 2012].
+    """Alter the intensities of input image using PCA.
+
+    This is used in "ImageNet Classification with Deep Convolutional Neural
+    Networks" [NIPS 2012].
 
     Args:
         image (numpy.ndarray): An image array to be augmented. This is in
--- a/chainercv/transforms/image/pca_lighting.py
+++ b/chainercv/transforms/image/pca_lighting.py
@@ -1,7 +1,7 @@
 import numpy
 
 
-def pca_lighting(image, sigma, eigen_value=None, eigen_vector=None):
+def pca_lighting(img, sigma, eigen_value=None, eigen_vector=None):
     """Alter the intensities of input image using PCA.
 
     This is used in "ImageNet Classification with Deep Convolutional Neural
@@ -33,7 +33,7 @@ def pca_lighting(image, sigma, eigen_value=None, eigen_vector=None):
 
     alpha = numpy.random.normal(0, sigma, size=3)
 
-    image = image.copy()
-    image += eigen_vector.dot(eigen_value * alpha).reshape(-1, 1, 1)
+    img = img.copy()
+    img += eigen_vector.dot(eigen_value * alpha).reshape(-1, 1, 1)
 
-    return image
+    return img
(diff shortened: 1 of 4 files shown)


yuyu2172:
Thanks for PR!

Could you point me to the page where you found the default values of `eign_*`?
Also, is there default value for `sigma`?

Hakuyume:
I copied the eigen values and vectors from facebook/fb.resnet.torch.
https://github.com/facebook/fb.resnet.torch/blob/master/datasets/imagenet.lua#L71-L78

In AlexNet, they used 0.1 for `sigma`.
So the default value should be 25.5 if the range of intensities is [0, 255].

tests/transforms_tests/image_tests/test_pca_lighting.py
```diff
@@ -0,0 +1,19 @@
...
+
+class TestPCALighting(unittest.TestCase):
+
+    def test_pca_lighting(self):
+        img = np.random.uniform(size=(3, 48, 32))
+
+        out = pca_lighting(img, 0.1)
+        self.assertEqual(img.shape, out.shape)
+        self.assertEqual(img.dtype, out.dtype)
+
```
yuyu2172: Could you add a test case when alpha is 0?

Hakuyume:
The value of `sigma` depends on the range of intensities.
In order to define the default value, we have to assume the range of intensities ([0, 1) or [0, 255]).

chainercv/transforms/image/pca_lighting.py
```diff
@@ -0,0 +1,39 @@
+import numpy
+
+
+def pca_lighting(image, sigma, eigen_value=None, eigen_vector=None):
+    """Alter the intensities of input image using PCA.
+
+    This is used in "ImageNet Classification with Deep Convolutional Neural
+    Networks" [NIPS 2012].
```
yuyu2172: Could you cite a paper using rst's functionality?

```
    This is used in training of AlexNet [Krizhevsky]_.

    .. [Krizhevsky] Alex Krizhevsky, Ilya Sutskever, Geoffrey E. Hinton. \
    ImageNet Classification with Deep Convolutional Neural Networks. \
    NIPS 2012.
```

chainercv/transforms/image/pca_lighting.py
```diff
@@ -0,0 +1,39 @@
+import numpy
+
+
+def pca_lighting(image, sigma, eigen_value=None, eigen_vector=None):
```
yuyu2172: :%s/image/img

yuyu2172:
> In AlexNet, they used 0.1 for sigma.
So the default value should be 25.5 if the range of intensities is [0, 255].

Could you add that to the documentation?
I get why you could not set the default value.

docs/source/reference/transforms.rst
```diff
@@ -23,6 +23,10 @@ pad
 ~~~
 .. autofunction:: pad
 
+pca_lighting
+~~~
```
yuyu2172:
~~~ should be same length as pca_lighting

chainercv/transforms/image/pca_lighting.py
```diff
@@ -0,0 +1,48 @@
...
+
+def pca_lighting(img, sigma, eigen_value=None, eigen_vector=None):
+    """Alter the intensities of input image using PCA.
+
+    This is used in training of AlexNet [Krizhevsky]_.
+
+    .. [Krizhevsky] Alex Krizhevsky, Ilya Sutskever, Geoffrey E. Hinton. \
+    ImageNet Classification with Deep Convolutional Neural Networks. \
+    NIPS 2012."ImageNet Classification with Deep Convolutional Neural
+    Networks" [NIPS 2012].
```
yuyu2172: `"ImageNet..." [NIPS 2012]. `
can be eliminated.

Just you know, you can check the processed html documentation by `cd docs; make html`.

Hakuyume:
I find Chainer uses a different style of citation.
http://docs.chainer.org/en/stable/reference/initializers.html#chainer.initializers.HeNormal
How do you think of it?

yuyu2172:
>I find Chainer uses a different style of citation.
> http://docs.chainer.org/en/stable/reference/initializers.html#chainer.initializers.HeNormal
> How do you think of it?

Thanks for pointing that out.
I find that there is an inconsistency among docs in Chainer.
For example, 
http://docs.chainer.org/en/v1.21.0/reference/links.html#vgg16layers

I like this one better because all authors are included. And looks a little bit better.

chainercv/transforms/image/pca_lighting.py
```diff
@@ -0,0 +1,47 @@
+import numpy
+
+
+def pca_lighting(img, sigma, eigen_value=None, eigen_vector=None):
+    """Alter the intensities of input image using PCA.
+
```
yuyu2172: This explanation sounded like PCA is computed every time this function is called.

For example,
"perturbate color channels of images in principle directions of the covariance matrix of color channels" 

This is a little bit wordy, so please take this just as a reference.
Maybe it is smart to steal words from the original paper.

Hakuyume: How about "Alter the intensities of color channels based on given principal components" ?
Perhaps, "AlexNet-style color augmentation" is enough because it it well known.

yuyu2172: Make "AlexNet style color augmentation" as the title of docstring, and add a line below that describes what it is (e.g. adding random amount of perturbation in the direction of principle components).

yuyu2172:
LGTM!

PR #54, opened 2017-03-15
Fallback to pillow

Hakuyume:
I modify `resize` function to use `Pillow` if `cv2` is not installed.

### DIFF
--- a/chainercv/transforms/image/resize.py
+++ b/chainercv/transforms/image/resize.py
@@ -1,9 +1,27 @@
+import warnings
+
 try:
     import cv2
-    _available = True
+
+    def _resize(img, size):
+        return cv2.resize(img, dsize=size)
 
 except ImportError:
-    _available = False
+    import numpy
+    import PIL
+
+    warnings.warn(
+        'cv2 is not installed on your environment. '
+        'ChainerCV will fallback on Pillow. ',
+        RuntimeWarning)
+
+    def _resize(img, size):
+        channels = []
+        for i in range(3):
+            ch = PIL.Image.fromarray(img[:, :, i], mode='F')
+            ch = ch.resize(size)
+            channels.append(numpy.array(ch))
+        return numpy.stack(channels, axis=2)
 
 
 def resize(img, output_shape):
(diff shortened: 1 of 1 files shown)


chainercv/transforms/image/resize.py
```diff
@@ -1,9 +1,27 @@
...
+    warnings.warn(
+        'cv2 is not installed on your environment. '
+        'ChainerCV will fallback on Pillow. ',
+        RuntimeWarning)
+
+    def _resize(img, size):
+        channels = []
+        for i in range(3):
+            ch = PIL.Image.fromarray(img[:, :, i], mode='F')
+            ch = ch.resize(size)
```
yuyu2172: Could you add a keyword argument `resample=PIL.Image.BILINEAR`?

chainercv/transforms/image/resize.py
```diff
@@ -21,15 +39,8 @@ def resize(img, output_shape):
         ~numpy.ndarray: A resize array in CHW format.
```
yuyu2172: Please add a documentation about modules used depending on the environment (CV2, PIL).

yuyu2172: Also, please add that cv2 is the faster implementation.

chainercv/transforms/image/resize.py
```diff
@@ -1,9 +1,27 @@
...
+        return cv2.resize(img, dsize=size)
 
 except ImportError:
-    _available = False
+    import numpy
+    import PIL
+
+    warnings.warn(
+        'cv2 is not installed on your environment. '
+        'ChainerCV will fallback on Pillow. ',
```
yuyu2172: Please note that CV2 is faster

yuyu2172:
Thanks.
I benchmarked both implementation and concluded that PIL is slower than cv2.
Do you agree with this?

The code I used for benchmarking is below.

```python
import numpy as np
import time

import cv2
import PIL


def _resize_cv2(img, size):
    return cv2.resize(img, dsize=size)


def _resize_pillow(img, size):
    channels = []
    for i in range(3):
        a = img[:, :, i]
        ch = PIL.Image.fromarray(img[:, :, i], mode='F')
        ch = ch.resize(size)
        channels.append(np.array(ch))
    return np.stack(channels, axis=2)

size = (256, 128)
img = astronaut().astype(np.float32)

for i in range(5):
    _resize_pillow(img, size)

times = []
for i in range(10):
    start = time.time()
    _resize_pillow(img, size)
    times.append(time.time() - start)
print('pillow ', np.mean(times))


for i in range(5):
    _resize_cv2(img, size)

times = []
for i in range(10):
    start = time.time()
    _resize_cv2(img, size)
    times.append(time.time() - start)
print('cv2 ', np.mean(times))

```

chainercv/transforms/image/resize.py
yuyu2172: Also, please add that `img` should be float type.

Hakuyume:
Thank you for your benchmark code.
On my environment, this code returned as follows.
```
pillow  0.00493938922882
cv2  0.00032696723938
```
So, we can say that `cv2` is much faster than `Pillow`.

yuyu2172:
LGTM!

PR #55, opened 2017-03-15
Fix warning message in semantic_semantic_segmentation_vis_report.py

Hakuyume:
Current code shows a warning "scikit-image is not installed", if `_available` is `False`.
https://github.com/pfnet/chainercv/blob/master/chainercv/extensions/semantic_segmentation/semantic_segmentation_vis_report.py#L200-L203

However. there is no `import` for `skimage` and `_available` indicates the availability of `matplotlib`.
This message should be "matplotlib is not installed".


yuyu2172:
Thanks for PR!

I found style inconsistency between `SemanticSegmentatinVisReport` and `DetectionVisReport`.
Since this PR is made to fix typo, I will send another PR.

PR #60, opened 2017-03-16
Use OpenCV3

Hakuyume:
Currently, ChainerCV is tested with OpenCV3.
https://github.com/pfnet/chainercv/blob/master/environment.yml#L8

But the docs tell users to install OpenCV2.
https://github.com/pfnet/chainercv/blob/master/chainercv/transforms/image/resize.py#L37


yuyu2172:
Thanks.
I rebased your PR in the related PR (#61).
I could not realize some problems because I forgot to `make html` at docs when reviewing your previous PR #54.

yuyu2172:
Since this commit is included in #61, I am closing this. I appreciate your contribution a lot!

PR #73, opened 2017-03-18
Support grayscale image in resize

Hakuyume:
`resize` function  with the `Pillow` backend can not treat grayscale image.
I add a test and fix this bug.

### DIFF
--- a/chainercv/transforms/image/resize.py
+++ b/chainercv/transforms/image/resize.py
@@ -18,7 +18,7 @@ except ImportError:
 
     def _resize(img, size):
         channels = []
-        for i in range(3):
+        for i in range(img.shape[2]):
             ch = PIL.Image.fromarray(img[:, :, i], mode='F')
             ch = ch.resize(size, resample=PIL.Image.BILINEAR)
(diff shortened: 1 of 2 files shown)


Hakuyume:
I found `cv2` backend also can not  treat grayscale image.
I fixed this bug.

chainercv/transforms/image/resize.py
```diff
@@ -1,13 +1,17 @@
+import numpy
 import warnings
 
 try:
     import cv2
 
     def _resize(img, size):
-        return cv2.resize(img, dsize=size)
+        img = cv2.resize(img, dsize=size)
+        if len(img.shape) == 3:
```
yuyu2172: Could you add an argument that explains this condition?
I thought it was not intuitive that `cv2` changed shape of img from `H, W, 1` to `H, W`.

So, add something like
"# When channel size of the image is one, the resized image returned by cv2 is two dimensional array.

yuyu2172:
Thanks. Nice PR.

yuyu2172:
LGTM

PR #74, opened 2017-03-18
BBox expansion

Hakuyume:
I added an data augmentation method used in SSD.


yuyu2172:
Thank you for PR!

I think this data augmentation code can be split into two smaller transforms that handles `img` and `bbox` independently. 
For example, `image/random_flip.py` and `bbox/flip_bbox` split transforms into smaller pieces.
Like that example, could you make a transform to handle only one data type?
So, I suggest you to make a transform `image/random_expand`. I would make the function to return randomly selected intermediate variables, so return data would be `img, top, left`.

By the way, I think that this design is good because `random_expand` of images may become useful for applications other than ones with bounding boxes.

On the transform for bbox, I am not sure if we really need a function implemented in this framework or not.

Furthermore, I think `random_expansion` is better because it is the name used in the original paper.

Hakuyume:
I agree it is not good to take both of image and bbox. OK, I will separate into two methods, `random_expand` and `translate_bbox`.

The reason why I use `expand` (not `expansion`) is this method is called as `expand` in the original Caffe implementation. How do you think of it?

Hakuyume:
The current policy can be defined as "Random parameters are determined in image functions. The users have to pass the parameters to bbox functions."
However, some functions, for example random crop in SSD, use bbox information to determine random parameters.

yuyu2172:
> I agree it is not good to take both of image and bbox. OK, I will separate into two methods, random_expand and translate_bbox (is it better to make anothor PR?) .

It is OK to work in this PR.
I think size the code of two transforms will not be too big.

> The reason why I use expand (not expansion) is this method is called as expand in the original Caffe implementation. How do you think of it?

I see.
Thinking twice, I think `expand` is fine.

yuyu2172:
> However, some functions, for example random crop in SSD, use bbox information to determine random parameters.

I think it is fine for a function to take both image and bounding boxes, as long as it only returns either one of them.
In this case, we transform the image using `random_crop_jaccard(img, bbox, min_jaccard, size)` and translate the bounding box according to how the image was cropped.

I think that it becomes more difficult to organize transforms if we allow them to transform multiple types.

Hakuyume:
> In this case, we transform the image using random_crop_jaccard(img, bbox, min_jaccard, size) and translate the bounding box according to how the image was cropped.

When I consider only this case, adding `random_crop_jaccard(bbox, img_shape, min_jaccard)` looks more straightforward way.  However this design is against to the policy I described above. Anyway, this topic should be discussed in the another Issue/PR.

Hakuyume:
I made `image.random_expand` and `bbox.translate_bbox`.

yuyu2172:
> When I consider only this case, adding random_crop_jaccard(bbox, img_shape, min_jaccard) looks more straightforward way. However this design is against to the policy I described above. Anyway, this topic should be discussed in the another Issue/PR.

As you said, we may need to prepare new functions for transforming bounding boxes.
Although this is contrary to what I said before, it started seem more natural to make `random_crop_jiccard(img, bbox, ...)` transform both `img` and `bboxes`. I think so because it is hard to imagine a situation where only `img` needs to be transformed.

I think we can put it under `transforms/image`. In this case, I think we need to soften the policy so such that "any transforms need to individually handle data type except in the case when it is required to take multiple data types for transformations".

Maybe we can continue this discussion in another PR.

chainercv/transforms/bbox/translate_bbox.py
```diff
@@ -0,0 +1,27 @@
+def translate_bbox(bbox, x_offset=0, y_offset=0):
+    """Translate bounding boxes.
+
```
yuyu2172: How about adding a description on the usage?

This is used together with image transforms, such as padding and cropping, which translates the left top point of the image from coordinate :math:`(0, 0)` to coordinate :math:`(x_offset, y_offset)`.

chainercv/transforms/image/random_expand.py
```diff
@@ -0,0 +1,54 @@
+import numpy
+import random
+
+
+def random_expand(img, max_ratio=4, fill=0, return_params=False):
+    """Expand image randomly.
```
yuyu2172: expand an image randomly

tests/transforms_tests/image_tests/test_random_expand.py
```diff
@@ -0,0 +1,50 @@
...
+        self.assertLessEqual(ratio, 4)
+        self.assertEqual(out.shape[1], int(64 * ratio))
+        self.assertEqual(out.shape[2], int(32 * ratio))
+
+        out = random_expand(img, max_ratio=2)
+
+    def test_random_expand_fill(self):
+        img = np.random.uniform(-1, 1, size=(3, 64, 32))
+
+        fills = (128, (104, 117, 123), np.random.uniform(255, size=3))
```
yuyu2172: I prefer using `chainer.testing.parameterize` to test one `fill` value in one test call.
In this way, all test cases will be tested even if some in the beginning fails.

yuyu2172: Examples can be found in Chainer's main repository and `detection_vis_report`'s test.

chainercv/transforms/image/random_expand.py
```diff
@@ -0,0 +1,54 @@
...
+        else:
+            return img
+
+    C, H, W = img.shape
+
+    ratio = random.uniform(1, max_ratio)
+    out_H, out_W = int(H * ratio), int(W * ratio)
+
+    top = random.randint(0, out_H - H)
+    left = random.randint(0, out_W - W)
```
yuyu2172: How about making the variable names consistent with the description and change them to `x_offset, y_offset`.

chainercv/transforms/image/random_expand.py
```diff
@@ -2,44 +2,39 @@
...
 
 
-def random_expand(img, bbox, max_ratio=4, fill=0):
+def random_expand(img, max_ratio=4, fill=0, return_params=False):
     """Expand image randomly.
 
     This method expands the size of image randomly by padding pixels. The
-    aspect ratio of the image is kept. The bounding boxes are move according
-    to the expansion.
+    aspect ratio of the image is kept.
```
yuyu2172: I think borrowing descriptions from the original paper is more intuitive.
It was a little bit difficult to understand the intent of this operation with the current description for people who have never seen this before.


Randomly place an image on a canvas whose edge length is at most :obj:`max_ratio` times larger than the original image. The canvas is filled by a value :obj:`fill` except for the region where the original image is placed.

This data augmentation trick is commonly used to create "zoom out" effect [1].

yuyu2172:
Could you check comments?

Hakuyume:
@yuyu2172 Thank your for comments. I pushed some commits.

yuyu2172:
LGTM!

PR #77, opened 2017-03-20
import numpy as np

Hakuyume:
Always import `numpy` as `np`.


yuyu2172:
Thanks.
I was thinking about this convention.
In Chainer, there are both `numpy` and `np`, but I don't think ChainerCV needs to follow this style.

I personally prefer consistency, so thumbs up for only using `np`.

yuyu2172:
LGTM

PR #94, opened 2017-03-24
Return intermediate values as a dict with argument return_param

### LINKED ISSUE #83: Inconsistency in arguments of transforms/image
Many image transform functions take some kind of argument to decide whether to return intermediate variables or not (e.g. `return_params` in `random_expand`).
These variables are necessary when handling multiple data types. For example, bounding boxes need to be transformed according to the transformation done to images.

The issue is that parameters used to control the behavior is inconsistent among transforms.
I think that the current API is confusing for users.

Examples of (function, parameter) pairs.
+ `random_expand` takes `return_params`
+ `random_crop` takes `return_slices`.
+  `random_flip` takes `return_flips`.

yuyu2172:
Fix #83.

### DIFF
--- a/chainercv/transforms/image/random_expand.py
+++ b/chainercv/transforms/image/random_expand.py
@@ -48,8 +48,7 @@ def random_expand(img, max_ratio=4, fill=0, return_param=False):
 
     if max_ratio <= 1:
         if return_param:
-            # img, ratio, x_offset, y_offset
-            return img, 1, 0, 0
+            return img, {'ratio': 1, 'x_offset': 0, 'y_offset': 0}
         else:
             return img
 
--- a/chainercv/transforms/image/random_expand.py
+++ b/chainercv/transforms/image/random_expand.py
@@ -2,7 +2,7 @@ import numpy as np
 import random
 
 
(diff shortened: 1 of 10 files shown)


chainercv/transforms/image/random_expand.py
```diff
@@ -24,16 +24,30 @@ def random_expand(img, max_ratio=4, fill=0, return_params=False):
...
+            * ('y_offset', *int*): The y coodinate of the top left corner of
+                the image after placing on the canvas.
 
     """
 
     if max_ratio <= 1:
-        if return_params:
+        if return_param:
             # img, ratio, x_offset, y_offset
             return img, 1, 0, 0
```
Hakuyume: `1, 0, 0` -> `{'ratio': 1, 'x_offset': 0, 'y_offset': 0}`

Hakuyume:
I think `ten_crop` should return intermediate values because it is similar to `center_crop`.
(Is it better to make another PR?)

yuyu2172:
@Hakuyume 

Thanks for reviewing the code.

I think modification to `center_crop` should be done in another PR.

PR #100, opened 2017-03-30
Fix citation in GoogLeNet

Hakuyume:
I found a wrong citation in docs of `googlenet`.
http://www.cv-foundation.org/openaccess/content_cvpr_2015/html/Szegedy_Going_Deeper_With_2015_CVPR_paper.html


yuyu2172:
Thanks. The names of the authors were completely wrong.

Issue #102, opened 2017-03-31
Function name of pad


yuyu2172:
Currently, `transforms/image/pad.py` has different interface to ones in NumPy.

The name of the function should be changed in order to avoid confusion.
The function implements a padding of image to match given shape.
It’s the same behavior as css `background-size: contain property`, so the name should be `resize_contain`, for example.

PR #104, opened 2017-03-31
Add resize_contain

yuyu2172:
Fix issue #102.


yuyu2172:
Example

```
import numpy as np
from chainercv import transforms


img = astronaut()
img = img.transpose(2, 0, 1).astype(np.float32)[::-1]
img = transforms.resize_contain(img, (300, 600), bg_value=255)
img = transforms.chw_to_pil_image(img)
# plt.imshow(img)
# plt.show()
```

yuyu2172:
figure_1

PR #105, opened 2017-03-31
Add flip_keypoint

yuyu2172:


chainercv/transforms/keypoint/flip_keypoint.py
```diff
@@ -0,0 +1,28 @@
...
+        size (tuple): A tuple of length 2. The width and the height
+            of the image before resized.
+        x_flip (bool): Modify keypoints according to a horizontal flip of
+            an image.
+        y_flip (bool): Modify keypoints according to a vertical flip of
+            an image.
+
+    Returns:
+        ~numpy.ndarray:
+        Keypoint modified according to image flips.
```
Hakuyume: There is an inconsistency, `Keypoints` and `Keypoint`.

yuyu2172: Thank you for reviewing!
You are right.

PR #109, opened 2017-03-31
Fix inconsistency of fill argument

Hakuyume:
Both of `resize_contain` and `random_expand` have an argument to specify the value of padded pixels.
However, the designs of this argument are different from each other.

- In `resize_contain`
    - The name is `bg_value`
    - It takes only a scalar value
- In `random_expand`
    - The name is `fill`
    - It takes a scalar, a tuple, or an array

I fitted `resize_contain` to `random_expand`.

This PR is based on #108. I will rebase this PR after #108 is merged.

### DIFF
--- a/chainercv/transforms/image/resize_contain.py
+++ b/chainercv/transforms/image/resize_contain.py
@@ -3,7 +3,7 @@ import numpy as np
 from chainercv.transforms import resize
 
 
-def resize_contain(img, size, bg_value=0, return_param=False):
+def resize_contain(img, size, fill=0, return_param=False):
     """Resize the image to fit in the given area while keeping aspect ratio.
 
     If both the width and the height in :obj:`size` are larger than the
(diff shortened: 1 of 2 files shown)


Hakuyume:
I rebased this PR on master.

chainercv/transforms/image/resize_contain.py
```diff
@@ -18,7 +18,7 @@ def resize_contain(img, size, bg_value=0, return_param=False):
             CHW format.
         size (tuple of two ints): A tuple of two elements:
             :obj:`width, height`. The size of the image after resizing.
-        bg_value (scalar): Value of the padded regions.
+        fill (float or tuple or ~numpy.ndarray): The value of padded pixels.
```
yuyu2172: s/float or tuple/ float, tuple

yuyu2172:
Thanks for PR!
I totally agree with your proporsal.

chainercv/transforms/image/resize_contain.py
Hakuyume: Thank you. I fixed it.
I found the same issue in `random_expand` (#111).

yuyu2172:
LGTM

PR #120, opened 2017-04-06
Add MultipleDataset

Hakuyume:
I added `MultipleDataset`. This dataset works as a concatenated dataset. I will add a doc.

### DIFF
--- a/chainercv/datasets/__init__.py
+++ b/chainercv/datasets/__init__.py
@@ -1,5 +1,6 @@
 from chainercv.datasets.cub.cub_keypoint_dataset import CUBKeypointDataset  # NOQA
 from chainercv.datasets.cub.cub_label_dataset import CUBLabelDataset  # NOQA
+from chainercv.datasets.multiple_dataset import MultipleDataset  # NOQA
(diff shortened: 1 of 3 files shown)


yuyu2172:
Thanks for PR!
I rally like the feature you support. However, I think that it is more appropriate to add this dataset to main Chainer repositiory because the functionality is more general to contain in ChainerCV.

FYI, it is perfectly fine to declare a class like this inside a train script when writing a example code.

yuyu2172:
Also, I like `MergeDataset` than `MultipleDataset`. `Merge` conveys a message that multiple datasets are put together into one.

Hakuyume:
I think `TransformDataset` is as general as `MultipleDataset`. It also can be used for general tasks. How do you think of it?
Anyway, as you pointed out, this dataset is not only for CV. I agree that it is better to add this dataset to Chainer. I will close this PR.
I agree that the name is not good. I am thinking to use `MergeDataset` or `ConcatenatedDataset`.

yuyu2172:
I agree with you that `TransformDataset` is general enough to be integrated into Chainer.

As you suggested, I sent PR. https://github.com/pfnet/chainer/pull/2513

PR #121, opened 2017-04-11
Fix image format in vis_bbox

Hakuyume:
`vis_bbox` assumes `img` is `HWC` format even though its example uses `VOCDetectionDataset`, which returns `CHW` style images.


Hakuyume:
Sorry, I was misunderstanding.

PR #122, opened 2017-04-11
Add SSD (Single Shot Multibox Detector)

Hakuyume:
I am implementing SSD (Single Shot Multibox Detector).


Hakuyume:
demo.py requires `scikit-image`.
result of demo:
figure_1

Hakuyume:
I added test code and SSD512. I think model conversion from Caffe and training should be in examples.

Hakuyume:
@yuyu2172 To fix the APIs, could you review my codes? It is better that all detectors have same (similar) APIs.

yuyu2172:
Sure. I will be done with Faster RCNN in few days, and after that I will check your code.

The PR for FasterRCNN is extremely messy with different PRs rebased place to place.
I need to clean that up...

yuyu2172:
Just a few comments.

+ Following #144, I think return values should be tuple of lists.
+ `predict`  shouldn't take NMS paramters. This is because there can be Detection models that do not use NMS. These parameters should be attributes of the model.
+ Relating to the above comment, the default NMS parameters should be optimized for visual quality (e.g. score_thresh = 0.6) instead of mAP.
+ Can you stop using private classes?
+ I think Caffe conversion code does not need to be in repository. This is because some models may fail to reproduce performance using Caffe model because its run-time behavior is changed for consistency reasons in this framework. I think that as long as pretrained model created by ChainerCV produces almost same or better results than the originally reported results, we can forget about Caffemodel.

yuyu2172:
How about adding `test_ssd.py` that evaluates mAP of a pretraiend model?

Hakuyume:
> Following #144, I think return values should be tuple of lists.

You are right. I wll fix it.
> predict shouldn't take NMS paramters. This is because there can be Detection models that do not use NMS. These parameters should be attributes of the model.

You are right. I will move them to attributes.
> Relating to the above comment, the default NMS parameters should be optimized for visual quality (e.g. score_thresh = 0.6) instead of mAP.

Why do you prefer threshold for visual quality? The reason why I chose current setting is only that I was more interested in mAP.
>  Can you stop using private classes?

Do you mean `_Normalize` or `_SSDVGG16`? I think no one want to use them directly because they are optimized to SSD.  Why do you want to make them public?. 
> I think Caffe conversion code does not need to be in the repository. This is because some models may fail to reproduce performance using Caffe model because its run-time behavior is changed for consistency reasons in this framework. I think that as long as pretrained models created by ChainerCV produce almost same or better results than the originally reported results, we can forget about Caffemodel.

For SSD, the conversion from Caffemodel works well. This is a big advantage. Why do you want to hide the model compatibility of ChainerCV's SSD? I think it is better to provide conversion code in examples.  Even in this case, we can provide the model trained on ChainerCV.
Of course, we can't provide conversion codes if we can't reproduce the performance. In this case, we can simply provide only the models trained on ChainerCV.

Hakuyume:
> How about adding test_ssd.py that evaluates mAP of a pretraiend model?
Nice idea! You mean adding an example which evaluates a pre-trained model, right? (`test_ssd.py` already exists in `tests/link_tests`. But I guess you mean a different thing).

Hakuyume:
I updated SSD for new APIs #144.

yuyu2172:
> Do you mean _Normalize or _SSDVGG16? I think no one want to use them directly because they are optimized to SSD. Why do you want to make them public?.

+ First of all, I think that it is beneficial to have `_Normalize` to be public, since it can be useful if someone wants to make their extension of SSD.
+ Second, I would encourage more reusable class design for SSDVGG16. It is only extendable to `SSD300` and `SSD512` in the current implementation, and not to other feature extractors.
I talked to you about this before, and I know that you have tried, but maybe we can do this. In that situation, this class should be public.
+ Third, I think that Chainer has loose definition of private classes. `links/models/vision/resnet.py` contains many public classes that have little documentation and tests. I think we can skip tests for helper classes like those since we do not have very much time.

> For SSD, the conversion from Caffemodel works well. This is a big advantage.

Okay. Let's add caffemodel converter. It would be nice if you add that and a README that explains where to download caffemodel.
Eventually, I would like that README to contain a procedures to
1. run test without caffemodel
2. run test with caffemodel and how to convert caffemodel
3. quick demo
4. train


> You mean adding an example which evaluates a pre-trained model,

Yes! I think this makes others to easily re-evaluate your model.

yuyu2172:
I think putting all files under `chainercv.links.ssd` is better because there will be multiple files relating to SSD including losses for SSD, model and possibly other feature extractors.

yuyu2172:
Adding to a note on private class, I am not familiar with the case when a public class inherits from a private class. Is this acceptable?

Hakuyume:
> Second, I would encourage more reusable class design for SSDVGG16. It is only extendable to SSD300 and SSD512 in the current implementation, and not to other feature extractors.
I talked to you about this before, and I know that you have tried, but maybe we can do this. In that situation, this class should be public.

I will rethink the design of classes and APIs. Adding a base class which doesn't depend on any feature extractor looks good to me. 

> I think putting all files under chainercv.links.ssd is better

We can decide this after all code are implemented.

> Adding to a note on private class, I am not familiar with the case when a public class inherits from a private class. Is this acceptable?

I don't feel this strange. But if this is not Python-style, I will obey it.

yuyu2172:
> I don't feel this strange. But if this is not Python-style, I will obey it

In my understanding, inheritance should preserve accessibility of classes.
In some static languages, this is not compilable.

Hakuyume:
> In my understanding, inheritance should preserve accessibility of classes.
In some static languages, this is not compilable.

Is it true? I don't know much about static languages. However, in C++, we can write as follows.
``` 
class Outer {
private:
  class PrivateInner {};

public:
  class PublicInner : PrivateInner {};
};

int main() {
  // this is ok
  Outer::PublicInner inner;

  // complie error
  // Outer::PrivateInner inner;

  return 0;
}
```

yuyu2172:
Actually, I did not know that there are languages that do not allow this in the world before this discussion.
Apparently, C# does not allow it.

http://stackoverflow.com/questions/10858107/c-sharp-private-hidden-base-class

Hakuyume:
The reason why I keep classes/methods private is that Making private things public is much easier than making public things private. If we make something public once, we have to keep it because of the backward compatibility.

yuyu2172:
Right. I am not against the concept of private something in general. Just you know.

yuyu2172:
I forgot to comment on an issue.

> Why do you prefer threshold for visual quality? The reason why I chose current setting is only that I was more interested in mAP.

I think that the model should work out of the box with the default paramters.
Most casual and likely usage of this model would be not to evaluate the performance rigidly using mAP, but just visualize some outputs for fun.
In that case, it would be better to optimize for visual quality.

Hakuyume:
> Most casual and likely usage of this model would be not to evaluate the performance rigidly using mAP, but just visualize some outputs for fun.
OK, I will set the default threshold to the value optimized for visual quality.

Hakuyume:
> I will rethink the design of classes and APIs. Adding a base class which doesn't depend on any feature extractor looks good to me.

I re-designed classes. How about this? (docs and tests are under development and I will add trainer in another PR)

yuyu2172:
> I re-designed classes. How do you think of it?

I like this more!

In the current implementation, you make SSD300 to inherit in two levels.
I personally like less inheritance.
For instance you can make SSD to take a feature extractor that outputs list of variables by `__call__` as argument to `__init__`.
And then, make SSD300 that inherits from SSD and instantiate a feature extractor.

The feature extractor I am talking about is something like this.
In the class like below, feature extractors can be easily swapped to other conv nets like ResNet.
I am using https://arxiv.org/abs/1611.10012 as a reference on possible extensions.

```python
class SSDVGG16Feature(chainer.Chain):

    def __init__(self):
        super(SSDVGG16Feature, self).__init__(
            conv1_1=L.Convolution2D(None, 64, 3, pad=1, **self.conv_init),
            conv1_2=L.Convolution2D(None, 64, 3, pad=1, **self.conv_init),

            conv2_1=L.Convolution2D(None, 128, 3, pad=1, **self.conv_init),
            conv2_2=L.Convolution2D(None, 128, 3, pad=1, **self.conv_init),

            conv3_1=L.Convolution2D(None, 256, 3, pad=1, **self.conv_init),
            conv3_2=L.Convolution2D(None, 256, 3, pad=1, **self.conv_init),
            conv3_3=L.Convolution2D(None, 256, 3, pad=1, **self.conv_init),

            conv4_1=L.Convolution2D(None, 512, 3, pad=1, **self.conv_init),
            conv4_2=L.Convolution2D(None, 512, 3, pad=1, **self.conv_init),
            conv4_3=L.Convolution2D(None, 512, 3, pad=1, **self.conv_init),
            norm4=Normalize(512, initial=initializers.Constant(20)),

            conv5_1=L.DilatedConvolution2D(
                None, 512, 3, pad=1, **self.conv_init),
            conv5_2=L.DilatedConvolution2D(
                None, 512, 3, pad=1, **self.conv_init),
            conv5_3=L.DilatedConvolution2D(
                None, 512, 3, pad=1, **self.conv_init),

            conv6=L.DilatedConvolution2D(
                None, 1024, 3, pad=6, dilate=6, **self.conv_init),
            conv7=L.Convolution2D(None, 1024, 1, **self.conv_init),
        )

    def __call__(self, x):
        ys = list()

        h = F.relu(self.conv1_1(x))
        h = F.relu(self.conv1_2(h))
        h = F.max_pooling_2d(h, 2)

        h = F.relu(self.conv2_1(h))
        h = F.relu(self.conv2_2(h))
        h = F.max_pooling_2d(h, 2)

        h = F.relu(self.conv3_1(h))
        h = F.relu(self.conv3_2(h))
        h = F.relu(self.conv3_3(h))
        h = F.max_pooling_2d(h, 2)

        h = F.relu(self.conv4_1(h))
        h = F.relu(self.conv4_2(h))
        h = F.relu(self.conv4_3(h))
        ys.append(self.norm4(h))
        h = F.max_pooling_2d(h, 2)

        h = F.relu(self.conv5_1(h))
        h = F.relu(self.conv5_2(h))
        h = F.relu(self.conv5_3(h))
        h = F.max_pooling_2d(h, 3, stride=1, pad=1)

        h = F.relu(self.conv6(h))
        h = F.relu(self.conv7(h))
        ys.append(h)

        return ys


class SSD300Head(chainer.Chain):

    # it would be nice if we can create a class that can generate both SSD300Head and SSD500Head

    insize = 300
    grids = (38, 19, 10, 5, 3, 1)
    aspect_ratios = ((2,), (2, 3), (2, 3), (2, 3), (2,), (2,))
    steps = [s / 300 for s in (8, 16, 32, 64, 100, 300)]
    sizes = [s / 300 for s in (30, 60, 111, 162, 213, 264, 315)]

    def __init__(self, n_class, feature_base):
        super(SSD300Head, self).__init__(
            conv8_1=L.Convolution2D(None, 256, 1, **self.conv_init),
            conv8_2=L.Convolution2D(
                None, 512, 3, stride=2, pad=1, **self.conv_init),

            conv9_1=L.Convolution2D(None, 128, 1, **self.conv_init),
            conv9_2=L.Convolution2D(
                None, 256, 3, stride=2, pad=1, **self.conv_init),

            conv10_1=L.Convolution2D(None, 128, 1, **self.conv_init),
            conv10_2=L.Convolution2D(None, 256, 3, **self.conv_init),

            conv11_1=L.Convolution2D(None, 128, 1, **self.conv_init),
            conv11_2=L.Convolution2D(None, 256, 3, **self.conv_init),
        )

    def __call__(self, x):
        for i in range(8, 11 + 1):
            h = ys[-1]
            h = F.relu(self['conv{:d}_1'.format(i)](h))
            h = F.relu(self['conv{:d}_2'.format(i)](h))
            ys.append(h)
        return ys


class SSD(chainer.Chain):

    def __init__(self, feature, head):
        super(SSD, self).__init__(
            feature=feature,
            head=head)

    def __call__(self, x):
        hs = self.feature(x)
        xs = self.head(hs)
        ys_loc = list()
        ys_conf = list()
        for i, x in enumerate(xs):
            loc = self.loc[i](x)
            loc = F.transpose(loc, (0, 2, 3, 1))
            loc = F.reshape(loc, (loc.shape[0], -1, 4))
            ys_loc.append(loc)

            conf = self.conf[i](x)
            conf = F.transpose(conf, (0, 2, 3, 1))
            conf = F.reshape(
                conf, (conf.shape[0], -1, self.n_class + 1))
            ys_conf.append(conf)

        y_loc = F.concat(ys_loc, axis=1)
        y_conf = F.concat(ys_conf, axis=1)

        return y_loc, y_conf


class SSD300(SSD):

    def __init__(self):
        feature = SSDVGG16Feature(...)
        head = SSD300Head(...)
        super(SSD300, self).__init__(
            feature, head)


```

chainercv/links/ssd/ssd_vgg16.py
```diff
@@ -0,0 +1,218 @@
+import numpy as np
```
yuyu2172: You forgot `from __future import division`

chainercv/links/ssd/ssd.py
```diff
@@ -0,0 +1,257 @@
...
+            conf = F.reshape(
+                conf, (conf.shape[0], -1, self.n_class + 1))
+            ys_conf.append(conf)
+
+        y_loc = F.concat(ys_loc, axis=1)
+        y_conf = F.concat(ys_conf, axis=1)
+
+        return y_loc, y_conf
+
+    def __call__(self, x):
```
yuyu2172: A documentation on `__call__` would be nice.

yuyu2172:
Similar to Chainer's  `VGG.__call__` method, it would be nice if that function returns intermediate variables by a dictionary.

I think it is not necessary to return all intermediate variables like Chainer's, but it would be nice if it can return some important features upon requests, such as outputs of the head and feature.

chainercv/links/ssd/ssd.py
Hakuyume: Do you know how to show docs of `__call__` in sphinx? `:members:` didn't show them.

yuyu2172: Like this
https://github.com/yuyu2172/chainercv/blob/d4f0188db8296efca6ec67e6d3f3de40f8c78e8e/docs/source/reference/links.rst

yuyu2172:
Here is my ongoing work on `__call__` for Faster RCNN (https://github.com/yuyu2172/chainercv/blob/95909de82359dec97cdfd4f0f6096f9e34c2dd87/chainercv/links/faster_rcnn/faster_rcnn.py#L45).

Hakuyume:
> I personally like less inheritance.

SSD300 and SSD512 share not only the base feature extractor but also `prepare`. Actually, I had tried to separate the feature detector into another class. But finally, I found this two-level inheritance is most simple.

> Similar to Chainer's VGG.__call__ method, it would be nice if that function returns intermediate variables by a dictionary.

All important features are returned by `feature()` method. We can access these values with an index. I don't want to assume all feature extractors have named feature maps (someone may implement feature extractor using `ChainLink`).

yuyu2172:
> SSD300 and SSD512 share not only the base feature extractor but also prepare. Actually, I had tried to separate the feature detector into another class. But finally, I found this two-level inheritance is most simple.

Why do you think that `prepare` can not be implemented at SSD?

Hakuyume:
I think different feature extractors may require different preparation. For VGG-16, it consists of resizing and subtracting mean. But I don't think this is general.

yuyu2172:
> All important features are returned by feature() method. We can access these values with an index. I don't want to assume all feature extractors have named feature maps (someone may implement feature extractor using ChainLink).

I think usage of ChainLink does not stop us from using dictionaries as output.
If the return outputs for ChainLink can only be a list in that case.

```
{'norm4': norm4_var, 'conv7': conv7_var, 'chain_link': [var_1, ..., var_n]}
```

yuyu2172:
> I think different feature extractors may require different preparation. For VGG-16, it consists of resizing and subtracting mean. But I don't think this is general.

Well, if limit the scope to SSD, we can assume that images are resized to some fixed size.
Also, subtracting mean is quite common.

Those parameters can be passed to `SSD` class when instantiated.
```python

class SSD
    def __init__(self, feature, ...):
         self.mean = feature.mean
```

Hakuyume:
For SSD, most important thing is the order of feature maps. `feature()` should return a list (or tuple) of feature maps.

yuyu2172:
> For SSD, most important thing is the order of feature maps. feature() should return a list (or tuple) of feature maps.

I see. OrderedDict can work in that case.

Hakuyume:
> Well, if limit the scope to SSD, we can assume that images are resized to some fixed size.
Also, subtracting mean is quite common.

Some extractor require rescale the range of value to `[-1, 1]`. And some others requires channel-wise rescaling. I think it is better to leave `prepare` unimplemented in SSD.

yuyu2172:
> Some extractor require rescale the range of value to [-1, 1]. And some others require channel-wise rescaling. I think it is better to leave prepare unimplemented in SSD.

I see. I will stop for today and think again tomorrow on this.

yuyu2172:
> In the current implementation, you make SSD300 to inherit in two levels.
I personally like less inheritance.

To elaborate this more, it takes more time for me to understand how links are grouped together when they are declared at multiple inheritance levels.

Hakuyume:
> I see. OrderedDict can work in that case.

I don't want to introduce such a thing in APIs of SSD because this is very optional and not necessary for SSD.  How about the following solution?
- `features()` returns a list of feature maps (defined as APIs of SSD)
- Some model can have an extra method, for example `extract()`. (not defined as APIs of SSD. SSD doesn't care of it)
    - This method returns a dictionary of features.
    - In such models, `features()` calls `extract()` internally to get feature maps.

Hakuyume:
> To elaborate this more, it takes more time for me to understand how links are grouped together when they are declared at multiple inheritance levels.

There is another group, SSD321 and SSD513. They are based on ResNet101.
In my idea, inheritance tree will be like below.
- SSD
    - SSDVGG16
        - SSD300
        - SSD512
    - SSDResNet101
        - SSD321
        - SSD513

yuyu2172:
> There is another group, SSD321 and SSD513. They are based on ResNet101.
In my idea, inheritance tree will be like below.
SSD
SSDVGG16
SSD300
SSD512
SSDResNet101
SSD321
SSD513
I think it looks clean.

In my taste, implementation of neural networks should make it easy to understand abstraction of layers at multiple levels.
In the case of SSD, I think that there are two levels of abstractions.

1. Divided the entire forward in three steps. (a) Extracting multiple features using a commonly used conv net, (b) Extract additonal features while more aggressively reducing spatial dimensions (c) Using the features to predict `loc` and `conf`.
2. The more lower level of abstraction is about the concrete steps taken at each module described above.

We are not quite agreed upon the higher-level abstraction as you consider the modules (a)(b) together. I can see that from `features` that returns aggregate of the outputs of the two modules.
I think that (a) and (b) should be separated because these two are often discussed as independent building blocks as we can see from how SSD is extended.

In order to make explicit that (a), (b) and (c) are at same level of abstractions, I preferred to separate feature extractors into different classes.
But, even if they are not separated into different classes, I would like to see a `__call__` method that resonates with my conceptual image of the network. For example, it should look like below.

```python
def __call__(self, x):
    xs = self.feature(x)
    xs += self.head(xs[-1])
    return self._multibox(xs)
```

Also, this is the reason why I did not like the current implementation of `features` being called inside the `features` of SSD300.

yuyu2172:
> Some extractor require rescale the range of value to [-1, 1]. And some others require channel-wise rescaling. I think it is better to leave prepare unimplemented in SSD.

In my idea, we prepare a concrete class (or a function) that puts together a feature extractor, head and `loc`, `conf`. It is safe to implement `predict` there because it assumes the feature extractor.

chainercv/links/ssd/ssd.py
```diff
@@ -0,0 +1,257 @@
...
+
+    variance = (0.1, 0.2)
+
+    nms_threshold = 0.45
+    score_threshold = 0.6
+
+    conv_init = {
+        'initialW': initializers.GlorotUniform(),
+        'initial_bias': initializers.Zero(),
+    }
```
yuyu2172: I think these paramters `variance`, `nms_threshold`, `score_threshold`, `conv_init` should be modifiable. At least `nms_threshold` and `score_threshoold`.

How about passing them at `__init__`.

Hakuyume: At first, I thought so. But sometimes we want to change `nms_threshold` and `score_threshoold` after initialization. For example, we do both visualization and evaluation. I think it is better to note "they can be changed anytime" in docs about those two parameters.
`variance` and `conv_init` won't change after initalization. So I agree adding them to arguments of `__init__()`.

Hakuyume:
In my opinion, SSD has only two concepts, feature extractor and multibox. Feature extractor reduces the spatial dimension from original to 1x1 and returns all feature maps. Multibox computes `loc` and `conf` from feature maps.

You divided feature extractor into two parts, (a) feature and (b) head. But this separation doesn't seem reasonable to me. According to your example, you defined from `conv1_1` to `conv7` as `feature` and from `conv8_1` to `conv11_2` as `head`. I guess your definition relies on whether original network (for example VGG16) contains a layer or not. But this rule will not work on a full-scratch network (designed fo SSD and there is no original network). In addition, separating feature extractor into two makes it difficult to use skip layers.

chainercv/links/ssd/ssd.py
yuyu2172: In my understanding, mutable variables should not be declared at class level because they are constants for class and should not vary among class instantiations.
Instead, it should be declared inside `__init__` regardless of taking it as an argument. Since, any public attributes are public in Python, this will let users to modify it whenever needed.

On giving paramters as arguments to `__init__` or not, I think either way is fine.

yuyu2172:
>  But this rule will not work on a full-scratch network (designed fo SSD and there is no original network).

You convinced me on this point.
So, from meta-architecture level, the features and (c) mentioned above are put together.

It would be easier to understand this structure if `features` is documented.

I still think that separating features into different classes makes it easier to understand because the features will be at least one level shorter in hierarchy than the current one.

> SSD300 and SSD512 share not only the base feature extractor but also prepare. Actually, I had tried to separate the feature detector into another class. But finally, I found this two-level inheritance is most simple.

But, I understand some  difficulities (e.g. `loc` and `conf` are dependent on number of features), and I have not fully tried it by myself, so I can only give suggestions.

I will clean up my implementation of Faster RCNN in short time, so that we can consider two together.
Please wait on this.

Hakuyume:
> I still think that separating features into different classes makes it easier to understand because the features will be at least one level shorter in hierarchy than the current one.

Yes, we can separate feature extractor into another class even if we don't separate it into head and feature.
However, I don't think shorter level in hierarchy is easier to understand in this case. I think current design is easy to understand the hierarchy of models.

Anyway, it is better to consider Faster-RCNN and SSD together as you said. I will leave this design as is and work on training scripts (it will be another PR).

chainercv/links/ssd/ssd.py
```diff
@@ -0,0 +1,257 @@
...
+
+    This is a base class of Single Shot Multibox Detector.
+    It requires a feature extraction method and a preprocessing method.
+
+    [1] Wei Liu, Dragomir Anguelov, Dumitru Erhan, Christian Szegedy,
+    Scott Reed, Cheng-Yang Fu, Alexander C. Berg.
+    SSD: Single Shot MultiBox Detector. ECCV 2016.
+
+    Args:
+        n_class (int): The number of classes.
```
yuyu2172: Does this count the background?

Hakuyume: No, my code classifies bounding boxes into `n_class + 1` classes. I think this design is good. I will note the number does not contain `background`. Does it make sense?

yuyu2172: Why do you want the classifier to output `n_class + 1` classes?
Shouldn't `n_class` be same as `len(pascal_voc_labels)` in the case of VOC?
It would be wierd if we exclude one class from a list of classes when counting.

Hakuyume: SSD internally distinguish `background` from other positive classes. In addition, some models (for example YOLO and YOLOv2) use `n_class` classifier (they use objectness instead).

yuyu2172: At least, this should be distinguishable from other `n_class` used, if we were to exclude `background` from counting.
Or, we can make all models to use `n_class` with your definition. However, I think this is unlikely case because this is probably only used in Detection (e.g. Semantic Segmentation does not).

In that case, perhaps `n_fg_class` or `n_foreground_class` is good.

yuyu2172: > SSD internally distinguish background from other positive classes. 

Where?
Do you mean `_suppress`?

Hakuyume: No. `_suppress` skips `background` in my current code but it is not necessary. We can treat `background` in the same way as others in `_suppress`.
I meant the loss of SSD. It uses `hard negative mining` which distinguishes `background` from others explicitly.

Hakuyume: > Or, we can make all models to use n_class with your definition. However, I think this is an unlikely case because this definition is probably only used in Detection (e.g. Semantic Segmentation does not).

I think my definition is better for detection links because no one wants `background` outputs. As you pointed out, this definition doesn't work for semantic segmentation. Giving another name is good idea. `n_fg_class` looks good.

chainercv/links/ssd/ssd.py
Hakuyume: > mutable variables should not be declared at class level

I agree. I will fix it.

Hakuyume:
> However, I don't think shorter level in hierarchy is easier to understand in this case. I think current design is easy to understand the hierarchy of models.

Although I said current design is good, I am wondering what is the best design. I will try to separate extractor again.

chainercv/links/ssd/ssd.py
```diff
@@ -0,0 +1,278 @@
...
+
+        Returns:
+            tuple of chainer.Variable:
+            This method returns two variables, :obj:`loc` and :obj:`conf`.
+
+            * **loc**: A variable of float arrays of shape :math:`(K, 4)`, \
+                where :math:`K` is the number of default bounding boxes.
+            * **conf**: A variable of float arrays of shape \
+                :math:`(K, n\_class + 1)`, \
+                where :math:`K` is the number of default bounding boxes.
```
yuyu2172: Looking at your tests, it seems that the shape of outputs are batched (i.e. (1, K, 4)).
Is the documentation outdated?

Hakuyume: Thank you. It's just my mistake. As you pointed out, the outputs are batched. I will fix it.

yuyu2172: What you think of #163 ?

chainercv/links/ssd/ssd.py
Hakuyume: Thank you. It worked.

Hakuyume:
I separated feature extractor into another class. In my design, an feature extractor has `grid`, `prepare` and `__call__`.

yuyu2172:
Can you move code from `links/*` to `links/model/*` ?

yuyu2172:
@Hakuyume 
Demo seems to not work with the image downloaded from the following link (https://goo.gl/CZpHWY).

docs/source/reference/index.rst
```diff
@@ -10,6 +10,7 @@ ChainerCV Reference Manual
    datasets
    evaluations
    extensions
+   links
```
yuyu2172: please remove this

docs/source/reference/links/ssd.rst
```diff
@@ -0,0 +1,57 @@
...
+   :members:
+   :special-members:  __call__
+
+SSD
+~~~
+.. autoclass:: SSD
+   :members:
+   :special-members:  __call__
+
+VGG16
```
yuyu2172: ~~ length should equal string.

chainercv/links/model/ssd/ssd_vgg16.py
```diff
@@ -0,0 +1,326 @@
...
+        for i in range(8, 12 + 1):
+            h = ys[-1]
+            h = F.relu(self['conv{:d}_1'.format(i)](h))
+            h = F.relu(self['conv{:d}_2'.format(i)](h))
+            ys.append(h)
+        return ys
+
+
+class SSD300(SSD):
+    """Single Shot Multibox Detector.
```
yuyu2172: be more specific

chainercv/links/model/ssd/ssd_vgg16.py
```diff
@@ -0,0 +1,326 @@
...
+    This model uses :class:`~chainercv.links.model.ssd.VGG16Extractor300` as
+    its feature extractor.
+
+    Args:
+       n_fg_class (int): The number of classes excluding the background.
+       pretrained_model (str): The weight file to be loaded.
+           This can take :obj:`'voc0712'`, `filepath` or :obj:`None`.
+           The default value is :obj:`None`.
+
+            * :obj:`'voc0712'`: Load weights trained on Pascal VOC 2007 and \
```
yuyu2172: trainval splits of PASCAL VOC 2007 and 2012.

chainercv/links/model/ssd/multibox.py
```diff
@@ -0,0 +1,90 @@
...
+    This link computes :obj:`loc` and :obj:`conf` from feature maps.
+    :obj:`loc` contains information of the coordinates of bounding boxes
+    and :obj:`conf` contains that of classes.
+
+    .. [#] Wei Liu, Dragomir Anguelov, Dumitru Erhan,
+       Christian Szegedy, Scott Reed, Cheng-Yang Fu, Alexander C. Berg.
+       SSD: Single Shot MultiBox Detector. ECCV 2016.
+
+    Args:
+        n_class (int): The number of classes.
```
yuyu2172: Should imply that the background can be included in n_class.
How about "The number of classes possibly including the background.

chainercv/links/model/ssd/ssd.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def to_gpu(self):
+        super(SSD, self).to_gpu()
+        self._default_bbox = chainer.cuda.to_gpu(self._default_bbox)
+
+    def __call__(self, x):
+        """Compute localization and classification from a batch of images.
+
+        This method computes two variables, :obj:`loc` and :obj:`conf`.
+        :meth:`_decode` converts these variables to prediction.
```
yuyu2172: s/prediction/bounding box coordinates and confidence scores.

chainercv/links/model/ssd/normalize.py
```diff
@@ -0,0 +1,48 @@
...
+        self.add_param(
+            'scale', n_channel,
+            initializer=initializers._get_initializer(initial))
+
+    def __call__(self, x):
+        """Normalize input and scale it.
+
+        Args:
+            x (chainer.Variable): A variable holding 4-dimensional array.
+                The dimensions is :math:`(batch, channel, height, width)`.
```
yuyu2172: I think the line `The dimension ....` is not really necessary as the phrase "4-dimensional array" implies the shape.

chainercv/links/model/ssd/ssd.py
```diff
@@ -0,0 +1,249 @@
...
+
+        return bbox, label, score
+
+    def _prepare(self, img):
+        img = img.astype(np.float32)
+        img = transforms.resize(img, (self.insize, self.insize))
+        img -= np.array(self.mean)[:, np.newaxis, np.newaxis]
+        return img
+
+    def use_preset(self, preset):
```
yuyu2172: can you add doc here?
You can use one for Faster RCNN

chainercv/links/model/ssd/ssd.py
```diff
@@ -0,0 +1,249 @@
...
+        steps (iterable of float): The step size for each feature map.
+        sizes (iterable of float): The base size of default bounding boxes
+            for each feature map.
+        variance (tuple of float): Two coefficients for encoding
+            the locations of bounding boxe. The first value is used to
+            encode coordinates of the centers. The second value is used to
+            encode the sizes of bounding boxes.
+            The default value is :obj:`(0.1, 0.2)`.
+
+    Parameters:
```
yuyu2172: docs does not seem to differentiate Args and `nms_thresh`, `score_thresh`

chainercv/links/model/ssd/ssd.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def _suppress(self, raw_bbox, raw_score):
+        xp = self.xp
+
+        bbox = list()
+        label = list()
+        score = list()
+        for l in range(self.n_fg_class):
+            bbox_l = raw_bbox
+            score_l = raw_score[:, l + 1]
```
yuyu2172: shouldn't you add a comment on why you add `l+1`?

chainercv/links/model/ssd/ssd_vgg16.py
```diff
@@ -0,0 +1,365 @@
...
+            if not os.path.exists(path):
+                download_file = download.cached_download(model['url'])
+                os.rename(download_file, path)
+
+            if not _available:
+                warnings.warn(
+                    'cv2 is not installed on your environment. '
+                    'Pretrained models are trained with cv2.'
+                    'The performace may change with Pillow backend.',
+                    RuntimeWarning)
```
yuyu2172: I thought the warning is raised in demo.py.

chainercv/links/model/ssd/ssd.py
Hakuyume: They look same although I separated them into `Args` and `Parameters`. Do you have any idea?

yuyu2172:
LGTM

yuyu2172:
No longer WIP?

examples/ssd/README.md
```diff
@@ -0,0 +1,25 @@
+# Examples of Single Shot Multibox Detector [1]
+
+## Performance
+PASCAL VOC2007 Test
+
+| Model | Original | ChainerCV |
+|:-:|:-:|:-:|:-:|
+| SSD300 | 77.5 % [2] | 77.5 % |
+| SSD512 | 79.5 % [2] | 79.6 % |
+
```
yuyu2172: The style is broken.

examples/ssd/README.md
```diff
@@ -0,0 +1,25 @@
+# Examples of Single Shot Multibox Detector [1]
+
+## Performance
+PASCAL VOC2007 Test
+
+| Model | Original | ChainerCV |
+|:-:|:-:|:-:|
+| SSD300 | 77.5 % [2] | 77.5 % |
+| SSD512 | 79.5 % [2] | 79.6 % |
```
yuyu2172: Isn't this mAP and not accuracy?

Hakuyume: Yes, this is mAP. `%` is commonly used not only for accuracy but also for mAP.

yuyu2172: I prefer more explicit notation.
I found that `%` is used, but it is used as something like `% mAP`.

Hakuyume: I agree with you. I added more notation about scores.

yuyu2172:
LGTM!

PR #124, opened 2017-04-12
use CHW and BGR

Hakuyume:


chainercv/utils/image_utils.py
```diff
@@ -2,22 +2,23 @@
 from PIL import Image
 
 
-def read_image_as_array(path, dtype=np.uint8, copy=True):
+def read_image_as_array(path, dtype=np.uint8, copy=True, force_color=False):
```
yuyu2172: Since images seem to follow data type protocol used in transforms, shouldn't the default value of `dtype` be `np.float32`?

chainercv/tasks/detection/vis_bbox.py
```diff
@@ -15,8 +17,8 @@ def vis_bbox(img, bbox, label=None, label_names=None, ax=None):
         >>> plot.show()
 
     Args:
-        img (~numpy.ndarray): An array of shape :math:`(height, width, 3)`.
-            This is in RGB format.
+        img (~numpy.ndarray): An array of shape :math:`(3, height, width)`.
+            This is in BGR format.
```
yuyu2172: Can you add the range of image? [0, 255]?

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -78,7 +79,7 @@ class DetectionVisReport(chainer.training.extension.Extension):
 
             img, bbox, label = vis_transform(inputs)
 
-        :obj:`img` should be an image which is in HWC format, RGB and
+        :obj:`img` should be an image which is in CHW format, BGR and
```
yuyu2172: Can you add the range of image, which is [0, 255].

chainercv/utils/image_utils.py
yuyu2172: I think the dafault value of `force_color` should be `True`.
This is because gray scale images are expected to be arrays of shape (3, H, W) and not (1, H, W), and this should be the default behavior.

yuyu2172:
@Hakuyume 

Do you have time to fix this?

Hakuyume:
@yuyu2172 Sorry, I fixed.

yuyu2172:
Can I have a discussion on the interface of `read_image`?
While having an option like `force_color` is fine, I am not satisfied with its behavior when `force_color = False`.

Since it is hard to come up with a scenario, where a user want to load both a color and a gray image depending on image format, I think that it is better to make this function to always load image as a gray image or always as a color image.

This can be done with an option like `grayscale`, which makes this function to load image always as color when False, and gray image (shape 1HW) otherwise.

yuyu2172:
Also, I would appreciate if you can add a test for read_image_array.

chainercv/utils/image_utils.py
```diff
@@ -2,22 +2,40 @@
 from PIL import Image
 
 
-def read_image_as_array(path, dtype=np.uint8, copy=True):
+def read_image(path, dtype=np.float32, copy=True, force_color=True):
+    """Read image from file.
```
yuyu2172: Read an image from a file.

chainercv/utils/image_utils.py
```diff
@@ -2,22 +2,40 @@
...
+
+    This function reads image from given file. The image is CHW format. The
```
yuyu2172: This function reads an image

chainercv/utils/image_utils.py
```diff
@@ -2,22 +2,40 @@
...
+    range of value is :math:`[0, 255]`. If the image is color, the order of the
```
yuyu2172: range of its value

yuyu2172: If the image is a color image,

chainercv/utils/image_utils.py
```diff
@@ -2,22 +2,40 @@
...
+    channels is BGR.
+
+    Args:
+        path (str): Path of image file.
```
yuyu2172: A path of an image file.

chainercv/utils/image_utils.py
```diff
@@ -2,22 +2,40 @@
...
+        dtype: The type of array. The default is :obj:`~numpy.float32`.
```
yuyu2172: The default value

chainercv/utils/image_utils.py
```diff
@@ -2,22 +2,40 @@
...
+        copy (bool): Make the array mutable.
+        force_color (bool): If :obj:`True`, the number of channels is 3.
```
yuyu2172: s/3/three

yuyu2172:
Can you fix vis_keypoint as well?

Hakuyume:
> it is hard to come up with a scenario, where a user want to load both a color and a gray image depending on image format,
I agree with you. I renamed the option `force_color` -> `color`. This option determines the number of channels.

Hakuyume:
I renamed `utils.image_utils` -> `utils.image` for consistency.
We are using `transforms.image` (not `transforms.image_transforms`) or `tasks.detection` (not `tasks.detection_tasks`).

yuyu2172:
I think changing the utility file names is good.

Can you change other utility module names to be consistent?

Hakuyume:
I renamed other utilities and fixed some bugs.

Hakuyume:
`chw_to_pil_image` is mainly used with `imshow`. I think it is better to add `show_image` and remove `chw_to_pil_image`.

Hakuyume:
I found #135. How about adding `chainercv.visualizations.vis_image` and using it from `vis_bbox` and `vis_keypoint`?

yuyu2172:
> chw_to_pil_image is used with imshow. I think it is better to add vis_image and remove chw_to_pil_image

Sure. That is nice.

yuyu2172:
LGTM

PR #126, opened 2017-04-19
Add fit option to scale

yuyu2172:


rezoo:
Why did you decide to define ``size`` option as ``int``? Is it better to accept and integer and a tuple of ints?

yuyu2172:
Sorry I am confused. What do you mean by tuple of ints?
Currently, the `size` indicates the length of the shorted edge.

chainercv/transforms/image/scale.py
```diff
@@ -21,9 +28,14 @@ def scale(img, size):
     if (W <= H and W == size) or (H <= W and H == size):
         return img
```
Hakuyume: If `fit_short` is `False`, this `return` statement causes an undesired behaviour.

tests/transforms_tests/image_tests/test_scale.py
```diff
@@ -6,19 +6,23 @@
...
+    {'in_shape': (3, 24, 16), 'size': 8,
+     'fit_short': True, 'out_shape': (3, 12, 8)},
+    {'in_shape': (3, 16, 24), 'size': 8,
+     'fit_short': True, 'out_shape': (3, 8, 12)},
+    {'in_shape': (3, 24, 16), 'size': 36,
+     'fit_short': False, 'out_shape': (3, 36, 24)},
+    {'in_shape': (3, 16, 24), 'size': 36,
+     'fit_short': False, 'out_shape': (3, 24, 36)}
+)
 class TestScale(unittest.TestCase):
```
Hakuyume: Cound you add more test cases that `size == H` and `size == W`?

yuyu2172:
Thanks. I fixed the code.

Hakuyume:
LGTM

Hakuyume:
Sorry, this function doesn't assume the number of channels is three. It may work with grayscale images.
Could you add test cases of grayscale?

yuyu2172:
Yeah. Thanks for pointing that out.

PR #127, opened 2017-04-20
Delete links.googlenet

yuyu2172:
Since GoogLeNet is added to Chainer, this link does not need to exist in ChainerCV anymore.


rezoo:
LGTM.

PR #130, opened 2017-04-26
Add an evaluation code for detection

yuyu2172:
Merge after #129.

- [x] Verify that it reports same score with the original MATLAB evaluation for PASCAL VOC Challenge

The code produces same result obtained by the original PASCAL dataset.
A code that verify this can be found at https://github.com/yuyu2172/pascal_evaluation.


yuyu2172:
The API for this evaluation code requires revision.
In particular, this implementation has not yet taken MS COCO style evaluation into account.

@hakuyume Do you have a suggestion?

Hakuyume:
You are using both `cls` and `label`. What is the difference between them?

Hakuyume:
As you know, MS COCO style evaluation is more complicated.
http://mscoco.org/dataset/#detections-eval

To make APIs simple, how about providing a function which returns precision/recall array.
```
pd_bboxes = {'001': bbox, ...}
pd_scores = {'001': score, ...}
gt_bboxes = {'001': bbox, ...}
prec, rec = prec_and_rec(pd_bboxes, pd_scores, gt_bboxes, threshold=0.5)
```

yuyu2172:
I limited the evaluation codes' scope to VOC.

@Hakuyume 
Can you check the code?

yuyu2172:
`cls` indicates that the contents of a list is conditioned by a class.
For example, `bboxes_cls == _bboxes[cls]` for some `cls in this code.

I think `bboxes_label` sounds different from what I intend.

Hakuyume:
`bboxes_cls` is "a list/tuple of arrays of bounding boxes which belong to a certain class", is it right?

yuyu2172:
Yes. Thanks for putting that into a word.

yuyu2172:
In the code, I have to access an element or small subset of arrays named in singular form.
In that case, I used abbreviation to refer to them.

For example, subset or element of `bbox` would be `bb` and `index` would be `idx`.

Is this convention reasonable?
If we extend this convention, `img` will be something like `im`.

yuyu2172:
I fixed code to be consistent with other part of ChainerCV.

yuyu2172:
Hmm. The output slightly changed from the original implementation. Please wait.

chainercv/evaluations/eval_detection_voc.py
```diff
@@ -0,0 +1,245 @@
...
+            its index corresponds to an index for the base dataset.
+            Its length is :math:`N`.
+        gt_bboxes (list of numpy.ndarray): List of ground truth bounding boxes
+            whose length is :math:`N`. An element of :obj:`gt_bboxes` is a
+            bounding box whose shape is :math:`(R, 4)`. Note that number of
+            bounding boxes in each image does not need to be same as the number
+            of corresponding predicted boxes.
+        gt_labels (list of numpy.ndarray): List of ground truth labels which
+            are organized similarly to :obj:`labels`.
+        n_class (int): Number of classes.
```
Hakuyume: Is `n_class` needed? We can compute `valid_cls_index` without `n_class`.

chainercv/evaluations/eval_detection_voc.py
```diff
@@ -0,0 +1,245 @@
+from __future__ import division
+
+import numpy as np
+import six
+
+
+def eval_detection_voc(
+        bboxes, labels, scores, gt_bboxes, gt_labels, n_class,
```
Hakuyume: `pred_bboxes`, `pred_labels` and `pred_scores`

chainercv/evaluations/eval_detection_voc.py
yuyu2172: Great idea

chainercv/evaluations/eval_detection_voc.py
yuyu2172: fixed

chainercv/evaluations/eval_detection_voc.py
```diff
@@ -0,0 +1,245 @@
...
+    has :math:`N` images.
+
+    Mean average precision is calculated by taking a mean of average
+    precision for all classes which have at least one bounding box
+    assigned by prediction or ground truth labels.
+    The code is based on the evaluation code used in PASCAL VOC Challenge.
+
+    Args:
+        pred_bboxes (list of numpy.ndarray): A list of bounding boxes.
+            The index to this list corresponds to the index of the data
```
Hakuyume: `index of this list`?

chainercv/evaluations/eval_detection_voc.py
```diff
@@ -0,0 +1,245 @@
...
+            obtained from the base dataset. Length of the list is :math:`N`.
+            The element of :obj:`pred_bboxes` is coordinates of bounding
```
Hakuyume: How about `Each element of :obj:`pred_bboxes` is a set of coordinates of~`?

chainercv/evaluations/eval_detection_voc.py
```diff
@@ -0,0 +1,245 @@
...
+            boxes. This is an array whose shape is :math:`(R, 4)`,
+            where :math:`R` corresponds
+            to the number of bounding boxes, which may vary among boxes.
+            The second axis corresponds to :obj:`x_min, y_min, x_max, y_max`
+            of a box.
```
Hakuyume: `a bounding box`

chainercv/evaluations/eval_detection_voc.py
```diff
@@ -0,0 +1,245 @@
...
+        gt_labels (list of numpy.ndarray): List of ground truth labels which
+            are organized similarly to :obj:`labels`.
+        gt_difficults (list of numpy.ndarray): List of boolean arrays which
+            is organized similarly to :obj:`labels`. This tells whether the
+            corresponding ground truth bounding box is difficult or not.
+            By default, this is :obj:`None`. In that case, this function
+            consider all bounding boxes to be not difficult.
+        min_iou (float): A prediction is correct if its Intersection over
+            Union with the ground truth is above this value.
+        use_07_metric (bool): Whether to use Pascal VOC 2007 evaluation metric
```
Hakuyume: You are using `PASCAL VOC` above.  Actually, both `Pascal VOC` and `PASCAL VOC` exist in the official document. However, I think it is better to unify the name in ChainerCV.

chainercv/evaluations/eval_detection_voc.py
```diff
@@ -0,0 +1,245 @@
...
+        bboxes, scores, gt_bboxes, gt_difficults, min_iou=0.5):
+    # Calculate detection metrics with respect to a class.
+    # This function is called only when there is at least one
+    # prediction or ground truth box which is labeled as the class.
+    # bboxes: List[numpy.ndarray]
+    # scores: List[numpy.ndarray]
+    # gt_bboxes: List[numpy.ndarray]
+    # gt_difficults: List[numpy.ndarray]
+
+    npos = 0
```
Hakuyume: Does `npos` stand for `the number of positives`? If so, `n_pos` fits our naming rules.

chainercv/evaluations/eval_detection_voc.py
```diff
@@ -0,0 +1,245 @@
...
+        # VOC evaluation follows integer typed bounding boxes.
+        gt_bb_area = np.prod(gt_bb[:, 2:] - gt_bb[:, :2] + 1., axis=1)
+
+        if gt_bb.size > 0:
+            lt = np.maximum(gt_bb[:, :2], bb[:2])
+            rb = np.minimum(gt_bb[:, 2:], bb[2:])
+            area = np.prod(np.maximum(rb - lt + 1, 0), axis=1)
+            iou = area / (bbox_area[d] + gt_bb_area - area)
+            ioumax = np.max(iou)
+            jmax = np.argmax(iou)
```
Hakuyume: Can you write this part using `utils.bbox.bbox_overlap`?

yuyu2172:
Fixed.
Perhaps, the name of `bbox_overlap` should be changed to `bbox_iou`.

Hakuyume:
> Perhaps, the name of bbox_overlap should be changed to bbox_iou.
LGTM

yuyu2172:
OK. I will change it in this PR.

Hakuyume:
LGTM

PR #141, opened 2017-05-03
Update testing utils

Hakuyume:
This PR contains some updates.

- Rename `utils.test` -> `utils.testing`
     In Chainer, utilities for `nosetests` are in `testing` namespace.
- Update `utils.testing.ConstantReturnsModel`
    - Change the base class to `chainer.Link` (this is more basic than `chainer.Chain`)
    - Modify to return random values.
        Currently, this class is used with `numpy.random.uniform`.
    - Rename `ConstantReturnsModel` -> `StubLink`
    - Add docs


chainercv/utils/testing.py
```diff
@@ -0,0 +1,79 @@
...
+            raise ValueError('value must be \'uniform\', int or float')
+
+        self._get_array = _get_array
+
+    def __call__(self, *_):
+        """Returns dummy value(s).
+
+        Args:
+            This method can take any values as argument.
+            All of them are ignored.
```
yuyu2172: How about 

This function returns values independent of the arguments.

chainercv/utils/testing.py
```diff
@@ -0,0 +1,79 @@
...
+        else:
+            raise ValueError('value must be \'uniform\', int or float')
+
+        self._get_array = _get_array
+
+    def __call__(self, *_):
+        """Returns dummy value(s).
+
+        Args:
+            This method can take any values as argument.
```
yuyu2172: s/argument/arguments

chainercv/utils/testing.py
```diff
@@ -0,0 +1,79 @@
+import numpy as np
+
+import chainer
+
+
+class StubLink(chainer.Link):
+    """A chainer.Link that returns dummy value(s).
+
+    This is a :obj:`chainer.Link` that returns dummy
+    :obj:`chainer.Variable` when :meth:`__call__` method is called.
```
yuyu2172:
```
s/:obj:`chainer.Variable`/:obj:`chainer.Variable`(s)
```

docs/source/reference/utils.rst
```diff
@@ -0,0 +1,14 @@
+Utils
+==========
```
yuyu2172: === and Utils should have same length.

yuyu2172:
I would like to have a functionality where this link can take arbitrary arrays and return them as is.

yuyu2172:
On the API side, instead of `*shapes*, **kwargas**`, I prefer this link to take a list or tuple of shapes and arrays as inputs.

Hakuyume:
> I would like to have a functionality where this link can take arbitrary arrays and return them as is.

Do you have any examples in which this functionality is required? I would like to design the APIs based on examples.

yuyu2172:
> Do you have any examples in which this functionality is required? I would like to design the APIs based on examples.

At one time, I tried to implement an extension that iterates over entire datasets and store all the embedded vectors in a filesystem.
The extension was deleted now, but it was helpful to have a StubLink that has deterministic output for testing.

The examples are a month old, but here it is.
https://github.com/pfnet/chainercv/blob/515ec1f3e8e73b56ba960d998124cff5a591edad/tests/extensions_tests/misc_tests/test_embed_images.py

Hakuyume:
I see. For this situation, how about adding `history` property to `StubLink`? This property stores all the arrays returned from `__call__` method.

Hakuyume:
I think `history` is better than deterministic outputs because we can test `EmbedImages` uses all outputs (not copying the first output ).

Hakuyume:
Sorry, I found current APIs are not good in some cases. For example, `DetectionVisReport` assumes the model returns `np.float32` for bbox and `np.int32` for label.

yuyu2172:
Sorry.
I think I have led the discussion in the wrong direction with the previous comment.
The example was not appropriate as a use case of a stub link with a deterministic __call__ function.
Please forget it.
Also, I think that this scenario is a lot more tricky to handle than what we are handling right now.
I would probably write a class that applies a deterministic function to input (e.g. linear transform). For testing, it can simply be done by comparing batch of outputs and tensors which the function applied to (shape NCHW).

In my opinion, it is better to stick with the original feature that `ConstantReturnModel` supported for the time being (although its name should be something like `ConstantLink` or `ConstantReturnLink`).
Although there is no example to use deterministic feature at present, it would be useful if you want to verify the contents produced by the link (e.g. check actual image created by Extension).
On the other hand, I am uncertain on when I would need to have a non-deterministic behaviour that you are trying to support.
Besides, the design would be much simpler if you stick with the original one.
The non-deterministic approach gets a little bit cumbersome when you try to support arbitrary kinds of shapes, shapes and length of tuple.

By the way, I like how you changed the link to inherit from `chainer.Link`.

I really thank all your effort. Also, sorry for some premature API designs that I have made with respect to modules relating to extensions. I think they are somewhat in a mess...

Hakuyume:
OK, I understood the importance of `Constant`. I changed `StubLink` to `ConstantStubLink`. I will add other `StubLink`s when they are needed.

tests/utils_tests/testing_tests/test_constant_stub_link.py
```diff
@@ -0,0 +1,78 @@
...
+    {'outputs': ({'shape': (3, 4), 'dtype': np.float32},)},
+    {'outputs': (
+        {'shape': (3, 4), 'dtype': np.float32},
+        {'shape': (3,), 'dtype': np.int32})},
+)
+class TestConstantStubLink(unittest.TestCase):
+
+    def setUp(self):
+        if isinstance(self.outputs, tuple):
+            self.outputs = tuple(np.empty(**output) for output in self.outputs)
```
yuyu2172: `np.random.uniform` is better because it is guaranteed that the output is not all 0 (which is a trivial test case).

docs/source/reference/utils.rst
```diff
@@ -0,0 +1,14 @@
...
+.. module:: chainercv.utils
+
+
+read_image
+~~~~~~~~~~
+.. autofunction:: read_image
+
+StubLink
+~~~~~~~~
+.. autoclass:: StubLink
```
yuyu2172: You need to update the name to  ConstantStubLink.

chainercv/utils/testing.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
+
+import chainer
+
+
+class ConstantStubLink(chainer.Link):
+    """A chainer.Link that returns constant value(s).
+
+    This is a :obj:`chainer.Link` that returns constant
+    :obj:`chainer.Variable`(s) when :meth:`__call__` method is called.
```
yuyu2172: I think you need a space between `chainer.Variable` and (s).

yuyu2172:
LGTM!

PR #142, opened 2017-05-03
Remove util.extension

Hakuyume:
- [x] add tests for current APIs
- [ ] change APIs
- [ ] update docs (if needed)


yuyu2172:
We can delete utils.extensions.

Hakuyume:
> We can delete utils.extensions since it is no longer necessary after #144.

Yes, extensions get more simple after #144. I changed this PR for removing `utils.extension`.

Hakuyume:
@yuyu2172 Could you check this PR?

yuyu2172:
LGTM

PR #144, opened 2017-05-04
Add abstract class for object detection

Hakuyume:
- [x] Add `DetectionLink`
- [x] Support `score` in `vis_bbox`
- [x] Modify `DetectionVisReport` to use `DetectionLink`


Hakuyume:
I defined APIs for object detectors. And modified some functions to fit the APIs.
@yuyu2172 how do you think of this APIs?

chainercv/links/detection_link.py
```diff
@@ -0,0 +1,39 @@
+import chainer
+
+
+class DetectionLink(chainer.Link):
```
yuyu2172: Why inherit from `chainer.Link`?
I think that most of the classes that inherit from this would want to inherit from `chainer.Chain`, which can have children links.

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -27,179 +22,52 @@ def _check_available():
...
 
     invoke_before_training = False
 
-    def __init__(self, indices, dataset, target,
-                 filename_base='detection', predict_func=None,
-                 vis_transform=_detection_vis_transform):
+    def __init__(
+            self, dataset, target,
+            filename='detection_idx={index}_iter={iteration}.jpg'):
         _check_available()
```
yuyu2172: I prefer filename in order of iteration/index because I want images produced in the same iteration to be the neighboring images in filesystems.

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -210,62 +78,30 @@ def __call__(self, trainer):
...
 
-        for idx in self.indices:
-            formated_filename_base = osp.join(trainer.out, self.filename_base)
-            out_file = (formated_filename_base +
-                        '_idx={}'.format(idx) +
-                        '_iter={}'.format(trainer.updater.iteration) + '.jpg')
-
-            inputs = self.dataset[idx]
-            self._check_type_dataset(inputs)
+        label_names = getattr(self.dataset, 'labels', None)
```
yuyu2172: `label_names` should be passed as an argument.
The dataset loses attributes once wrapped by classes like TransformDataset.

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -27,179 +22,52 @@ def _check_available():
...
-            method will be called.
-        vis_transform (callable): A callable that is used to convert tuple of
-            arrays returned by :obj:`dataset.__getitem__`. This function
-            should return tuple of arrays which can be used for visualization.
-            More detail can be found at the description on the step 3 of
-            internal mechanics found above.
-
+        target (chainercv.links.DetectionLink): Link object used for detection.
+        dataset : Dataset class that produces inputs to :obj:`target`.
+        filename (str): Basename for saved image.
```
yuyu2172: for saved images

chainercv/extensions/detection/detection_vis_report.py
yuyu2172: Also, `vis_transform` is necessary because you can not visualize images produced by datasets that subtract mean values.

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -210,62 +78,30 @@ def __call__(self, trainer):
...
-            if pred_bbox.ndim == 2:
-                # force output to have batch axis
-                pred_bbox = pred_bbox[None]
-                pred_label = pred_label[None]
-            self._check_type_model((pred_bbox, pred_label))
-            pred_bbox = pred_bbox[0]  # (B, R, 5) -> (R, 5)
+        for idx in six.moves.range(len(self.dataset)):
+            out_file = self.filename.format(
+                index=idx, iteration=trainer.updater.iteration)
+            out_file = osp.join(trainer.out, out_file)
```
yuyu2172: It was my mistake, but could you use os.path instead of osp?
Also, osp should be deleted from imports.

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -210,62 +78,30 @@ def __call__(self, trainer):
...
 
-            vis_transformed = self.vis_transform(inputs)
-            self._check_type_vis_transformed(vis_transformed)
-            vis_img = vis_transformed[0]
-            raw_bbox = vis_transformed[1]
+            img, gt_bbox, gt_label = self.dataset[idx]
```
yuyu2172: how about self.dataset[idx][:3]

chainercv/links/detection_link.py
```diff
@@ -0,0 +1,39 @@
...
+        """Detect objects in an image.
+
+        This method detects objects in an image.
+        This method returns a tuple of three arrays,
+        :obj:`(bbox, label, score)`.
+
+        Args:
+            img (~numpy.ndarray): An array of shape :math:`(3, height, width)`.
+                This is in BGR format and the range of its value is
+                :math:`[0, 255]`.
```
yuyu2172: I think inputs and outputs should be batches.
I think inputs of Links are usually batched in Chainer models including ResNet.

yuyu2172:
Here is my comments.
Since this is WIP I haven't checked tests and docs yet.

chainercv/extensions/detection/detection_vis_report.py
Hakuyume: OK, I will use `os.path`.

chainercv/links/detection_link.py
Hakuyume: The number of bounding boxes is not constant. It is difficult to pack predictions of some images into an array.

chainercv/extensions/detection/detection_vis_report.py
Hakuyume: Actually, current `filename` was copied from original codes. As you suggested, I will change the order to iteration/index.

chainercv/extensions/detection/detection_vis_report.py
Hakuyume: OK, I will add `label_names` argument to `__init__`.

chainercv/extensions/detection/detection_vis_report.py
Hakuyume: Are you planning add some detection datasets which return additional data?

chainercv/links/detection_link.py
yuyu2172: I see.
How about iterable of numpy.ndarrays and return list of bboxes.

This seems to be the convention that ResNet is using.

chainercv/extensions/detection/detection_vis_report.py
yuyu2172: Oh, my bad.

chainercv/links/detection_link.py
yuyu2172: Besides, list of bounding boxes is used in `eval_detection`.
I think it is not strange to return them.

chainercv/links/detection_link.py
Hakuyume: I used `chainer.Link` because it is most basic class. If you want to have child links, you can write `class MyDetector(chianer.Chain, DetectionLink)`. If you want to treat child links as a list, you can write `class MyDetector(chianer.ChainList, DetectionLink)`.

chainercv/links/detection_link.py
```diff
@@ -0,0 +1,39 @@
...
+
+        Returns:
+            (~numpy.ndarray, ~numpy.ndarray, ~numpy.ndarray):
```
yuyu2172: I think it is better to return variables.
Also, it should be noted that data contained is xp.ndarray.

chainercv/extensions/detection/detection_vis_report.py
yuyu2172: FasterRCNN takes amount of scaling to input image as input.
For instance, if an image is scaled from 600 to 900, FRCNN takes scaling value 1.5 as input.

yuyu2172: Thus, the input it takes is `img, bbox, label, scale`.

Hakuyume: If a model needs scaling before forward, I think the scaling should be done in `predict` method of the model.

chainercv/links/detection_link.py
Hakuyume: I thought `numpy.ndarray` is natural for outputs because the inputs are `numpy.ndarray`.
But this design is not fixed yet.

chainercv/links/detection_link.py
Hakuyume: I know we sometimes want to pass a batch to a model for speed.
How about `[(bbox, label, score), (bbox, label, score)...]`?

chainercv/links/detection_link.py
yuyu2172: I am not a big fan of double inheritance in general because its `__init__` is confusing.
Also, it may give an impression that ChainerCV builds a lot of abstraction on top of Chainer for using a complex OO design.

On the other hand, I can see benefit of using multiple inheritance, too.

I am probably not convinced because there isn't any example that seems to benefit from extending from `ChainList` or `Link` directly with `DetectionList`.

chainercv/links/detection_link.py
yuyu2172: Why do you think that list of tuples is good?

chainercv/links/detection_link.py
Hakuyume: Actually, I don't like multiple inheritance, either. However, I don't like to assume that all detectors inherit `chainer.Chain` although there are no examples inheriting `chainer.ChainList`.
One simple solution is to remove `DetectionLink` and define APIs as a document. We can choose base class for each detector without multiple inheritance.
(cf. `ResNet` and `VGG16` have similar APIs, but they don't share any base classes.)

yuyu2172: Since there is no actual code in `DetectionLink`, we may not need it.
It was still fruitfull trying to implement it and deepen our understandings on API of the predict function.

chainercv/extensions/detection/detection_vis_report.py
yuyu2172: `scale` is actually used to filter out some small region proposals.
In my opinion, scaling bounding boxes to the size of original image should not be done in `predict`.
The current implementation in FRCNN does it, but I am going to fix it.

chainercv/links/detection_link.py
Hakuyume: Sorry, I think `[bbox, bbox], [label, label], [score, score]` is better because it is consistent with the shapes of batch.
I thought a list of tuples is more convenient to iterate each example. But it can be easily done by `zip` function.

yuyu2172: OK. Thanks!

Hakuyume: Let me confirm. The input of `predict` is a list of `numpy.ndarray`. Is it OK?
If it is so, `[(bbox, label, score), (bbox, label, score)...]` looks consistent to me.

chainercv/extensions/detection/detection_vis_report.py
Hakuyume: In my design, `predict` takes images in any sizes and returned bounding boxes are scaled to fit the original sizes.

chainercv/links/detection_link.py
yuyu2172: Input is a list of `np.ndarray`.

I think both plans are consistent with inputs.
Not only that, tuple of lists is consistent with how other Chainer models return tuple of batched arrays.

chainercv/extensions/detection/detection_vis_report.py
yuyu2172: I thought `predict` does not take `scale`.
Also, I think that it would be a burden for all detection models to support `scale` options.

chainercv/links/detection_link.py
yuyu2172: Yeah, not only is it consistent among inputs and outputs, bounding boxes may not be included in computational graphs making it impossible to return as variables.
I think return values of `predict` should be Variables in ChainerCV, although it is different from Chainer's classifiers.

Hakuyume: Are there anyone who want to back-prop from the results of `predict`? If no, I think `ndarray` is better than `Variables`?

Hakuyume:
I updated.

- Remove `DetectionLink` https://github.com/pfnet/chainercv/pull/144#discussion_r114769462.
    I will add a doc which describes the APIs of detection links. (where should I add it?)
- Change the input of `predict` from `img` to `imgs` https://github.com/pfnet/chainercv/pull/144#discussion_r114757410.
- Change the output of `predict` from `bbox, label, score` to `bboxes, labels, scores`. https://github.com/pfnet/chainercv/pull/144#discussion_r114777871
    The type of output, `ndarray` or `Variable`, is under discussion https://github.com/pfnet/chainercv/pull/144#discussion_r114762476. I simply add `cupy` support.

chainercv/links/detection_link.py
yuyu2172: Oh, this was typo! Sorry!

> I think return values of predict should be Variables
What I meant was opposite. I think it should be `ndarray`.

yuyu2172:
> Remove DetectionLink #144 (comment).
I will add a doc which describes the APIs of detection links. (where should I add it?)

This can be added at top of `links.rst`.
However, I think that this can be done after adding concrete models.
For the time being, we only need to make sure that newly added models (e.g. SSD and FRCNN) to follow the conventions we agreed on.

Hakuyume:
> However, I think that this can be done after adding concrete models
OK. I will add the doc in another PR.

chainercv/links/detection_link.py
Hakuyume: OK, that makes sense.

Hakuyume:
APIs for detection links
- It inherits `chainer.Link` directly or indirectly.
- It has `predict` method.
    This method takes a batch of images and returns predictions.
    - inputs:
        - imgs: a list of `numpy.ndarray`. Each array holds an image, in CHW, BGR, [0, 255] format. The sizes of images are not same even in a batch.
    - outputs:
        - bboxes: a list of `numpy/cupy.ndarray`
        - labels: a list of `numpy/cupy.ndarray`
        - scores: a list of `numpy/cupy.ndarray`

yuyu2172:
Thanks for summarizing the discussion.
I support that conclusions.

Hakuyume:
I updated this PR to obey the design mentioned in https://github.com/pfnet/chainercv/pull/144#issuecomment-299682956.
@yuyu2172 Could you review this PR?

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -210,62 +82,48 @@ def __call__(self, trainer):
...
-    trainer = mock.MagicMock()
-    out_dir = tempfile.mkdtemp()
-    print('outdir ', out_dir)
-    trainer.out = out_dir
-    trainer.updater.iteration = 0
-    extension = DetectionVisReport([3], train_data, model)
-    extension(trainer)
+        if hasattr(self.iterator, 'reset'):
+            self.iterator.reset()
+            it = self.iterator
```
yuyu2172: I do not like to change state of the iterator that this class takes as an argument using `reset`.
In the current code, if the iterator passed to this Reporter is used somewhere else, unexpected effects will occur.

How about copying the iterator at `__init__`?

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -210,62 +82,48 @@ def __call__(self, trainer):
...
+
+            for (img, gt_bbox, gt_label), pred_bbox, pred_label, pred_score \
+                    in zip(batch, pred_bboxes, pred_labels, pred_scores):
+
+                pred_bbox = chainer.cuda.to_cpu(pred_bbox)
+                pred_label = chainer.cuda.to_cpu(pred_label)
+                pred_score = chainer.cuda.to_cpu(pred_score)
+
+                out_file = self.filename.format(
+                    index=idx, iteration=trainer.updater.iteration)
```
yuyu2172: What if filename does not contain keywords `index` or `iteration`?

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -27,179 +22,56 @@ def _check_available():
...
-        Given the inputs from :meth:`dataset.__getitem__`, a method
-        :meth:`vis_transform` should convert them into visualizable forms.
-        The values returned by :meth:`vis_transform` should be
-
-        .. code:: python
-
-            img, bbox, label = vis_transform(inputs)
-
-        :obj:`img` should be an image which is in CHW format, BGR,
-        :math:`[0, 255]` and :obj:`dtype==numpy.uint8`.
```
yuyu2172: One summary sentence can be added like below.

Internally, this extension takes examples from an iterator, predict bounding boxes from the images in the examples, and visualizes them using `chainercv.visualizations.vis_bbox`.

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -27,179 +22,56 @@ def _check_available():
...
-        pred_bbox, pred_label = predict_func((img[None]))
-        pred_bbox = pred_bbox[0]  # (B, R, 4) -> (R, 4)
-        pred_label = pred_label[0]  # (B, R)  -> (R,)
-        vis_img, vis_bbox, vis_label = vis_transform(inputs)
+        batch = next(iterator)
+        # Convert batch -> imgs, gt_bboxes, gt_labels
+        pred_bboxes, pred_labels, pred_scores = target.predict(imgs)
         # Visualization code
-        # Uses (vis_img, vis_bbox, vis_label) as the ground truth output
-        # Uses (vis_img, pred_bbox, pred_label) as the predicted output
```
yuyu2172: Variables with single objects (e.g. `img, gt_bbox`) have not appeared previously in the doc.

Can you add something like

```
# Visualization code
for img, gt_bbox, gt_label, pred_bbox, pred_label, pred_score in ... :
    vis_bbox(img, gt_bbox, gt_label)
    vis_bbox(img, pred_bbox, pred_label, pred_score)
```

chainercv/extensions/detection/detection_vis_report.py
Hakuyume: This is the behaviour `chainer.training.extensions.Evaluator` does.

chainercv/extensions/detection/detection_vis_report.py
Hakuyume: This extension may raise an exception. Is it better to skip absent keywords?

Hakuyume: Sorry, it worked well without `index` or `iteration`.

yuyu2172: You are right.
I didn't know that it works.

Hakuyume: I didn't know, either. I added tests to confirm it works.

chainercv/extensions/detection/detection_vis_report.py
yuyu2172: I guess these evaluator like extensions are assumed to invoke `reset` at the beginning of `__call__`.

It looks fine now.

Hakuyume:
I fixed.

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -27,179 +22,69 @@ def _check_available():
...
-        .. code:: python
-
-            img, bbox, label = vis_transform(inputs)
-
-        :obj:`img` should be an image which is in CHW format, BGR,
-        :math:`[0, 255]` and :obj:`dtype==numpy.uint8`.
 
+    Internally, this extension takes examples from an iterator,
+    predict bounding boxes from the images in the examples,
+    and visualizes them using chainercv.visualizations.vis_bbox.
```
yuyu2172: Make the function words into a link by `:meth:`chainercv.visualizations.vis_bbox``

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -27,179 +22,69 @@ def _check_available():
...
+        batch = next(iterator)
+        # Convert batch -> imgs, gt_bboxes, gt_labels
+        pred_bboxes, pred_labels, pred_scores = target.predict(imgs)
         # Visualization code
-        # Uses (vis_img, vis_bbox, vis_label) as the ground truth output
-        # Uses (vis_img, pred_bbox, pred_label) as the predicted output
+        for img, gt_bbox, gt_label, pred_bbox, pred_label, pred_score \\
+                in zip(imgs,
+                       gt_boxes, gt_labels,
+                       pred_bboxes, pred_labels, pred_scores):
```
yuyu2172: How about making zip( ) into two lines?

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -68,7 +68,7 @@ class DetectionVisReport(chainer.training.extension.Extension):
         label_names (iterable of str): Name of labels ordered according
             to label_ids. If this is :obj:`None`, labels will be skipped.
         filename (str): Basename for the saved image. It can contain two
-            keywords, `'{iteration}'` and `'`{index}'`. They are replaced
+            keywords, `'{iteration}'` and `'{index}'`. They are replaced
             with the iteration of the trainer and the index of the sample
             when this extension save an image. The default value is
             `'detection_iter={iteration}_idx={index}.jpg'`.
```
yuyu2172: I like using :obj: instead of italic.
There are three words in this paragraph that are related.

chainercv/extensions/detection/detection_vis_report.py
```diff
@@ -27,179 +22,69 @@ def _check_available():
...
-            More detail can be found at the description on the step 3 of
-            internal mechanics found above.
+        :obj:`pred_score` is a float array of shape :math:`(R,)`.
+        Each score indicates how confident the prediction is.
 
+    Args:
+        iterator: Iterator object that produces images and ground truth.
+        target: Link object used for detection.
+        label_names (iterable of str): Name of labels ordered according
+            to label_ids. If this is :obj:`None`, labels will be skipped.
```
yuyu2172: s/label_ids/label ids

yuyu2172:
LGTM

PR #145, opened 2017-05-04
Add bbox_overlap

yuyu2172:
I added dependency to Cython.
I think this is a reasonable dependency because CuPy depends on Cython.

Also `links.rst` imports bbox_overlap in a same way VGG model is imported in Chainer's `link.rst`.


tests/links_tests/utils_tests/bbox_tests/test_bbox_overlap.py
```diff
@@ -0,0 +1,42 @@
...
+class TestBboxOverlap(unittest.TestCase):
+
+    def test_bbox_overlap(self):
+        bbox = np.array([[0, 0, 8, 8]], dtype=np.float32)
+        query_bbox = np.array([[3, 5, 10, 12], [9, 10, 11, 12]],
+                              dtype=np.float32)
+
+        overlap = bbox_overlap(bbox, query_bbox)
+
+        o0 = float(4 * 6) / (9 * 9 + 8 * 8 - 4 * 6)
```
Hakuyume: use `from __future__ import division` to force float division.

chainercv/links/utils/bbox/bbox_overlap.pyx
```diff
@@ -0,0 +1,78 @@
...
+    K = query_bbox.shape[0]
+    if bbox.shape[1] < 4 or query_bbox.shape[1] < 4:
+        raise IndexError
+
+    overlap  = np.zeros((N, K), dtype=DTYPE)
+
+    for k in range(K):
+        box_area = (
+            (query_bbox[k, 2] - query_bbox[k, 0] + 1) *
+            (query_bbox[k, 3] - query_bbox[k, 1] + 1)
```
Hakuyume: When the input values is continuous (`float`), the width is `xmax - xmin`. `+1` is only for discrete inputs (`int`).

Hakuyume:
I had implemented a very similar function in chainer-ssd (https://github.com/Hakuyume/chainer-ssd/blob/master/lib/iou.py)
Is the Cython version much better than this numpy version?

chainercv/links/utils/bbox/bbox_overlap.pyx
yuyu2172: Really? Adding +1 is used regardless of data type in Faster RCNN for VOC. I am not sure about other cases.

yuyu2172:
> I had implemented a very similar function in chainer-ssd (https://github.com/Hakuyume/chainer-ssd/blob/master/lib/iou.py)
> Is the Cython version much better than this numpy version?

I think numpy version is better for accepting arbitrary data type and extendable to CuPy.

chainercv/links/utils/bbox/bbox_overlap.pyx
yuyu2172: You are right.
It seems that definition of IoU is different between VOC evaluation code and COCO evaluation code.

COCO does use the convention you mentioned.
https://github.com/pdollar/coco/blob/master/common/maskApi.c#L109

Hakuyume: VOC assumes that the values of bounding boxes are discrete. It is better to design `bbox_overlap` for continuous values and adding `1` in the VOC evaluation function.

yuyu2172:
I will change code to NumPy code with support for CuPy.

yuyu2172:
Fixed.

chainercv/links/utils/bbox/bbox_overlap.py
```diff
@@ -0,0 +1,38 @@
...
+
+    """
+    if bbox.shape[1] < 4 or query_bbox.shape[1] < 4:
+        raise IndexError
+    xp = cuda.get_array_module(bbox)
+
+    # left top
+    lt = xp.maximum(bbox[:, None, :2], query_bbox[:, :2])
+    # right bottom
+    rb = xp.minimum(bbox[:, None, 2:4], query_bbox[:, 2:])
```
Hakuyume: `:4` is redundant because we assume the shape of bbox is `(R, 4)`. By the way, you forgot adding `:4` to `query_bbox[:, 2:]`.

chainercv/links/utils/bbox/bbox_overlap.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Returns:
+        ~numpy.ndarray:
+        An array whose shape is :math:`(N, K)`. \
+        An element at index :math:`(n, k)` contains overlap between \
+        :math:`n` th bounding box in :obj:`bbox` and :math:`k` th bounding \
+        box in :obj:`query_bbox`.
+
+    """
+    if bbox.shape[1] < 4 or query_bbox.shape[1] < 4:
```
Hakuyume: Are there any cases that `bbox.shape[1] > 4`?

yuyu2172: I thought I *can* support those cases, but never mind.

chainercv/links/utils/bbox/bbox_overlap.py
yuyu2172: Right.

chainercv/links/utils/bbox/bbox_overlap.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # left top
+    lt = xp.maximum(bbox[:, None, :2], query_bbox[:, :2])
+    # right bottom
+    rb = xp.minimum(bbox[:, None, 2:4], query_bbox[:, 2:])
+
+    area_i = xp.prod(rb - lt, axis=2) * (lt < rb).all(axis=2)
+    area_0 = xp.prod(bbox[:, 2:4] - bbox[:, :2], axis=1)
+    area_1 = xp.prod(query_bbox[:, 2:4] - query_bbox[:, :2], axis=1)
+    return area_i / (area_0[:, None] + area_1 - area_i)
```
Hakuyume: I don't like to use both `area_i` and `area_1` because they look very similar.

yuyu2172: How about `area_b` and `area_q`.

chainercv/links/utils/bbox/bbox_overlap.py
```diff
@@ -0,0 +1,42 @@
+from chainer import cuda
+
+
+def bbox_overlap(bbox, query_bbox):
```
Hakuyume: This function is very general. I think it is better to change the names of arguments more general. For example, 'bbox_a' and `bbox_b`. In this case, we can use `area_a` and `area_b` for the names of variables.

chainercv/links/utils/bbox/bbox_overlap.py
```diff
@@ -0,0 +1,42 @@
...
+    """Calculate overlap coefficients between bounding boxes.
+
+    The coefficient is calculated as a ratio of area of the intersection
+    and area of the union.
```
Hakuyume: This overlap is well known as `Jaccard overlap`.

chainercv/links/utils/bbox/bbox_overlap.py
```diff
@@ -0,0 +1,42 @@
...
+
+    This function accepts both :obj:`numpy.ndarray` and :obj:`cupy.ndarray` as
+    inputs.
```
Hakuyume: Please note that both of two arrays must be same type. For example, this function will not work when `bbox` is `numpy.ndarray` and `query_bbox` is `cupy.ndarray`.

Hakuyume:
This function is very common. It will be used not only from `links`. It is better to rethink the namespace. I know it is not good to move it into `trainsforms` because of its APIs. How about `chainercv.utils.bbox`?

yuyu2172:
Sure. 
That makes sense.
`non_maximum_suppression` can be moved there too.

yuyu2172:
I rebased  the master. The new commit is f19dc9.

chainercv/utils/bbox/bbox_overlap.py
```diff
@@ -0,0 +1,43 @@
...
+
+    """
+    if bbox_a.shape[1] != 4 or bbox_b.shape[1] != 4:
+        raise IndexError
+    xp = cuda.get_array_module(bbox_a)
+
+    # left top
+    lt = xp.maximum(bbox_a[:, None, :2], bbox_b[:, :2])
+    # right bottom
+    rb = xp.minimum(bbox_a[:, None, 2:], bbox_b[:, 2:])
```
Hakuyume: This is only a question. We can use `None` or `np.newaxis` for new axis. Which one is better?

docs/source/reference/index.rst
```diff
@@ -10,6 +10,7 @@ ChainerCV Reference Manual
    datasets
    evaluations
    extensions
+   links
```
Hakuyume: This modification should not be in this PR.

docs/source/reference/links.rst
```diff
@@ -0,0 +1,4 @@
+Links
+=====
+
+.. module:: chainercv.links
```
Hakuyume: ditto.

tests/utils_tests/bbox_tests/test_bbox_overlap.py
```diff
@@ -0,0 +1,62 @@
...
+from chainer.testing import attr
+
+from chainercv.utils import bbox_overlap
+
+
+class TestBboxOverlap(unittest.TestCase):
+
+    def setUp(self):
+        self.bbox = np.array([[0, 0, 8, 8]], dtype=np.float32)
+        self.query_bbox = np.array([[3, 5, 10, 12], [9, 10, 11, 12]],
```
Hakuyume: Please change these names to `bbox_a` and `bbox_b`.

tests/utils_tests/bbox_tests/test_bbox_overlap.py
```diff
@@ -0,0 +1,62 @@
...
+import numpy as np
+
+from chainer import cuda
+from chainer import testing
+from chainer.testing import attr
+
+from chainercv.utils import bbox_overlap
+
+
+class TestBboxOverlap(unittest.TestCase):
```
Hakuyume: Please add test case in which `N == 0` and `N > 2`. It may work well.

chainercv/utils/bbox/bbox_overlap.py
yuyu2172: I do not have preference.
Do you?

I like `None` because its shorter. Also, I hesitate to use `xp.newaxis`, which is what I need to use in this case.

But, `np.newaxis` is somewhat more explicit.

yuyu2172:
Thanks. I fixed the code.
I am working on docs. Please wait.

chainercv/utils/bbox/bbox_overlap.py
Hakuyume: I usually use `np.newaxis` even in `cupy.ndarray` because I think it is explicit and better. In Chainer, they use both styles.

yuyu2172:
I am done with changing docs.
Could you check it?

chainercv/utils/bbox/bbox_overlap.py
yuyu2172: I think either is OK.

tests/utils_tests/bbox_tests/test_bbox_overlap.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_bbox_overlap_cpu(self):
+        self.check(self.bbox_a, self.bbox_b, self.expected)
+
+    @attr.gpu
+    def test_bbox_overlap_gpu(self):
+        self.check(
+            cuda.to_gpu(self.bbox_a),
+            cuda.to_gpu(self.bbox_b),
+            cuda.to_gpu(self.expected))
```
Hakuyume: Do you need to send `expected` to gpu?

tests/utils_tests/bbox_tests/test_bbox_overlap.py
```diff
@@ -0,0 +1,71 @@
...
+class TestBboxOverlapInvalidShape(unittest.TestCase):
+
+    def test_bbox_overlap_invalid_bbox(self):
+        bbox_a = np.array([[0, 0, 8]], dtype=np.float32)
+        bbox_b = np.array([[1, 1, 9, 9]], dtype=np.float32)
+
+        with self.assertRaises(IndexError):
+            bbox_overlap(bbox_a, bbox_b)
+
+    def test_bbox_overlap_invalid_query_bbox(self):
```
Hakuyume: Please rename this function. And please add test case in which `bbox.shape[1] > 4`.

tests/utils_tests/bbox_tests/test_bbox_overlap.py
yuyu2172: I just wanted the check function to assume that all inputs are same type, which seems to follow the convention in Chainer's tests (e.g. https://github.com/pfnet/chainer/blob/master/tests/chainer_tests/functions_tests/array_tests/test_split_axis.py)

tests/utils_tests/bbox_tests/test_bbox_overlap.py
yuyu2172: Right. Totally missed that.

tests/utils_tests/bbox_tests/test_bbox_overlap.py
Hakuyume: How about replacing `self.assertIsInstance(overlap, xp.ndarray)` with `self.assertIsInstance(overlap, type(expected)` ?

yuyu2172: Sure. It gets simpler.

Hakuyume:
LGTM

PR #146, opened 2017-05-04
Add bbox_regression_target

yuyu2172:
~~Merge after #145.~~
Merge after #148.


yuyu2172:
I found that original implementation had some bug.
It was raised in the original Github page https://github.com/rbgirshick/py-faster-rcnn/issues/333 (but with no answer from the author).

yuyu2172:
I changed convention to one without offseting bboxes by +1.

yuyu2172:
I am trying to come up with a better name.
I think `bbox2delta` and `delta2bbox` would be a good alternative.

@hakuyume do you think this is good?

Hakuyume:
@yuyu2172 I think `delta` is too ambiguous. How about `encode_bbox_regression` and `decode_bbox_regression`? (Are they too long?)

yuyu2172:
I agree with using `encode` and `decode` is good.
I found that `bbox_regression` is not a good word because this function is more general than a particular formulation of a loss.
How about `encode_delta` and `decode_delta`?

Also, this PR became obsolete. I merged changes to #153. Let's discuss there.

PR #147, opened 2017-05-04
Delete semantic_segmentation_vis_report

yuyu2172:
The interface of VisReports are changing rapidly.
I think it is better to restrict our focus on DetectionVisReport.


yuyu2172:
@Hakuyume 
Can you merge this and #133?

We should concentrate on detection.

Hakuyume:
I agree we should concentrate on detection.

PR #148, opened 2017-05-04
Add generate_anchor_base

yuyu2172:
This implementation is different from the original one in that it does not offset bounding boxes by `+1`.
This change is consistent with other bounding box functions.


yuyu2172:
Current implementation rounds all coordinates to integers.
I think this constraint can be safely removed.
Or, if keeping it, this should be documented.

yuyu2172:
Integrated to #153

PR #150, opened 2017-05-06
Add non_maximum_suppression that supports both CPU and GPU arrays.

yuyu2172:
Merge after #145.

This PR moves `transforms/bbox/non_maximum_suppression.py` to `links/utils/bbox/non_maximum_suppression.py`.

Here are few changes.

+ It supports both CuPy arrays and NumPy arrays. Supporting CuPy is very critical for training Faster RCNN. Usually, NumPy implementation (not the one in this PR, but Cythonized version) is 10x slower than CuPy implementation. Not only that, the magnitude of speed is quite big taking 200ms per iteration for NumPy  version. 
+ Change interface of `non_maximum_suppression` to not return transformed bounding boxes.
+ Add `score` option.
+ Add Cython dependencies. The Cythonized part is important for performance. Plain Python is order of magnitude slower.


yuyu2172:
@Hakuyume 

Can you check this PR?

yuyu2172:
Updated directory structure and rebased dependencies.

Hakuyume:
I'm checking this PR. Please wait.

tests/utils_tests/bbox_tests/test_non_maximum_suppression.py
```diff
@@ -0,0 +1,141 @@
...
+            self.threshold,
+            cuda.to_gpu(self.expect)
+        )
+
+
+class TestNonMaximumSuppressionConsistency(unittest.TestCase):
+
+    @attr.gpu
+    def test_non_maximum_suppression_consistency(self):
+        bbox = _generate_bbox(6000, (600, 800), 32, 512)
```
Hakuyume: `_generate_bbox` is used only here. It is better to merge it into `setUp`.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 chainer >= 1.21
+cython
```
Hakuyume: Please add note to `README.md`.

chainercv/utils/bbox/non_maximum_suppression.py
```diff
@@ -0,0 +1,202 @@
...
+    return selec
+
+
+nms_gpu_code = '''
+#include <vector>
+
+#define DIVUP(m,n) ((m) / (n) + ((m) % (n) > 0))
+int const threadsPerBlock = sizeof(unsigned long long) * 8;
+
+__device__ inline float devIoU(float const * const a, float const * const b) {
```
Hakuyume: Inconsistency between `IoU` and `jaccard`

chainercv/utils/bbox/non_maximum_suppression.py
```diff
@@ -0,0 +1,202 @@
...
+    bbox_area = np.prod(bbox[:, 2:] - bbox[:, :2], axis=1)
+    if score is not None:
+        order = score.argsort()[::-1]
+        bbox = bbox[order]
+
+    selec = np.zeros(bbox.shape[0], dtype=bool)
+    for i, b in enumerate(bbox):
+        lt = np.maximum(b[:2], bbox[selec, :2])
+        rb = np.minimum(b[2:], bbox[selec, 2:])
+        area = np.prod(rb - lt, axis=1) * (lt < rb).all(axis=1)
```
Hakuyume: I think this is more simple.
```
for i in score.argsort()[::-1]:
    lt = np.maximum(bbox[i, :2], bbox[selec, :2])
    rb = np.minimum(bbox[i, 2:], bbox[selec, 2:])
...
return selec.nonzero().astype(np.int32)
```

chainercv/utils/bbox/non_maximum_suppression.py
```diff
@@ -0,0 +1,202 @@
...
+    The bounding boxes are selected from ones with higher scores.
+    If :obj:`score` is not provided as an argument, the bounding box
+    is ordered by its index in ascending order.
+
+    The bounding boxes are expected to be packed into a two dimensional
+    tensor of shape :math:`(R, 4)`, where :math:`R` is the number of
+    bounding boxes in the image. The second axis represents attributes of
+    the bounding box. They are :obj:`(x_min, y_min, x_max, y_max)`,
+    where the four attributes are coordinates of the bottom left and the
+    top right vertices.
```
Hakuyume: I think `x_min, y_min` is coordinates of the top left. This may be my mistake.

chainercv/utils/bbox/non_maximum_suppression.py
```diff
@@ -0,0 +1,202 @@
...
+        thresh (float): Thresold of Jaccard overlap.
+        score (array): An array of confidences whose shape is :math:`(R,)`.
+        limit (int): The upper bound of the number of the output bounding
+            boxes. If it is not specified, this method selects as many
+            bounding boxes as possible.
+
+    Returns:
+        array:
+        An array with indices of bounding boxes that are selected. \
+        The shape of this array is :math:`(R,)` and its dtype is\
```
Hakuyume: :math:`(R,)` looks the length of the output is same as that of input. How about `(K,)` or `(R',)`? It is better to note `R' <= R`.

tests/utils_tests/bbox_tests/test_non_maximum_suppression.py
```diff
@@ -0,0 +1,141 @@
...
+            (0, 0, 4, 4),
+            (1, 1, 5, 5),  # 9/23
+            (2, 1, 6, 5),  # 6/26, 12/20
+            (4, 0, 8, 4),  # 0/32, 3/29, 6/26
+        ))
+
+    def check_non_maximum_suppression(self, bbox, threshold, expect):
+        xp = cuda.get_array_module(bbox)
+        selec = non_maximum_suppression(bbox, threshold)
+        self.assertIsInstance(selec, xp.ndarray)
```
Hakuyume: https://github.com/pfnet/chainercv/pull/145/files#r115136375

tests/utils_tests/bbox_tests/test_non_maximum_suppression.py
```diff
@@ -0,0 +1,141 @@
...
+        # Pass all options to the tested function
+        scored_selec = non_maximum_suppression(bbox, threshold, score, limit)
+        self.assertIsInstance(scored_selec, xp.ndarray)
+
+        # Reorder inputs befor passing it to the function.
+        # Reorder the outputs according to scores.
+        # CuPy does not support argsort
+        order = cuda.to_cpu(score).argsort()[::-1]
+        reordered_selec = non_maximum_suppression(
+            bbox[order], threshold, score=None, limit=limit)
```
Hakuyume: It is better to pass `limit=None` and truncate the returned array.

tests/utils_tests/bbox_tests/test_non_maximum_suppression.py
```diff
@@ -0,0 +1,141 @@
...
+    def setUp(self):
+        self.bbox = np.zeros((0, 4))
+
+    def check_non_maximum_suppression_zero_legnth_bbox(
+            self, bbox, threshold):
+        xp = cuda.get_array_module(bbox)
+        selec = non_maximum_suppression(bbox, threshold)
+        self.assertIsInstance(selec, xp.ndarray)
+        np.testing.assert_equal(
+            cuda.to_cpu(selec), np.zeros((0,), dtype=np.int32))
```
Hakuyume: Testing `selec.shape` is enough.

chainercv/utils/bbox/non_maximum_suppression.py
yuyu2172: Since `selec.nonzero()` returns indices that are not in order with `score.argsort()`, it can not be written as simple as you wrote.
I think reordering `bbox` before for loop is easier to read.

tests/utils_tests/bbox_tests/test_non_maximum_suppression.py
yuyu2172: I had to use `_generate_bbox` in two places, so I left it there.

yuyu2172:
Thanks.
I found a bug in CPU mode when size of bbox is big and score is given.
I changed a test to catch this case b60ad48 and fixed it 4f8caca.

Other than that, I reflected your comments.

yuyu2172:
Jaccard overlap seems to be not a valid word since I could not find a usage of that word on any of the Google search results.

How about using Intersection over Union (IoU)?
Wikipedia says it is one of the correct word.

> The Jaccard index, also known as Intersection over Union and the Jaccard similarity coefficient (From Wikipedia)

Hakuyume:
@yuyu2172 Sorry, `Jaccard overlap` is not a valid word. `Jaccard index` is valid. In the area of object detection, `IoU (Intersection over Union)` is more familiar. I think `IoU` is OK. Of course, we have to keep consistency.

chainercv/utils/bbox/non_maximum_suppression.py
Hakuyume: I thought this function doesn't need to sort the indices according to their scores. The doc says `An array with indices of bounding boxes that are selected`. If we have to return sorted indices, sorting `bbox` before is good as you said.

tests/utils_tests/bbox_tests/test_non_maximum_suppression.py
Hakuyume: I see.

yuyu2172:
I reflected our discussions to the code.

chainercv/utils/bbox/non_maximum_suppression.py
```diff
@@ -0,0 +1,205 @@
...
+
+    This function accepts both :obj:`numpy.ndarray` and :obj:`cupy.ndarray` as
+    inputs. Please note that both :obj:`bbox` and :obj:`score` need to be
+    same type.
+    The output is same type as the type of the inputs.
+
+    Args:
+        bbox (array): Bounding boxes to be transformed. The shape is
+            :math:`(R, 4)`. :math:`R` is the number of bounding boxes.
+        thresh (float): Thresold of IoU.
```
Hakuyume: Typo, `threshold`.

chainercv/utils/bbox/non_maximum_suppression.py
```diff
@@ -0,0 +1,205 @@
...
+        kernel_code = cupy.cuda.compile_with_cache(code, options=options)
+        return kernel_code.get_function(kernel_name)
+
+
+def non_maximum_suppression(bbox, thresh, score=None,
+                            limit=None):
+    """Suppress bounding boxes according to their IoUs.
+
+    This method checks each bounding box sequentially and selects the bounding
+    box if the Intersection over Union (IoU) between the bounding box and the
```
Hakuyume: IoU**s**? You are using `IoU` as a countable noun in this doc.

yuyu2172:
Fixed.

Hakuyume:
LGTM

PR #151, opened 2017-05-06
Add ProposalCreator

yuyu2172:
Depends on #146 and #150.

- [x] docs
- [x] tests


yuyu2172:
Since it is confusing to separate PRs into multiple ones, I will close this.

PR #153, opened 2017-05-06
Add test code for Faster RCNN

yuyu2172:
Merge after #148, #146

Adds

+ encode_delta.py
+ decode_delta.py
+ generate_anchor_base.py
+ faster_rcnn.py
+ faster_rcnn_vgg.py
+ region_proposal_network.py
+ proposal_creator.py


yuyu2172:
In the current design, I made `__call__` to return intermediate representations.
It is helpful to make FRCNN chain return intermediate features for reusability. On top of that, this is useful concretely for the following cases.

+ In training mode, the forward computation is different from the test mode computation in this PR. In particular, computation is different after RPN. Therefore, inside the loss, I used `__call__(x, layers=['roi', ...])` to get RPN outputs without computing following features like `score`.

On the design of `__call__`, I decided to return features produced by chains at the second highest abstraction level.
This means that collectible outputs are features produced by Feature Extractor, RPN and Head.
On the other hand, intermediate features such as `conv3` in Feature Extractor can not be collected.
I think this is a good balance between complexity of `__call__` and re-usability.
In Chainer's ResNet implementation, similar design is used.

yuyu2172:
On class design of FRCNN, I separated feature extractors with other links.
As discussed in #122, class design of Detection links are under discussion.

yuyu2172:
@Hakuyume  @mitmul  
Can you comment on the design of  FRCNN links?

+ https://github.com/yuyu2172/chainercv/blob/faster-rcnn-test/chainercv/links/faster_rcnn/faster_rcnn.py
+ https://github.com/yuyu2172/chainercv/blob/faster-rcnn-test/chainercv/links/faster_rcnn/faster_rcnn_vgg.py

yuyu2172:
I am reworking on design.

yuyu2172:
Done with design.

Here is a snippet to run a demo.

```
cd examples/faster_rcnn
python demo GPU_ID IMAGE_PATH
```

You can try with this demo image.


The output 
figure_1-1

yuyu2172:
Notations I have used

+ `anchor_base` for anchors generated at each pixel. (9 in this case)
+ `anchor` for entire collection of anchors shifted across pixels. (total of 9 * feature_size)

+ `raw_bbox` to refer to bounding boxes with `x_min, y_min, x_max, y_max`.
+ `bbox` to refer to offsets and scales of bounding boxes.

yuyu2172:
The performance of the current pretrained model is 69.9 mAP on VOC test.

I had two implementations. One offsets `x_max` and `y_max` by +1, and the other do not offset.
In the former implementation, the performance fluctuated between 69.1 and 70.5 mAP due to randomness in training. In the original paper, score was 69.9 mAP, but some other Faster RCNN implementations claim 70.0 mAP or 70.2 mAP.

I have only trained once with the latter implementation, so I need to check if it reaches higher scores with different random seeds with more experiments.

yuyu2172:
Some definition of weights have changed for consistency.
The new weights have been uploaded again.

The updated list of notations I have used

+ When `roi` is on prefix (e.g. `roi_scores`), the objects from different batches are concatenated on the first axis. Otherwise, batched values are stacked on the first axis to have shapes like `N, C, H, W`.
+ `loc` is for offsets/scales from base anchors. `bbox` is coordinates of left top and right bottom vertices. `cls_loc` is `loc` for multiple classes and its shape is `R, n_class * 4`.

chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py
```diff
@@ -0,0 +1,264 @@
...
+
+        fc6 = F.dropout(_relu(self.fc6(pool)), train=False)
+        fc7 = F.dropout(_relu(self.fc7(fc6)), train=False)
+        roi_cls_locs = self.cls_loc(fc7)
+        roi_scores = self.score(fc7)
+        return roi_cls_locs, roi_scores
+
+
+class VGG16FeatureExtractor(chainer.Chain):
+    """Truncated VGG that extracts an conv5_3 features.
```
Hakuyume: `a conv5_3 feature` ?

chainercv/links/model/faster_rcnn/faster_rcnn.py
```diff
@@ -0,0 +1,321 @@
...
+    2. **Region Proposal Networks**: Given the feature maps calculated in \
+        the previous stage, produce set of RoIs around objects.
+    3. **Localization and Classification Heads**: Using features that belong \
+        to the proposed RoIs, classify the categories of the objects in the \
+        RoIs and improve localizations.
+
+    Each stage is carried out by one of the callable
+    :class:`chainer.Chain` objects :obj:`feature`, :obj:`rpn` and :obj:`head`.
+
+    There are two functions :func:`predict` and :func:`__call__` to conduct
```
Hakuyume: `:func:` -> `:meth:`

chainercv/links/model/faster_rcnn/faster_rcnn.py
```diff
@@ -0,0 +1,321 @@
...
+    the same interface. Please refer to :func:`FasterRCNN.predict` for
+    further details.
+
+    .. [#] Shaoqing Ren, Kaiming He, Ross Girshick, Jian Sun. \
+    Faster R-CNN: Towards Real-Time Object Detection with \
+    Region Proposal Networks. NIPS 2015.
+
+    Args:
+        extractor (callable Chain): A callable that takes a BCHW image
+            array and returns a feature.
```
Hakuyume: `feature` -> `feature map`

chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py
```diff
@@ -0,0 +1,291 @@
...
+
+def _relu(x):
+    # use_cudnn = False is sometimes x3 faster than otherwise.
+    # This will be the default mode in Chainer v2.
+    return F.relu(x, use_cudnn=False)
+
+
+class FasterRCNNVGG16(FasterRCNN):
+
+    """Faster R-CNN based on VGG16.
```
Hakuyume: `VGG16` -> `VGG-16`

chainercv/links/model/faster_rcnn/faster_rcnn.py
```diff
@@ -0,0 +1,321 @@
...
+        img = (img - self.mean).astype(np.float32, copy=False)
+        return img
+
+    def _suppress(self, raw_cls_bbox, raw_prob):
+        bbox = list()
+        label = list()
+        score = list()
+        # skip cls_id = 0 because it is the background class
+        for l in range(1, self._n_class):
+            cls_bbox_l = raw_cls_bbox[:, l * 4: (l + 1) * 4]
```
Hakuyume: Can you write `[:, l * 4: (l + 1) * 4]` using `reshape`?

chainercv/links/model/faster_rcnn/faster_rcnn.py
```diff
@@ -0,0 +1,321 @@
...
+    Args:
+        extractor (callable Chain): A callable that takes a BCHW image
+            array and returns a feature.
+        rpn (callable Chain): A callable that has the same interface as
+            :class:`chainercv.links.RegionProposalNetwork`. Please refer to
+            the documentation found there.
+        head (callable Chain): A callable that takes
+            a BCHW array, RoIs and batch indices for RoIs. This returns class
+            dependent localization paramters and class scores.
+        n_fg_class (int): The number of classes excluding the background.
```
Hakuyume: We can get the number of classes from `head`. This argument looks redundant and may cause mismatch issue.

chainercv/links/model/faster_rcnn/faster_rcnn.py
```diff
@@ -0,0 +1,321 @@
...
+            x (~chainer.Variable): 4D image variable.
+            scale (float): Amount of scaling applied to the raw image
+                during preprocessing.
+            test (bool): If :obj:`True`, the test time behavior is used.
+
+        Returns:
+            Variable, Variable, array, array:
+            Returns tuple of four values listed below.
+
+            * **roi_cls_locs**: Offsets and scalings for the proposed RoIs. \
```
Hakuyume: What does `_cls_` in `roi_cls_locs` mean?

chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py
```diff
@@ -0,0 +1,291 @@
...
+            mean=np.array([102.9801, 115.9465, 122.7717],
+                          dtype=np.float32)[:, None, None],
+            min_size=min_size,
+            max_size=max_size
+        )
+
+        if pretrained_model in self._models:
+            data_root = get_dataset_directory('pfnet/chainercv/models')
+            url = self._models[pretrained_model]['url']
+            fn = url.rsplit('/', 1)[-1]
```
Hakuyume: `os.path.basename`?

chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py
```diff
@@ -0,0 +1,291 @@
...
+            'n_fg_class': 20,
+            'url': 'https://github.com/yuyu2172/share-weights/releases/'
+            'download/0.0.1/faster_rcnn_vgg_voc07.npz'
+        }
+    }
+    feat_stride = 16
+
+    def __init__(self,
+                 n_fg_class=None,
+                 pretrained_model='voc07',
```
Hakuyume: I prefer `pretrained_model=None` because we can add more pretraind models in the future.

examples/faster_rcnn/demo.py
```diff
@@ -0,0 +1,35 @@
...
+
+from chainercv.datasets.pascal_voc import voc_utils
+from chainercv.links import FasterRCNNVGG16
+from chainercv import utils
+from chainercv.visualizations import vis_bbox
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('gpu', type=int)
```
Hakuyume: https://github.com/pfnet/chainercv/pull/119#discussion_r117867030

chainercv/links/model/faster_rcnn/faster_rcnn.py
```diff
@@ -0,0 +1,321 @@
...
+            # clip bounding box
+            cls_bbox[:, slice(0, 4, 2)] = self.xp.clip(
+                cls_bbox[:, slice(0, 4, 2)], 0, W / scale)
+            cls_bbox[:, slice(1, 4, 2)] = self.xp.clip(
+                cls_bbox[:, slice(1, 4, 2)], 0, H / scale)
+
+            prob = F.softmax(roi_score).data
+
+            raw_cls_bbox = cuda.to_cpu(cls_bbox)
+            raw_prob = cuda.to_cpu(prob)
```
Hakuyume: If we send `cls_bbox` and `prob` to CPU, we cannot use GPU for NMS in _suppress. Is it OK?

Hakuyume: However, it is a good idea to return numpy arrays from `predict`.

chainercv/links/model/faster_rcnn/faster_rcnn.py
yuyu2172: `loc` is (R, 4)
`cls_loc` is (R, (L+1) * 4)

chainercv/links/model/faster_rcnn/faster_rcnn.py
yuyu2172: I found that GPU NMS is not as relatively fast in this part of the code because the number of boxes is small.

I think forcing CPU is fine. For better performance, we should cythonize CPU NMS.

Hakuyume: I see. I think current design is OK.

chainercv/links/model/faster_rcnn/faster_rcnn.py
Hakuyume: I see. Why don't you use `(R, L+1, 4)` instead?

chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py
```diff
@@ -0,0 +1,291 @@
...
+        proposal_creator_params (dict): Key valued paramters for
+            :obj:`chainercv.links.model.faster_rcnn.ProposalCreator`.
+
+    """
+
+    _models = {
+        'voc07': {
+            'n_fg_class': 20,
+            'url': 'https://github.com/yuyu2172/share-weights/releases/'
+            'download/0.0.1/faster_rcnn_vgg_voc07.npz'
```
Hakuyume: Is there a VGG-19 based Faster R-CNN? If so, `faster_rcnn_vgg16_voc07.npz' is better.

yuyu2172: OK!

Hakuyume:
LGTM

PR #165, opened 2017-05-17
Use split instead of mode as a keyword of datasets

### LINKED ISSUE #131: Argument for dataset to determine split is vague
Currently an option to determine train/val/test split of a dataset is determined by a keyword `mode`.
I think this is vague, and possibly be improved.

One option is to use a keyword `split`.

yuyu2172:
Fix #131.

### DIFF
--- a/chainercv/datasets/online_products/online_products_dataset.py
+++ b/chainercv/datasets/online_products/online_products_dataset.py
@@ -37,7 +37,7 @@ class OnlineProductsDataset(chainer.dataset.DatasetMixin):
     Images are in BGR and CHW format.
     Class ids start from 0.
 
-    The :obj:`mode` selects train and test split of the dataset as done in
+    The :obj:`split` selects train and test split of the dataset as done in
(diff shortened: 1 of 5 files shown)


chainercv/datasets/online_products/online_products_dataset.py
```diff
@@ -49,20 +49,20 @@ class OnlineProductsDataset(chainer.dataset.DatasetMixin):
         data_dir (string): Path to the root of the training data. If this is
             :obj:`auto`, this class will automatically download data for you
             under :obj:`$CHAINER_DATASET_ROOT/pfnet/chainercv/online_products`.
-        mode ({'train', 'test'}): Mode of the dataset.
+        split ({'train', 'test'}): split of the dataset.
```
Hakuyume: How about `Select a split of the dataset`? Anyway, the sentence should start with upper case.

yuyu2172:
Thanks. I fixed the code.

PR #167, opened 2017-05-22
Change VOCDetectionDataset's label names

yuyu2172:
Detection's label names do not include background anymore.
Also, I changed `dataset.labels` to be imported directly.


Hakuyume:
I think we have to modify `VOCSemanticSegmentationDataset` at the same time.

Hakuyume:
Sorry, it was my mistake. LGTM

PR #172, opened 2017-05-23
Consistency of citations

Hakuyume:


yuyu2172:
Why do you put references above arguments in some and below in the others?

For instance, `pca_lightning` puts references above its arguments.

yuyu2172:
Personally, references should be put above the arguments because that is how the compiled doc looks like.

Hakuyume:
> Why do you put references above arguments in some and below in the others?

It was just my mistake. I tried to put all references below arguments.

> In my opinion, references should be put above the arguments because that is how the compiled doc looks like.

OK, I will change.

yuyu2172:
LGTM

PR #177, opened 2017-05-23
Add tests for VOCDetectionDataset

yuyu2172:
This test does not check the data downloader.


tests/datasets_tests/test_pascal_voc/test_voc_detection_dataset.py
```diff
@@ -0,0 +1,71 @@
...
+            self.assertEqual(img.dtype, np.float32)
+            self.assertEqual(img.shape[0], 3)
+            self.assertEqual(img.ndim, 3)
+            self.assertGreaterEqual(np.min(img), 0)
+            self.assertLessEqual(np.max(img), 255)
+
+            self.assertIsInstance(bbox, np.ndarray)
+            self.assertEqual(bbox.dtype, np.float32)
+            self.assertEqual(bbox.ndim, 2)
+            self.assertEqual(bbox.shape[1], 4)
```
Hakuyume: How about checking if `x_min < x_max`?

chainercv/datasets/pascal_voc/voc_detection_dataset.py
```diff
@@ -61,6 +61,8 @@ class VOCDetectionDataset(chainer.dataset.DatasetMixin):
         return_difficult (bool): If true, this dataset returns a boolean array
             that indicates whether bounding boxes are labeled as difficult
             or not. The default value is :obj:`False`.
+            If :obj:`use_difficult` is :obj:`False`, this has to be
+            :obj:`False`.
```
Hakuyume: How about returning a false array as `difficult` when `use_difficult=False and return_difficult=True`?

tests/datasets_tests/test_pascal_voc/test_voc_detection_dataset.py
```diff
@@ -0,0 +1,71 @@
...
+        self.dataset = VOCDetectionDataset(
+            split=self.split,
+            year=self.year,
+            use_difficult=self.use_difficult,
+            return_difficult=self.return_difficult)
+        self.n_out = 4 if self.return_difficult else 3
+
+    @attr.slow
+    def test_get_example(self):
+        for i in np.random.randint(0, len(self.dataset), size=(10,)):
```
Hakuyume: We can use `chainer.testing.condition.repeat`.

yuyu2172:
The maximum coordinates in bounding boxes should be smaller than height and width of an image.
These changes will be reflected after #246.

chainercv/datasets/voc/voc_detection_dataset.py
```diff
@@ -124,7 +125,12 @@ def get_example(self, i):
             label.append(voc_utils.voc_detection_label_names.index(name))
         bbox = np.stack(bbox).astype(np.float32)
         label = np.stack(label).astype(np.int32)
-        difficult = np.array(difficult, dtype=np.bool)
+        if self.use_difficult:
```
Hakuyume: This `if` statement is unnecessary.

yuyu2172: Are you saying that `np.array(difficult, dtype=bool)` should be called after `if return_difficult`?

Hakuyume: No, I mean that `difficult = np.array(difficult, dtype=np.bool)` is enough. If `use_difficult` is False, `difficult` becomes `[False, False, ... ]` because it skips True GTs (https://github.com/yuyu2172/chainercv/blob/52714429cc7b87c12a5f40a49f8a58c77149c66e/chainercv/datasets/voc/voc_detection_dataset.py#L116-L117). So, we can use `difficult = np.array(difficult, dtype=np.bool)` to make a False array.

yuyu2172:
Perhaps, we can add a utility function `check_valid_bbox(bbox, H, W)`, which runs the check done in this test.
However, I am not sure how to call `unittest.TestCase.assertEqual` outside of the class `unittest.TestCase`.
One workaround is to use `np.testing.assert_equal`.

@Hakuyume 
What do you think about making the utility function inside this PR?

Hakuyume:
> @Hakuyume
What do you think about making the utility function inside this PR?

I'm working on assertion utilities in #278.

yuyu2172:
@Hakuyume 

Nice. Should I close this PR?

Hakuyume:
> Nice. Should I close this PR?

No. I want to merge this PR before mine. Now I'm checking this PR. So could you leave this PR open?

yuyu2172:
OK

PR #185, opened 2017-05-24
Add eval examples

Hakuyume:


yuyu2172:
Why did you put the function under `utils/testing/`?
I do not think this is only for testing.

Do you think `evaluations/eval_detction_mode.py` is a bad idea?

Also, `eval.py` can be changed to take both Faster R-CNN and SSD.

Hakuyume:
> Why did you put the function under utils/testing/?
I do not think this is only for testing.

As you pointed out , this not only for testing. I agree this function should be moved.

> Do you think evaluations/eval_detction_mode.py is a bad idea?

I think `evaluations.eval_detection_link_voc` is a good idea. However, `apply_detection_link` is not only for evaluation (for example, it can be used for visualization and testing images without ground truth). So I don't want to put `apply_detection_link` under `evaluations`.

Hakuyume:
> Also, eval.py can be changed to take both Faster R-CNN and SSD.

Yes, I think this is a good idea. How about putting `eval.py` and `visualize_detection_models.py` into `examples/detection`?

yuyu2172:
> I think evaluations.eval_detection_link_voc is a good idea. However, apply_detection_link is not only for evaluation (for example, it can be used for visualization and testing images without ground truth). So I don't want to put apply_detection_link under evaluations.

How about `chainercv/iterators/apply_detection_link`.
I think there will be more functions that take iterators as inputs.

yuyu2172:
> Yes, I think this is a good idea. How about putting eval.py and visualize_detection_models.py into examples/detection?

Yeah. I think that is good.

Hakuyume:
> How about chainercv/iterators/apply_detection_link.
I think there will be more functions that take iterators as inputs.
Also, this directory will imply that the detecion_link is going to be applied on iterators.

In Chainer, `chainer/iterators` is a collection of `Iterator`s. How about `chainercv/utils/iterator`?

yuyu2172:
> In Chainer, chainer/iterators is a collection of Iterators. How about chainercv/utils/iterator?

That sounds good!

chainercv/utils/iterator/apply_detection_link.py
```diff
@@ -0,0 +1,78 @@
...
+    into three lists,
+    :obj:`pred_bboxes`, :obj:`pred_labels` and :obj:`pred_scores`.
+    This function also stacks the values returned by the iterator.
+    These values can be used for evaluation.
+
+    Args:
+        target (chainer.Link): An detection link. This link must have
+            :meth:`predict` method which take a list of images and returns
+            :obj:`bboxes`, :obj:`labels` and :obj:`scores`.
+        iterator (chainer.Iterator): An iterator. Each samples must have
```
yuyu2172: s/samples/sample

By the way, I think "should" sounds friendlier than "must".

chainercv/utils/iterator/apply_detection_link.py
```diff
@@ -0,0 +1,78 @@
...
+            an image for its first element. This image is passed to
```
yuyu2172: an image as its first elemnt

chainercv/utils/iterator/apply_detection_link.py
```diff
@@ -0,0 +1,78 @@
...
+            :obj:`target`. The rests are stacked into :obj:`gt_values`.
+        hook: An callable which called after each iteration.
```
yuyu2172: which is

chainercv/utils/iterator/apply_detection_link.py
```diff
@@ -0,0 +1,78 @@
...
+            :obj:`pred_bboxes`, :obj:`pred_labels`, :obj:`pred_scores` and
+            :obj:`gt_values` are passed. Note that these values do not contain
```
yuyu2172: are passed as arguments

chainercv/utils/iterator/apply_detection_link.py
```diff
@@ -0,0 +1,78 @@
...
+            the data of previous iterations.
```
yuyu2172: data from the previous iterations

chainercv/utils/iterator/apply_detection_link.py
```diff
@@ -0,0 +1,78 @@
...
+
+    returns:
```
yuyu2172: s/returns/Returns/

chainercv/utils/iterator/apply_detection_link.py
```diff
@@ -0,0 +1,78 @@
...
+        tuple of list:
+        This function returns four lists, :obj:`pred_bboxes`,
```
yuyu2172: s/four lists,/four lists:

examples/detection/README.md
```diff
@@ -0,0 +1,23 @@
...
+- SSD512
+
+For the details, please check the documents and examples of each model.
+
+## Performance
+
+### PASCAL VOC2007 Test
+
+| Model | FPS | mAP (PASCAL VOC2007 metric) |
+|:-:|:-:|:-:|
```
yuyu2172: perhaps, it is better to add a column on training data.

yuyu2172:
How about adding a doc rst

yuyu2172:
Some grammar comments are not reflected yet.

chainercv/utils/iterator/apply_detection_link.py
```diff
@@ -0,0 +1,78 @@
...
+        iterator (chainer.Iterator): An iterator. Each sample should have
+            an image as its first element. This image is passed to
+            :obj:`target`. The rests are stacked into :obj:`gt_values`.
+        hook: An callable which is called after each iteration.
+            :obj:`pred_bboxes`, :obj:`pred_labels`, :obj:`pred_scores` and
+            :obj:`gt_values` are passed. Note that these values do not contain
+            the data of previous iterations.
+
+    Returns:
+        tuple of list:
```
yuyu2172: tuple of lists

yuyu2172:
LGTM

PR #188, opened 2017-05-24
Move pascal_voc --> voc

yuyu2172:
Change the directory name because `pascal_` is redundant.


Hakuyume:
Currently, ChainerCV downloads the dataset under `.chainer/dataset/pfnet/chainercv/pascal_voc/`.
For consistency, is it better to change the destination?

yuyu2172:
Right. Thanks for catching that.

docs/source/reference/datasets.rst
```diff
@@ -32,8 +32,8 @@ OnlineProductsDataset
 .. autofunction:: OnlineProductsDataset
 
 
-PASCAL VOC
-----------
+VOC
```
Hakuyume: I think `PASCAL VOC` is better for documents. How do you think of it?

yuyu2172: OK.

Hakuyume:
LGTM

Issue #187, opened 2017-05-24
Change image channels from BGR to RGB


yuyu2172:
What to change

- [ ] utils/image.py
- [ ] visualizations/vis_image.py
- [ ] imagenet mean https://github.com/pfnet/chainercv/blob/master/chainercv/links/model/ssd/ssd_vgg16.py#L25 and https://github.com/pfnet/chainercv/blob/master/chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py#L135
- [ ] change weights uploaded on  https://github.com/yuyu2172/share-weights/releases

yuyu2172:
Fixed #190

PR #190, opened 2017-05-24
Change convention from BGR to RGB

yuyu2172:
Fix points raised in #187.
Weights uploaded on the internet and the links to them will be changed in a different PR.

### DIFF
--- a/chainercv/transforms/image/pca_lighting.py
+++ b/chainercv/transforms/image/pca_lighting.py
@@ -38,9 +38,9 @@ def pca_lighting(img, sigma, eigen_value=None, eigen_vector=None):
         eigen_value = np.array((0.2175, 0.0188, 0.0045))
     if eigen_vector is None:
         eigen_vector = np.array((
-            (0.4009, -0.814,  0.4203),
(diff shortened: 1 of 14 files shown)


Hakuyume:
Could you fix `trainsfroms.pca_lighting`?

yuyu2172:
Right. Thanks for pointing that out.

chainercv/transforms/image/pca_lighting.py
```diff
@@ -38,9 +38,9 @@ def pca_lighting(img, sigma, eigen_value=None, eigen_vector=None):
         eigen_value = np.array((0.2175, 0.0188, 0.0045))
     if eigen_vector is None:
         eigen_vector = np.array((
-            (0.4009, -0.814,  0.4203),
+            (-0.5675, -0.5808, -0.5836),
```
Hakuyume: Could you add "img is CHW and RGB format" to `Args` section?

yuyu2172: Ok

Hakuyume:
LGTM

PR #203, opened 2017-05-27
Add speed of Faster RCNN to README

yuyu2172:
Code to benchmark ChainerCV

```python
from chainercv.links import FasterRCNNVGG16
import cupy
import chainer
import time
from chainercv.utils import read_image


img = read_image('data/demo/000456.jpg')

model = FasterRCNNVGG16(pretrained_model='voc07')
model.to_gpu()

_, H, W = img.shape

img_cpu = model.prepare(img)
_, o_H, o_W = img_cpu.shape
scale = float(o_H) / H


for i in range(5):
    img = chainer.Variable(img_cpu[None], volatile=chainer.flag.ON)
    img.to_gpu()
    start = time.time()
    out = model(img, scale)
    out[0].to_cpu()
    print time.time() - start


print(1./ 0.0629138946533 ) # time of Caffe
print(1. / 0.0617141723633)  # time of ChainerCV

```

### DIFF
--- a/examples/faster_rcnn/README.md
+++ b/examples/faster_rcnn/README.md
@@ -7,6 +7,20 @@
 | VOC 2007 trainval | VOC 2007 test|  69.9 mAP [1] | 70.5 mAP |
 
 
+### Speed
+
+Speed of the Caffe implementation and ChainerCV implementation are almost same in the test mode.
+We compared the time it takes to forward an image.
+
+We compared `chainercv.links.FasterRCNN.__call__` and [Caffe's equivalent function](https://github.com/rbgirshick/py-faster-rcnn/blob/master/lib/fast_rcnn/test.py#L154).
+We used [an image](https://github.com/rbgirshick/py-faster-rcnn/blob/master/data/demo/000456.jpg) from PASCAL VOC 2007 test split.
+Our experiment was conducted on Ubuntu 14.04.5 with Pascal Titan X.
+
+| Reference Implementation | ChainerCV |
+|:-:|:-:|
+|  15.9 img/s | 16.2 img/s |
+
+
 ### Demo
 
 ```
--- a/examples/faster_rcnn/README.md
+++ b/examples/faster_rcnn/README.md
@@ -9,16 +9,16 @@
 
 ### Speed
 
-Speed of the Caffe implementation and ChainerCV implementation are almost same in the test mode.
+ChainerCV and Caffe (py-faster-rcnn) implementations run at almost the same speed.
 We compared the time it takes to forward an image.
 
 We compared `chainercv.links.FasterRCNN.__call__` and [Caffe's equivalent function](https://github.com/rbgirshick/py-faster-rcnn/blob/master/lib/fast_rcnn/test.py#L154).
 We used [an image](https://github.com/rbgirshick/py-faster-rcnn/blob/master/data/demo/000456.jpg) from PASCAL VOC 2007 test split.
 Our experiment was conducted on Ubuntu 14.04.5 with Pascal Titan X.
 
-| py-faster-rcnn | ChainerCV |
+| Reference | ChainerCV |
 |:-:|:-:|
-|  15.9 img/s | 16.2 img/s |
+|  15.9 FPS | 16.2 FPS |
 
 
 ### Demo
--- a/examples/faster_rcnn/README.md
+++ b/examples/faster_rcnn/README.md
@@ -2,7 +2,7 @@
 
 ### Performance
 
-| Training Setting | Evaluation | Reference Implementation | ChainerCV |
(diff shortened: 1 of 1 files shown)


yuyu2172:
Column for performance can reference multiple sources.

examples/faster_rcnn/README.md
```diff
@@ -2,11 +2,25 @@
...
+Speed of the Caffe implementation and ChainerCV implementation are almost same in the test mode.
+We compared the time it takes to forward an image.
+
+We compared `chainercv.links.FasterRCNN.__call__` and [Caffe's equivalent function](https://github.com/rbgirshick/py-faster-rcnn/blob/master/lib/fast_rcnn/test.py#L154).
+We used [an image](https://github.com/rbgirshick/py-faster-rcnn/blob/master/data/demo/000456.jpg) from PASCAL VOC 2007 test split.
+Our experiment was conducted on Ubuntu 14.04.5 with Pascal Titan X.
+
+| py-faster-rcnn | ChainerCV |
+|:-:|:-:|
+|  15.9 img/s | 16.2 img/s |
```
Hakuyume: How about `FPS`? This unit is commonly used in the area of object detection.

examples/faster_rcnn/README.md
```diff
@@ -2,11 +2,25 @@
...
+### Speed
+
+Speed of the Caffe implementation and ChainerCV implementation are almost same in the test mode.
+We compared the time it takes to forward an image.
+
+We compared `chainercv.links.FasterRCNN.__call__` and [Caffe's equivalent function](https://github.com/rbgirshick/py-faster-rcnn/blob/master/lib/fast_rcnn/test.py#L154).
+We used [an image](https://github.com/rbgirshick/py-faster-rcnn/blob/master/data/demo/000456.jpg) from PASCAL VOC 2007 test split.
+Our experiment was conducted on Ubuntu 14.04.5 with Pascal Titan X.
+
+| py-faster-rcnn | ChainerCV |
```
Hakuyume: Both `py-faster-rcnn` and `Reference` refer the same implementation?

examples/faster_rcnn/README.md
```diff
@@ -2,11 +2,25 @@
...
 
-| Training Setting | Evaluation | Reference Implementation | ChainerCV |
+| Training Setting | Evaluation | Reference | ChainerCV |
 |:-:|:-:|:-:|:-:|
 | VOC 2007 trainval | VOC 2007 test|  69.9 mAP [1] | 70.5 mAP |
 
 
+### Speed
+
+Speed of the Caffe implementation and ChainerCV implementation are almost same in the test mode.
```
Hakuyume: `Speed ... are`

yuyu2172: Can you explain it in detail?

examples/faster_rcnn/README.md
yuyu2172: Yes, they both refer to the same implementation.
I did not want to say `py-faster-rcnn` at Performance section because it can contain results from different implementations.
For example, ResNet implementation is not included in py-faster-rcnn.

Anyway, I think this can be anything until we have different results to post.

examples/faster_rcnn/README.md
yuyu2172: Sure.

examples/faster_rcnn/README.md
Hakuyume: The subject of this sentence is `speed`, which is a singular form. And the verb of it is `are`, which is a plural form. Is it OK? I mean `Speed ... is` or `Speeds ... are` are better.

examples/faster_rcnn/README.md
Hakuyume: I see. How about using `Reference` in Speed section?

examples/faster_rcnn/README.md
yuyu2172: I am not sure about that.
Since we are talking of the `speed of A and B` (i.e. shorthand for `speed of A and speed of B`), shouldn't they be plural?
I looked it up on the internet :smiling_imp:  [related](http://www.examveda.com/the-speed-of-a-and-b-are-in-the-ratio-34-a-takes-20-minutes-more-than-b-to-reach-a-destination-time-in-which-a-reach-the-destination-386/).

By the way, it seems that it should be `the speed`.

yuyu2172: I looked up more on the internet, and it seems that `The speed ... is` is ok,
 too (https://english.stackexchange.com/questions/134703/using-a-singular-or-plural-verb-after-and-or).

I will just say that `ChainerCV and Caffe implementations run at almost the same speed.`

Hakuyume: I see, thank you!

Hakuyume:
LGTM

PR #208, opened 2017-05-29
Update installation guide in the documentation

yuyu2172:
I considered adding a docker file, but I didn't because it was tricky to setup GPUs and matplotlib.
Instead, I added more detailed steps to install ChainerCV using anaconda.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -16,6 +16,8 @@ You can install ChainerCV using `pip`.
 
 .. code-block:: shell
 
+    # If Cython has not been installed yet, install it by commands like
+    # pip install Cython
     pip install chainercv
 
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -7,15 +7,48 @@ ChainerCV is a **deep learning based computer vision library** built on top of `
 
 
 Install Guide
--------------
+=============
 
-Please install `Chainer <https://github.com/pfnet/chainer/>`_ first.
+Pip
+~~~
+
 You can install ChainerCV using `pip`.
 
 .. code-block:: shell
 
     pip install chainercv
 
+
+Anaconda
+~~~~~~~~
+
+You can setup ChainerCV including its dependencies using anaconda.
+
+.. code-block:: shell
+
+    # For python 3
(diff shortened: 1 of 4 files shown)


docs/source/index.rst
```diff
@@ -7,16 +7,49 @@ ChainerCV is a **deep learning based computer vision library** built on top of `
...
+
+Pip
+~~~
 
-Please install `Chainer <https://github.com/pfnet/chainer/>`_ first.
 You can install ChainerCV using `pip`.
 
 .. code-block:: shell
 
     pip install chainercv
```
Hakuyume: Users cannot install ChainerCV by this instruction.
```
$ pip install chainer
$ pip install chainercv
Collecting chainercv
  Using cached chainercv-0.4.7.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-i4rdgwhe/chainercv/setup.py", line 6, in <module>
        from Cython.Distutils import build_ext
    ModuleNotFoundError: No module named 'Cython'
    
    ---------------------------------------
```

environment.yml
```diff
@@ -9,6 +9,5 @@ dependencies:
   - opencv3=3.2.0
   - matplotlib
   - numpy
-  - scikit-learn
   - pip:
-    - chainer
+    - chainer=1.24
```
Hakuyume: How about adding `pillow`?

docs/source/index.rst
```diff
@@ -7,16 +7,51 @@ ChainerCV is a **deep learning based computer vision library** built on top of `
...
+
+Pip
+~~~
 
-Please install `Chainer <https://github.com/pfnet/chainer/>`_ first.
 You can install ChainerCV using `pip`.
 
 .. code-block:: shell
 
+    # If Cython has not been installed yet, install it by commands like
```
Hakuyume: `pip install Cython` is a single command. How about `install it by the following command`?

yuyu2172: I wanted it to imply that `pip install Cython` is not the only way to install Cython.
People install them using apt as an alternative to pip for instance.

Hakuyume: I see.

Hakuyume:
LGTM

PR #213, opened 2017-05-30
Add DetectionVOCEvaluator to train script of Faster R-CNN

yuyu2172:


examples/faster_rcnn/train.py
```diff
@@ -104,6 +110,12 @@ def transform(in_data):
             trigger=plot_interval
         )
 
+    trainer.extend(
+        DetectionVOCEvaluator(
+            test_iter, model.faster_rcnn, use_07_metric=True),
+        trigger=(args.step_size, 'iteration'),
+        invoke_before_training=False)
+
```
Hakuyume: I guess this evaluator runs only once. Is it OK?

yuyu2172: Yes

Hakuyume: How about running evaluator twice (when lr changes and when the training finishes)?

yuyu2172: OK.

Hakuyume:
LGTM

PR #214, opened 2017-05-30
Fix readthedoc to compile properly by installing Cython

yuyu2172:


Hakuyume:
In the title of the PR, `copile` -> `compile`?

yuyu2172:
Thanks

docs/requirements.txt
```diff
@@ -1,4 +1,5 @@
-chainer
-scikit-image
-scikit-learn
+chainer=1.24
+Cython
 matplotlib
+numpy
+pillow
```
Hakuyume: Both `Cython` and `Pillow` start with an upper character in pypi. Why are you using upper character only for `Cython`?

yuyu2172: Sorry, I didn't know that Pillow start with a capital letter.

Hakuyume: https://pypi.python.org/pypi/Pillow/4.1.1

Hakuyume:
LGTM

PR #216, opened 2017-05-30
Fix case of package names

Hakuyume:
`Cython` and `Pillow` start with  a capital letter in PyPi.
https://pypi.python.org/pypi/Cython/0.25.2
https://pypi.python.org/pypi/Pillow/4.1.1


yuyu2172:
Thanks!

yuyu2172:
LGTM

PR #224, opened 2017-05-30
Add padding

Hakuyume:


yuyu2172:
Is `resize_contain` inadequate?

Hakuyume:
I'm sorry. As you pointed out, `resize_contain` is enough.

PR #226, opened 2017-05-30
Add eval_hook to DetectionVOCEvaluator

yuyu2172:


Hakuyume:
Evaluators are hooked by the trainer. I think registering a hook to the trainer is enough. Why do you want to add hook to evaluator?

yuyu2172:
Initially, I wanted to configure preset of a model inside an extension.
I found that I can't do that with this modification, so I have no strong reason to add this feature.

PR #227, opened 2017-05-30
Improve split_iterator

Hakuyume:
Currently, `split_iterator` keeps all elements even if the corresponding iterator is deleted.
```
import resource
from chainercv.utils import split_iterator

# set limit of memory usage
resource.prlimit(0, resource.RLIMIT_AS, (1024, 1024))


def infinite_iterator():
    i = 0
    while True:
        yield i, i * i
        i += 1


# use the first element only
iter_0, iter_1 = split_iterator(infinite_iterator())
del iter_1

for _ in range(65536):
    next(iter_0)  # MemoryError
```
With this PR, this code works well.


yuyu2172:
Is it possible to add a test?

Also, I think it is better to write a simple docstring for BufferedIterator. (meaning of args at least)

Can you add a comment on line 20? (e.g. # Values for deleted buffers are ignored)

yuyu2172:
What is the difference with this function and `itertools.tee`?

yuyu2172:
I have found a stack overflow question related to this function.
Do you think that our implementation is better than the solution proposed there?

https://stackoverflow.com/questions/30805000/how-to-unzip-an-iterator

By the way, I like the new name.

yuyu2172:
How about `iunzip` to indicate that this function unzips an iterator

Hakuyume:
Thank you for your suggesion. `itertools.tee` solves all problems.

Hakuyume:
I will remove `unzip`.

yuyu2172:
Does it have memory issue that you solved in your implementation?

Hakuyume:
> Does it have memory issue that you solved in your implementation?

`itertools.tee` solves this memory issue. It supports deletion of iterators.
The only problem is that `itertools.tee` requires the number of iterators to be returned as its argument.

yuyu2172:
> The only problem is that itertools.tee requires the number of iterators to be returned as its argument.

Yes. I am concerned on that too.
Some solutions I cam up with: 

+ pass size of tuple to `apply_prediction_to_iterator`
+ copy an iterator and check the size of tuple.

yuyu2172:
Or, make _apply return the number of values.

Hakuyume:
I think it is better to modify `unzip` and `apply_prediction_to_iterator` at the same time. I will close the PR and work on #229

PR #217, opened 2017-05-30
Change eval_semantic_segmentation

yuyu2172:

PR #232, opened 2017-06-01
Improve eval_detection_voc

Hakuyume:
Perhaps, the return values will change according to #217.


Hakuyume:
I tested the computing time using results of SSD300 on VOC2007 test.
Current implementation took 37 secs and my implementation took 16 secs.
The computed scores were same.

yuyu2172:
Is the behavior consistent with the original VOC code?
I always check the consistency by running the code below.

https://github.com/yuyu2172/pascal_evaluation

Hakuyume:
It got more faster. Now it takes 1.6 secs.

yuyu2172:
I confirmed that it runs identical

tests/evaluations_tests/test_eval_detection_voc.py
```diff
@@ -97,3 +98,57 @@ def test_eval_detection_voc_difficult(self):
...
+
+    @classmethod
+    def setUpClass(cls):
+        base_url = 'https://github.com/yuyu2172/' \
+                   'share-weights/releases/download/0.0.2'
+
+        cls.dataset = np.load(request.urlretrieve(os.path.join(
+            base_url, 'example_detection_dataset_2017_06_01.npz'))[0])
+        cls.result = np.load(request.urlretrieve(os.path.join(
+            base_url, 'example_detection_result_2017_06_01.npz'))[0])
```
yuyu2172: eval_detection_voc_result_*
How about?
We may add similar kind of file for COCO.

Hakuyume: Nice idea. I will rename them.

chainercv/evaluations/eval_detection_voc_ap.py
```diff
@@ -0,0 +1,297 @@
...
+        prec[l] = tp / np.maximum(fp + tp, np.finfo(np.float64).eps)
+        rec[l] = tp / n_pos[l]
+
+    return prec, rec
+
+
+def calc_detection_voc_ap(prec, rec, use_07_metric=False):
+    """Calculate average precision based on evaluation code of PASCAL VOC.
+
+    This function calculates average precision from given precision and recall.
```
yuyu2172: `from a given precision and recall`

chainercv/evaluations/eval_detection_voc_ap.py
```diff
@@ -0,0 +1,297 @@
...
+    The code is based on the evaluation code used in PASCAL VOC Challenge.
+
```
yuyu2172: Can you add what you mean by `a class is invalid`?
It sounded as if the class is categorized as `ignore`.

chainercv/evaluations/eval_detection_voc_ap.py
```diff
@@ -0,0 +1,297 @@
...
+        rec (list of numpy.array): A list of arrays.
+            :obj:`rec[l]` indicates recall for class :math:`l`.
+            If class :math:`l` is invalid, :obj:`rec[l]` is
+            set to :obj:`None`.
+        use_07_metric (bool): Whether to use PASCAL VOC 2007 evaluation metric
+            for calculating average precision. The default value is
+            :obj:`False`.
+
+    Returns:
+        list:
```
yuyu2172: How about returning an array and setting invalid class ap as np.nan?
One advantage of this approach is that taking mean can simply be done using np.nanmean.

chainercv/evaluations/eval_detection_voc_ap.py
```diff
@@ -0,0 +1,297 @@
...
+        if next(iter_, None) is not None:
+            raise ValueError('Length of input iterables need to be same.')
+
+    n_fg_class = max(n_pos.keys()) + 1
+    prec = [None] * n_fg_class
+    rec = [None] * n_fg_class
+
+    for l in n_pos.keys():
+        score_l = np.array(score[l])
+        match_l = np.array(match[l], dtype=np.int8)
```
Hakuyume: This variable does not exist in the original code because I introduced this. There is no problem about using `np.int8` because `match_l` keeps only `0`, `1` and `-1`.

yuyu2172: Why not `np.array`?
I don't think it has to be np.array(..., dtype=np.int8)

yuyu2172: You needed to make it an integer array.
I got it.
Lgtm

chainercv/evaluations/eval_detection_voc_ap.py
Hakuyume: Nice idea.

Hakuyume:
@yuyu2172 I reflected your comments.

chainercv/evaluations/eval_detection_voc_ap.py
yuyu2172: You can change the return type from list to array.

chainercv/evaluations/eval_detection_voc_ap.py
```diff
@@ -0,0 +1,298 @@
...
+
+        order = score_l.argsort()[::-1]
+        match_l = match_l[order]
+
+        tp = np.cumsum(match_l == 1)
+        fp = np.cumsum(match_l == 0)
+
+        prec[l] = tp / np.maximum(fp + tp, np.finfo(np.float64).eps)
+        rec[l] = tp / n_pos[l]
+
```
yuyu2172: prec and rec can be array too with nan for invalid classes.

Hakuyume: I think we can't do that. Currently, `prec` and `rec` are lists of arrays. The length of each array is same as the number of predicted bounding boxes of corresponding class. So the lengths are different from each other.

yuyu2172: Right.
That was my mistake.

chainercv/evaluations/eval_detection_voc_ap.py
```diff
@@ -0,0 +1,294 @@
...
+    Returns:
+        tuple of two lists:
+        This function returns two lists: :obj:`prec` and :obj:`rec`.
+
+        * :obj:`prec`: A list of arrays. :obj:`prec[l]` indicates precision \
+            for class :math:`l`. If class :math:`l` does not exist in \
+            either :obj:`pred_labels` or :obj:`gt_labels`, :obj:`prec[l]` is \
+            set to :obj:`None`.
+        * :obj:`rec`: A list of arrays. :obj:`rec[l]` indicates recall \
+            for class :math:`l`. If class :math:`l` does not exist in \
```
yuyu2172: Inindicates--> is

yuyu2172:
LGTM

yuyu2172:
By the way, it has been confirmed that detection scores have not changed in this branch.

Hakuyume:
> By the way, it has been confirmed that detection scores have not changed in this branch.

Thank you for confirming. I added a test code that compares results of ChainerCV with that of the original MATLAB code. This will make our codes more reliable.

PR #191, opened 2017-05-24
Convert convention from WH to HW

yuyu2172:
Fix subset of #189 

Changes include following list of directories.

+ transforms/image
+ transforms/bbox
+ transforms/keypoint
+ links/models/faster_rcnn
+ links/model/ssd
+ links/model/segnet
+ utils/testing/generate_random_bbox

PR #234, opened 2017-06-01
Convert convention from WH to HW

yuyu2172:
Reopening #191 and send it to branch reverse-coords-order


chainercv/transforms/image/resize_contain.py
```diff
@@ -17,7 +17,7 @@ def resize_contain(img, size, fill=0, return_param=False):
         img (~numpy.ndarray): An array to be transformed. This is in
```
Hakuyume: Please fix L9-L10.
https://github.com/yuyu2172/chainercv/blob/wh-hw/chainercv/transforms/image/resize_contain.py#L9-L10

Hakuyume:
LGTM

PR #239, opened 2017-06-01
Fix eval_voc07

Hakuyume:


yuyu2172:
#237

Hakuyume:
@yuyu2172 Travis CI failed only for Python 3 without optional modules. Do you know the reason?

yuyu2172:
Let's see if it fails again after restarting.

yuyu2172:
It passed.
LGTM

PR #206, opened 2017-05-28
Add training code of SSD

Hakuyume:
merge after #291

PR #245, opened 2017-06-05
[WIP] Improve SSD

Hakuyume:


chainercv/links/model/ssd/multibox_encoder.py
```diff
@@ -0,0 +1,218 @@
...
+
+import itertools
+import numpy as np
+
+import chainer
+
+from chainercv import utils
+
+
+class MultiboxEncoder(object):
```
yuyu2172: I wouldn't name an object with method `decode` as `*Encoder`.

Hakuyume:
This PR contains a part of training.  I think it is better to develop everything in #206.

PR #254, opened 2017-06-07
Add coarse level class names for OnlineProductsDataset

yuyu2172:


Hakuyume:
Is it possible to provide `fine_label_names` (simply `label_names`)?

yuyu2172:
It has to be generated by a program.
For instance, the first class belonging to `bicycle` super-class would be `bicycle_0000`.

Hakuyume:
@yuyu2172 Could you solve the conflict?

yuyu2172:
OK

PR #255, opened 2017-06-08
Improve segnet eval

Hakuyume:


yuyu2172:
On master branch, the reported pixel accuracy was wrong.
Thanks for fixing that.

PR #258, opened 2017-06-09
[WIP] add omniglot dataset

yuyu2172:


chainercv/datasets/omniglot/omniglot_label_names.py
```diff
@@ -0,0 +1,1640 @@
+# This is a script to generate this label names
```
Hakuyume: these

yuyu2172:
Perhaps, this should be added to Chainer's main repo.

PR #265, opened 2017-06-12
Add VGG16

yuyu2172:
Merge after #271.

This PR adds a VGG16Layers, which will have APIs consistent with rest of the links in ChainerCV.

- [x] RGB images will work for pretraiend model. (TODO: A script to convert a caffe weight to npz file will be added).
- [x] Separate caffe loader from the model code. This is good for making the model code shorter.
- [x] predict returns an iterable of `numpy.array`.
- [x] Have an interface to select weight initializers at `__init__` (users can manually select to stop using random initializer)
- [x] `__call__` does not have `layers` option. Also, the return value is not a dictionary but a `chainer.Variable`.
- [x] `__init__` takes `feature` option, which selects the feature that is going to be returned by `__call__`.
- [x] depending on `feature` option, layers that are unnecessary will automatically be deleted.

These changes are made for the following concrete scenarios in mind.
+ The chain will be used as a feature extractor for networks used in tasks other than classification. For example, this chain can be used as a feature extractor for FasterRCNNVGG16 with a proper initialization.
+ Usage together with other ChainerCV functions, which assume RGB image order.
+ I am assuming that the chain is used to extract only one type of feature. Also, I am assuming that the kind of extracted feature is fixed after initializing the chain. (EDIT: This assumption is no longer true in our final design)


chainercv/links/model/classification/vgg.py
```diff
@@ -0,0 +1,122 @@
...
+            self.conv4_2 = L.Convolution2D(512, 512, 3, 1, 1, **kwargs)
+            self.conv4_3 = L.Convolution2D(512, 512, 3, 1, 1, **kwargs)
+            self.conv5_1 = L.Convolution2D(512, 512, 3, 1, 1, **kwargs)
+            self.conv5_2 = L.Convolution2D(512, 512, 3, 1, 1, **kwargs)
+            self.conv5_3 = L.Convolution2D(512, 512, 3, 1, 1, **kwargs)
+            self.fc6 = L.Linear(512 * 7 * 7, 4096, **kwargs)
+            self.fc7 = L.Linear(4096, 4096, **kwargs)
+            self.fc8 = L.Linear(4096, 1000, **kwargs)
+
+        self.functions = collections.OrderedDict([
```
t-abe: I do not recommend this style of network definition. It might cause problems when an instance of this class is copied.
(See https://github.com/chainer/chainer/pull/2810 .)

yuyu2172: I didn't know that! Thank you!

yuyu2172:
I am trying to reproduce the evaluation score reported [here](http://www.vlfeat.org/matconvnet/pretrained/).
The website reports that VGG16 scores error 28.5 % for Top-1 Error with single center crop.

Currently, my implementation scored 32%, and I am going to investigate the reason why it scored below the reported score.

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,299 @@
...
+
+            if name in features:
+                features.remove(name)
+            if len(features) == 0:
+                pop_funcs = True
+
+        return funcs
+
+    def __call__(self, x):
+        """Fowrard VGG16.
```
Hakuyume: `Fowrard` -> `Forward`

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,299 @@
...
+            'n_class': 1000,
+            'url': 'https://github.com/yuyu2172/share-weights/releases/'
+            'download/0.0.3/vgg16_imagenet_convert_2017_06_15.npz'
+        }
+    }
+
+    def __init__(self, pretrained_model=None, n_class=None,
+                 features='prob', initialW=None, initial_bias=None,
+                 mean=_imagenet_mean, do_ten_crop=False):
+        if isinstance(features, (list, tuple)):
```
Hakuyume: How about checking whether `feature` is `str` or not? This is better because `iterable` is not only `list` and `tuple`.

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,299 @@
...
+        pretrained_model (str): The destination of the pre-trained
+            chainer model serialized as a :obj:`.npz` file.
+            If this is one of the strings described
+            above, it automatically loads weights stored under a directory
+            :obj:`$CHAINER_DATASET_ROOT/pfnet/chainercv/models/`,
+            where :obj:`$CHAINER_DATASET_ROOT` is set as
+            :obj:`$HOME/.chainer/dataset` unless you specify another value
+            by modifying the environment variable.
+        n_class (int): The dimension of the output of fc8.
+        feature (str or iterable of strings): The name of the feature to output
```
Hakuyume: `features`?

chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py
```diff
@@ -103,7 +102,7 @@ def __init__(self,
         if vgg_initialW is None and pretrained_model:
             vgg_initialW = chainer.initializers.constant.Zero()
 
-        extractor = VGG16FeatureExtractor(initialW=vgg_initialW)
+        extractor = VGG16Layers(feature='conv5_3', initialW=vgg_initialW)
```
Hakuyume: `features=`

yuyu2172:
I think we can change the name of the link to `VGG16` instead of `VGG16Layers`.

yuyu2172:
On specification of Feature Extractor models.
This is assumed by `FeatureExtractionPredictor` that wraps around a Feature Extractor model to add `predict` method.

+ It has a method `__call__` that takes batch of images and returns batch of features. One or multiple features can be extracted simultaneously.
+ It has an attribute `mean` that determines mean image that has shape `(C, 1, 1)`.

On Implementation of Feature Extractor models.
+ It takes `layer_names` that determines the kind of features to return at `__init__`.
+ From the `layer_names`, it deletes a link that can be unnecessary.
+ I used a base class `SequentialExtractor` that can be used for any feature extraction links that always have linear flow of computation by block level. For instance, VGG, ResNet and ResNext can use this class. A link that has multiple branches can not use this base class (e.g. GoogleNet has multiple losses branching from the network).

Hakuyume:
@yuyu2172 Could you make another PR for `assert_is_classification_link`?

Hakuyume:
I think `assert_is_classification_link` is OK. If we want to check VGG's other behaviours, it is better to add another assertion utility.

yuyu2172:
> I think assert_is_classification_link is OK. If we want to check VGG's other behaviours, it is better to add another assertion utility.

Is this the part that I should make utility for?
https://github.com/yuyu2172/chainercv/blob/a5f0be746ca1d43bb102a07bbf7a9790cc6823e3/tests/links_tests/model_tests/test_feature_extraction_predictor.py#L37

yuyu2172:
I updated Faster R-CNN VGG16 to use `remove_unused`.

yuyu2172:
#265 and #271 are ready to be merged.

Hakuyume:
@yuyu2172 OK, I will check them.

chainercv/links/connection/convolution_2d_block.py
```diff
@@ -0,0 +1,50 @@
...
+    batch normalization or not. The default value is :obj:`False`.
+
+    """
+
+    def __init__(self, in_channels, out_channels, ksize=None, stride=1, pad=0,
+                 activation=relu, use_bn=False,
+                 nobias=False, initialW=None, initial_bias=None,
+                 decay=0.9, eps=2e-5, dtype=np.float32,
+                 use_gamma=True, use_beta=True,
+                 initial_gamma=None, initial_beta=None):
```
Hakuyume: I don't like flatten all parameters in arguments list. How about taking two dicts?
```
def __init__(self, in_channels, out_channels, activation=relu, use_bn=False, conv_kwargs=dict(), bn_kwargs=dict()):
```

Hakuyume: It is better to support `Convolution2D(512)` style if possible.

chainercv/links/model/feature_extraction_predictor.py
```diff
@@ -0,0 +1,137 @@
...
+class FeatureExtractionPredictor(chainer.Chain):
+
+    """Wrapper class that adds predict method to a feature extraction model.
+
+    The :meth:`predict` takes three steps to make predictions.
+
+    1. Preprocess images
+    2. Forward the preprocessed images to the network
+    3. Average features in the case when ten-crop is used.
+
```
Hakuyume: How about adding `Args` section?

chainercv/links/model/feature_extraction_predictor.py
```diff
@@ -0,0 +1,137 @@
...
+        >>> prob = model.predict([img])
+        # Predicting multiple features
+        >>> model.extractor.feature_names = ['conv5_3', 'fc7']
+        >>> conv5_3, fc7 = model.predict([img])
+
+    """
+
+    def __init__(self, extractor,
+                 size=(224, 224), scale_size=256,
+                 do_ten_crop=False):
```
Hakuyume: There are other cropping techniques, such as 5-crop and 144-crops. How about taking a string `crop` instead of a bool `do_ten_crop`? If `crop == 'center'`, `center_crop` is used (default behaviour). if `crop == '10'`, ten crop is used. We can add `'5'` and `'144'` when they become needed.

chainercv/links/model/feature_extraction_predictor.py
```diff
@@ -0,0 +1,137 @@
...
+        >>> model = FeatureExtractionPredictor(base_model)
+        >>> prob = model.predict([img])
+        # Predicting multiple features
+        >>> model.extractor.feature_names = ['conv5_3', 'fc7']
+        >>> conv5_3, fc7 = model.predict([img])
+
+    """
+
+    def __init__(self, extractor,
+                 size=(224, 224), scale_size=256,
```
Hakuyume: It is strange that `size` is a tuple of ints and `scale_size` is an int.

Hakuyume: `size` -> `crop_size`

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,157 @@
...
+            If this is one of the strings described
+            above, it automatically loads weights stored under a directory
+            :obj:`$CHAINER_DATASET_ROOT/pfnet/chainercv/models/`,
+            where :obj:`$CHAINER_DATASET_ROOT` is set as
+            :obj:`$HOME/.chainer/dataset` unless you specify another value
+            by modifying the environment variable.
+        n_class (int): The dimension of the output of fc8.
+        mean (numpy.ndarray): A mean image. If :obj:`None` and
+            a supported pretrained model is used,
+            the mean image used to train the pretrained model will be used.
```
Hakuyume: If `pretrained_model == 'imagenet'`, `mean` is not a mean image but a mean value.

chainercv/links/model/feature_extraction_predictor.py
Hakuyume: Another solution is taking a callable that crops image(s). For example, `crop_func=center_crop`.

chainercv/links/connection/convolution_2d_block.py
```diff
@@ -0,0 +1,50 @@
+import numpy as np
+
+import chainer
+from chainer.functions import relu
+from chainer.links import BatchNormalization
+from chainer.links import Convolution2D
+
+
+class Convolution2DBlock(chainer.Chain):
+    """Convolution2D --> (Batch Normalization) --> Activation
```
Hakuyume: It is better to give more specific name because some networks use `BN->ReLU->Conv` block.

yuyu2172: Do you have any suggestion?

chainercv/links/connection/convolution_2d_block.py
yuyu2172: `ksize` needs to be specified in `Convolution2D`.

How about
`self, in_channels, out_channels,  ksize=None, activation=relu, use_bn=False, conv_kwargs=dict(), bn_kwags=dict()`

chainercv/links/connection/convolution_2d_block.py
yuyu2172: I think `Convolution2DBlock` is not bad because `BN->ReLU->Conv` is much less common.

chainercv/links/model/feature_extraction_predictor.py
yuyu2172: I found it convenient to support following two cases.

+ crop_size is not a square
+ resize to certain shape instead of scaling

Therefore, I changed `prepare` to support both an integer and a tuple. 4bbdd5a

chainercv/links/connection/convolution_2d_block.py
Hakuyume: DenseNet and some newer researches use `BN->ReLU->Conv` as a basic block. I guess it will be more common. We should not assume `Conv->BN->ReLU` is common in the future.

Hakuyume: How about `ConvBNBlock` and `BNConvBlock`?

yuyu2172: > DenseNet and some newer researches use BN->ReLU->Conv as a basic block. I guess it will be more common. We should not assume Conv->BN->ReLU is common in the future.

OK.

If `BN` is in the name, it would be awkward to use it without batch normalization.

yuyu2172: One way to avoid this issue is to separate Conv->ReLU and Conv->BN->ReLU.

yuyu2172: In other words, make `Convolution2DBlock` and `Convolution2DBNBlock`

Hakuyume: > One way to avoid this issue is to separate Conv->ReLU and Conv->BN->ReLU.

It looks good to me. So `BN->ReLU->Conv` will be `BNConvolution2DBlock`?

I think `Convolution2D` is too lengthy and  2D is very common in computer vision. How about `ConvBlock` and `ConvBNBlock`?

yuyu2172: I agree that it is lengthy. 
However, I think that it is too dangerous to assume that all inputs are 2D.

Also, I realized that we are making an assumption that the activation function always comes last.
How about `Conv2DActivation` and `Conv2DBNActivation`?
So, `Bn->ReLU->Conv` would be `BNActivationConv2D`.

chainercv/links/connection/convolution_2d_block.py
Hakuyume: > ksize needs to be specified in Convolution2D

I see. It is OK. Another solution is raising an exception if `ksize` doesn't exist in`conv_kwargs`.

chainercv/links/connection/convolution_2d_block.py
Hakuyume: > Also, I realized that we are making an assumption that the activation function always comes last.

That is a good point. Can you shorten `Activation`?  (`Activ`? I have no idea).

yuyu2172: `Activ` sounds good.
So, `Conv2DActiv` and `Conv2DBNActiv`?

Hakuyume: How about inheriting `SequentialFeatureExtractor` to implement this block? We can remove `__call__` function.

chainercv/links/connection/convolution_2d_block.py
yuyu2172: For `Conv2DBNActiv`, we can flatten all parameters for Convolution2D, but pass kwargs for BatchNorm by dictionary.
One advantage of this approach is that the interface will be completely same with `Convolution2D` if nothing additional is passed for BatchNormalization.

chainercv/links/connection/convolution_2d_block.py
Hakuyume: > So, Conv2DActiv and Conv2DBNActiv?

OK, it sounds good.

yuyu2172: Personally, I find the verbose `chainer.Chain` to be easier to read than `SequentialX`.
(I strongly believe that `SequentialFeatureExtractor` is good for implementing a feature extractor like VGG.)

Besides, I do not think that blocks need any of the features `SequentialFeatureExtractor` provides.

chainercv/links/connection/convolution_2d_block.py
Hakuyume: > One advantage of this approach is that the interface will be completely same with Convolution2D if nothing additional is passed for BatchNormalization.

I see. So, the additional parameters will be `activation` and `bn_kwargs`? (For `Conv2DActiv, it will be only `activation`)? That sounds good.

yuyu2172: Yes.
I will modify it with this design.

chainercv/links/connection/convolution_2d_block.py
Hakuyume: `SequentialFeatureExtractor` was only a suggestion. `chainer.Chain` is OK.

yuyu2172:
Thanks for the discussion.
I will start reflecting the discussion.

yuyu2172:
Please check if this is OK https://github.com/chainer/chainercv/pull/265/commits/6d4cf2946717be40df9b8e6aa309b7a24d34a03d.

chainercv/links/connection/conv_2d_activ.py
```diff
@@ -9,6 +9,10 @@ class Conv2DActiv(chainer.Chain):
     This is a chain that does two-dimensional convolution
     and applies an activation.
 
+    The arguments are the same as that of
+    :class:`chainer.links.Convolution2D`
+    except for `obj:`activ`.
```
Hakuyume: `obj: -> :obj:

chainercv/links/connection/conv_2d_activ.py
```diff
@@ -0,0 +1,73 @@
...
+from chainer.links import Convolution2D
+
+
+class Conv2DActiv(chainer.Chain):
+    """Convolution2D --> Activation
+
+    This is a chain that does two-dimensional convolution
+    and applies an activation.
+
+    The arguments are the same as that of
```
Hakuyume: `that of` -> `those of`

Hakuyume:
> Please check if this is OK 6d4cf29.

I think it is OK.

tests/links_tests/connection_tests/test_conv_2d_activ.py
```diff
@@ -0,0 +1,57 @@
...
+            self.l = Conv2DActiv(
+                self.in_channels, self.out_channels, self.ksize,
+                self.stride, self.pad)
+        elif self.args_style == 'None':
+            self.l = Conv2DActiv(
+                None, self.out_channels, self.ksize, self.stride, self.pad)
+        elif self.args_style == 'omit':
+            self.l = Conv2DActiv(
+                self.out_channels, self.ksize, stride=self.stride, pad=self.pad)
+
```
Hakuyume: It is better to add test of forward and test of `activ`.

yuyu2172: OK. I updated.

chainercv/links/connection/conv_2d_activ.py
yuyu2172: fixed

Hakuyume:
@yuyu2172 I merged #271. Could you merge `master` to this PR?

yuyu2172:
I merged `master` to this PR.

yuyu2172:
Do you prefer to separate this PR? (`Conv2DActiv`, `FeatureExtractionPredictor` and this)

Hakuyume:
Yes, especially `FeatureExtractionPredictor`. It makes reviewing much easier.

yuyu2172:
Merged master

chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py
```diff
@@ -74,7 +73,8 @@ class FasterRCNNVGG16(FasterRCNN):
         'voc07': {
             'n_fg_class': 20,
             'url': 'https://github.com/yuyu2172/share-weights/releases/'
-            'download/0.0.3/faster_rcnn_vgg16_voc07_2017_06_06.npz'
+            'download/0.0.4/'
+            'faster_rcnn_vgg16_voc07_trained_2017_08_06_trial_4.npz'
```
Hakuyume: Why do you use `trial_4`? Is this the best model?

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+# RGB order
+_imagenet_mean = np.array(
+    [123.68, 116.779, 103.939], dtype=np.float32)[:, np.newaxis, np.newaxis]
+
+
+class VGG16(SequentialFeatureExtractor):
+
+    """VGG16 Network for classification and feature extraction.
```
Hakuyume: `VGG16` -> `VGG-16`

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,156 @@
...
+
+    This is a feature extraction model.
+    The network can choose to output features from set of all
```
Hakuyume: `to output features` -> `features to output` or ` output features`?

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,156 @@
...
+            if pretrained_model in self._models:
+                n_class = self._models[pretrained_model]['n_class']
+            else:
+                n_class = 1000
+
+        if mean is None:
+            if pretrained_model in self._models:
+                mean = self._models[pretrained_model]['mean']
+            else:
+                mean = _imagenet_mean
```
Hakuyume: How about adding a note about these behaviours? The default values of `n_class` and `mean`.

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,156 @@
...
+    that can be collected.
+
+    Examples:
+
+        >>> model = VGG16()
+        # By default, VGG16.__call__ returns a probability score.
+        >>> prob = model(imgs)
+
+        >>> model.feature_names = 'conv5_3'
+        # This is feature conv5_3.
```
Hakuyume: How about `# This is feature conv5_3 (after ReLU).`?

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,156 @@
...
+    intermediate features.
+    The value of :obj:`VGG16.feature_names` selects the features that are going
+    to be collected by :meth:`__call__`.
+    :obj:`self.all_feature_names` is the list of the names of features
+    that can be collected.
+
+    Examples:
+
+        >>> model = VGG16()
+        # By default, VGG16.__call__ returns a probability score.
```
Hakuyume: How about `a probability score (after Softmax)`?

chainercv/links/model/vgg/vgg16.py
```diff
@@ -0,0 +1,156 @@
...
+        >>> prob = model(imgs)
+
+        >>> model.feature_names = 'conv5_3'
+        # This is feature conv5_3.
+        >>> feat5_3 = model(imgs)
+
+        >>> model.feature_names = ['conv5_3', 'fc6']
+        >>> # These are features conv5_3 and fc6.
+        >>> feat5_3, feat6 = model(imgs)
```
Hakuyume: `# These are features conv5_3 (after ReLU) and fc6 (before ReLU).`

docs/source/reference/links.rst
```diff
@@ -1,10 +1,28 @@
...
 
-.. module:: chainercv.links.model.faster_rcnn
+
+Model
+-----
+
+
+Feature Extraction
+~~~~~~~~~~~~~~~~~~
+Feature extraction models can be used to extract feature(s) given images.
```
Hakuyume: `extract feature(s) from given images.` ?

examples/classification/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+This instructions are based on the instruction found [here](https://github.com/facebook/fb.resnet.torch/blob/master/INSTALL.md#download-the-imagenet-dataset).
+
+The ImageNet Large Scale Visual Recognition Challenge (ILSVRC) dataset has 1000 categories and 1.2 million images. The images do not need to be preprocessed or packaged in any database, but the validation images need to be moved into appropriate subfolders.
+
+1. Download the images from http://image-net.org/download-images
+
+2. Extract the training data:
+  ```bash
+  mkdir train && mv ILSVRC2012_img_train.tar train/ && cd train
```
Hakuyume: `$ mkdir ...`

examples/classification/convert_from_original/convert_vgg.py
```diff
@@ -0,0 +1,37 @@
+import chainer
```
Hakuyume: Considering consistency with other examples, this file should be put under `examples/vgg/`.
And it is better to convert from `.caffemodel`.

chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py
yuyu2172: It is the model that is converted from `faster_rcnn_vgg16_voc07_2017_06_06.npz`.
It performs the same with the previously distributed model.

examples/classification/convert_from_original/convert_vgg.py
yuyu2172: 1. I think that it is OK to leave conversion code under the current directory. This is because unlike other examples, VGG does not have its own `demo.py` or `train.py`.
2. OK

Hakuyume: Putting a model-specific utility under task-specific directory seems against your intention. The conversion code for VGG-16 can be used for other task than classification.

> This is because unlike other examples, VGG does not have its own demo.py or train.py

You won't add any training codes for VGG or other classification networks?

Anyway, we should keep consistency with other examples. If you want to reduce the number of subdirectories, how about putting detection models under `detection`?

yuyu2172: > Putting a model-specific utility under task-specific directory seems against your intention. The conversion code for VGG-16 can be used for other task than classification.

OK. Your suggestion seems easier for users who are not interested in classification to access the conversion code.

> If you want to reduce the number of subdirectories, how about putting detection models under detection?

The current layout of directories are easier to explore than the nested directories.
I think it is still manageable.
If things start to get out of control, we can write a README under `examples/`.

examples/vgg/caffee2npz_vgg_16.py
```diff
@@ -0,0 +1,47 @@
+import argparse
```
Hakuyume: You can make this code more general. Please check SSD's conversion code.

examples/vgg/caffee2npz_vgg_16.py
```diff
@@ -0,0 +1,47 @@
...
+    args = parser.parse_args()
+
+    caffemodel = caffe.CaffeFunction(args.caffemodel)
+    model = VGG16(pretrained_model=None, n_class=1000)
+
+    model.conv1_1.conv.copyparams(caffemodel.conv1_1)
+    # The pretrained weights are trained to accept BGR images.
+    # Convert weights so that they accept RGB images.
+    model.conv1_1.conv.W.data[:] = model.conv1_1.conv.W.data[:, ::-1]
+    model.conv1_2.conv.copyparams(caffemodel.conv1_2)
```
Hakuyume: These `copyparams` can be reduced.

yuyu2172: What do you mean more general?

examples/vgg/caffee2npz_vgg_16.py
yuyu2172: What do you mean more general?

Hakuyume: It can take other VGG model (e.g. VGG-19).

yuyu2172: Is there a caffemodel for VGG-19?

examples/vgg/caffee2npz_vgg_16.py
Hakuyume: We can save `CaffeFunction` directly. Why do you copy the paramerters from `CaffeFunction` to `VGG16`?

yuyu2172: Because `VGG16` uses `Conv2DActiv`, which is not used in the `caffemodel`.

Hakuyume: How about this https://github.com/yuyu2172/chainercv/pull/4 ?

chainercv/links/model/faster_rcnn/faster_rcnn_vgg.py
Hakuyume: Perhaps, `faster_rcnn_vgg16_voc07_2017_08_06.npz` is better. User will think "what is trial_4?" like me.

yuyu2172: OK. Thanks for you feedback.

yuyu2172:
This PR is no-compat because the child links of the FasterRCNN has changed.
It now uses `Conv2DActiv`.

examples/vgg/README.md
```diff
@@ -0,0 +1,13 @@
+# VGG
+
+For evaluation, please go to [`examples/classification`](https://github.com/chainer/chainercv/tree/master/examples).
+
+## Convert Caffe model
+Convert `*.caffemodel` to `*.npz`.
+
+```
+$ python caff2npz_vgg_16.py <source>.caffemodel <target>.npz
```
Hakuyume: Sorry, I forgot to change this line. Please update `caff2npz_vgg_16.py` -> `caffe2npz.py`.

examples/vgg/README.md
```diff
@@ -6,7 +6,7 @@ For evaluation, please go to [`examples/classification`](https://github.com/chai
 Convert `*.caffemodel` to `*.npz`.
 
 ```
-$ python caff2npz_vgg_16.py <source>.caffemodel <target>.npz
+$ python caff2npz.py <source>.caffemodel <target>.npz
```
Hakuyume: Typo. `caff` -> `caffe`.

PR #266, opened 2017-06-12
Fix a bug in Faster R-CNN

yuyu2172:


yuyu2172:
@Hakuyume 

I think this is critical.
Perhaps minor version update 0.5.1 is necessary.

PR #268, opened 2017-06-12
Update 0.4.11

yuyu2172:
Fix for a critical bug found in #266.


yuyu2172:
We found a better name for the Chainer v1 branch.

Issue #264, opened 2017-06-12
Add classification examples


yuyu2172:
Although, there are already classification models and examples in main Chainer, it still makes sense to add them to ChainerCV.
Here are reasons why:

1. Chainer's classification models support BGR inputs. It can be reimplemented in ChainerCV while making it consistent with other links here.
2. The example does not reproduce expected performance.
3. Their are no clear explanation on how to prepare imagenet. 
4. The examples in main Chainer can not use all the features that ChainerCV supports such as transforms.

Having said that, there are at minimum three things to do.

- [ ] Add links. Possibly, it can be located at `chainercv.links.model.classifcation`.
- [ ] Prepare `examples/imagenet`, which contains scripts to train and evaluate various classification models.
- [ ] We can use `chainer.datasets.LabeledImageDataset` for imagenet task. However, the current implementation does not force gray scale image to RGB image. This has to be fixed by sending PR to main Chainer. We can also write our own `image_dataset` inside ChainerCV.

yuyu2172:
To make this more concrete, I added a classification link that I would like to add to ChainerCV.
https://github.com/chainer/chainercv/pull/265

Hakuyume:
@yuyu2172 I agree with both 1 and 2. Classification links and datasets in Chainer don't follow the convention of ChainerCV (CHW, RGB). We need our own version (Of course, if we can change the models in Chainer, we don't need ChainerCV version). I think the ChainerCV implementation of fb.resnet.torch is very convenient.

> Add links. Possibly, it can be located at chainercv.links.model.classifcation.

This namespace is not consistent with other existing models. We don't have `chainercv.links.model.detection` or `chainercv.links.model.semantic_segmentation`.

yuyu2172:
> This namespace is not consistent with other existing models. We don't have 
 chainercv.links.model.detection or chainercv.links.model.semantic_segmentation.

OK.
How about `chainercv.links.model.vgg`, `chainercv.links.model.resnet` etc.
We can have different variants of VGGs under `chainercv.links.model.vgg`. For example, VGG16 and VGG19.
This seems better. Thanks for a suggestion.

yuyu2172:
> Of course, if we can change the models in Chainer, we don't need ChainerCV version
The change I would like to introduce (BGR->RGB and a lot of change in API) is too large. I think we can not hope to change the main Chainer code in the way I want.

Hakuyume:
[...]

PR #271, opened 2017-06-13
Add ImageFolderDataset

yuyu2172:
Following the discussion in #264, I made a dataset that obtains image classification dataset by just parsing the directory tree.

There can be some image that a user do not want to include in the dataset. For example, some datasets organize images in sub-directories according to classes, but they contain both color and depth images. The option `check_img_file` is helpful in this case.

Any suggestions are welcome as this class can have wide range of potential applications, which I have not probably covered.


chainercv/datasets/image_folder_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+import os
+
+import chainer
+from chainercv.utils import read_image
+
+
+def find_label_names(root):
+    label_names = [d for d in os.listdir(root)
+                   if os.path.isdir(os.path.join(root, d))]
+    label_names.sort()
```
Hakuyume: How about supporting numerical sort? Sometimes labels are given as numbers.

chainercv/datasets/image_folder_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+        for cur_dir, _, filenames in sorted(os.walk(label_dir)):
+            for filename in filenames:
+                if check_img_file(filename):
+                    img_paths.append(os.path.join(cur_dir, filename))
+                    labels.append(label_name_to_idx[label_name])
+
+    return img_paths, np.array(labels, np.int32)
+
+
+class ImageFolderDataset(chainer.dataset.DatasetMixin):
```
Hakuyume: How about `ClassificationDatasetFromDirectory`?  I think `<task>Dataset` is consistent with `DetectionDataset` and `SemanticSegmentationDataset`. Another choice is `DirectoryClassificationDataset` but it sounds that the task is classifying directories.

chainercv/datasets/image_folder_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+    """A data loader that loads images arranged in directory by classes.
```
Hakuyume: `data loader` -> `classification dataset`?
`directory` -> `directories`?

chainercv/datasets/image_folder_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+            root
+            |-- class_0
+            |   |-- img_0.png
+            |   |-- img_1.png
+            |
+            --- class_1
+                |-- img_0.png
+
+        >>> from chainercv.dataset import ImageFolderDataset
+        >>> dataset = ImageFolderDataset(root)
```
Hakuyume: `root` -> `'root'`

chainercv/datasets/image_folder_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+    label_names.sort()
+    return label_names
+
+
+def _ends_with_img_ext(filename):
+    img_extensions = [
+        '.jpg', '.JPG', '.jpeg', '.JPEG', '.png', '.PNG',
+        '.ppm', '.PPM', '.bmp', '.BMP',
+    ]
+    return any(filename.endswith(extension) for extension in img_extensions)
```
Hakuyume: How about using `os.path.splitext` and `str.lower()`?

chainercv/datasets/image_folder_dataset.py
yuyu2172: What kind of directory structure do you have in your mind?

chainercv/datasets/image_folder_dataset.py
yuyu2172: You are worried that this name may confuse users to think that there is a task "ImageFolder"?

yuyu2172: Oh. You want to say that this is a ClassificationDataset.

yuyu2172: `ClassificationDatasetFromDirectory` is OK, but this breaks the rule that dataset object class ends with "Dataset".

Perhaps, `FolderParsingClassificationDataset`, `DirectoryParsingClassificationDataset` or `ParsingClassificationDataset`.

Hakuyume: > Oh. You want to say that this is a ClassificationDataset.

Yes. `DirectoryParsingClassificationDataset` looks good to me.

yuyu2172:
I added `asset_is_classification_dataset`.

chainercv/datasets/image_folder_dataset.py
Hakuyume: Here is an example.
```
- root
    - 0
        - a.jpg
        - b.jpg
    - 1
        - c.jpg
    - 2
...
    - 10
        - y.jpg
        - z.jpg
```
In this case, user may want to sort directories numerical order (`0`, `1`, `2` ... `10`). Current code sort them by alphabetical order (`0`, `1`, `10`, `2`, ...).

yuyu2172: I see.

yuyu2172: Done

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+
+def parse_label_names(root, numerical_sort=False):
+    """Get label names from directories that are named by them.
+
+    The label names are names of the directories that locate a layer below the
+    root.
+
+    The label names can be used together with
+    :class:`chainercv.datasets.DirectoryParsingClassificationDataset`.
+    An index of the label names correspond to a label id
```
Hakuyume: `correspond` -> `corresponds`

chainercv/datasets/__init__.py
```diff
@@ -5,6 +5,8 @@
 from chainercv.datasets.cub.cub_keypoint_dataset import CUBKeypointDataset  # NOQA
 from chainercv.datasets.cub.cub_label_dataset import CUBLabelDataset  # NOQA
 from chainercv.datasets.cub.cub_utils import cub_label_names  # NOQA
+from chainercv.datasets.directory_parsing_classification_dataset import DirectoryParsingClassificationDataset  # NOQA
+from chainercv.datasets.directory_parsing_classification_dataset import parse_label_names  # NOQA
```
Hakuyume: `parse_label_names` is too ambiguous. If you want to import it under `chainercv.datasets`, the name should be more specific. How about `directory_parsing_label_names`? I think the name should be similar to `DirectoryParsingClassificationDataset` in order to show the relationship.

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+
+import chainer
+from chainercv.utils import read_image
+
+
+def parse_label_names(root, numerical_sort=False):
+    """Get label names from directories that are named by them.
+
+    The label names are names of the directories that locate a layer below the
+    root.
```
Hakuyume: `root` -> `root directory`

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+    else:
+        label_names = [int(name) for name in label_names]
+        label_names.sort()
+        label_names = [str(name) for name in label_names]
+    return label_names
+
+
+def _ends_with_img_ext(filename):
+    img_extensions = ['.jpg', '.jpeg', '.png', '.ppm', '.bmp']
+    return any(os.path.splitext(filename)[1].lower().endswith(extension) for
```
Hakuyume: You can use `==` instead of `endswith`.

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    label_names = [d for d in os.listdir(root)
+                   if os.path.isdir(os.path.join(root, d))]
+
+    if not numerical_sort:
+        label_names.sort()
+    else:
+        label_names = [int(name) for name in label_names]
+        label_names.sort()
+        label_names = [str(name) for name in label_names]
```
Hakuyume: How about using `key=` of `sorted`?

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+    return label_names
+
+
+def _ends_with_img_ext(filename):
```
Hakuyume: How about `_check_img_ext`?

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+    img_extensions = ['.jpg', '.jpeg', '.png', '.ppm', '.bmp']
+    return any(os.path.splitext(filename)[1].lower().endswith(extension) for
+               extension in img_extensions)
+
+
+def _parse_classification_dataset(root, label_names,
+                                  check_img_file=_ends_with_img_ext):
+    # Use label_name_to_idx for performance.
+    label_name_to_idx = {label_names[i]: i for i in range(len(label_names))}
```
Hakuyume: `{label_name: l for l, label_name in enumerate(label_names)}`

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+
+    img_paths = []
+    labels = []
+    for label_name in os.listdir(root):
+        label_dir = os.path.join(root, label_name)
+        if not os.path.isdir(label_dir):
+            continue
+
+        for cur_dir, _, filenames in sorted(os.walk(label_dir)):
```
Hakuyume: `numerical_sort` is not used for this sort. I don't think this is bad. However, current doc sounds `numerical_sort` is used anywhere.

chainercv/utils/testing/assertions/assert_is_classification_dataset.py
```diff
@@ -0,0 +1,50 @@
...
+        'img and label.'
+
+    img, label = example[:2]
+
+    assert_is_image(img, color=color)
+
+    assert isinstance(label, np.int32), \
+        'label must be a numpy.int32.'
+    assert label.ndim == 0, 'The ndim of label must be 0'
+    assert label.min() >= 0 and label.max() < n_class, \
```
Hakuyume: `min` and `max` are unnecessary because `label` is a scalar value.

docs/source/reference/datasets.rst
```diff
@@ -4,6 +4,14 @@ Datasets
 .. module:: chainercv.datasets
 
 
+DirectoryParsingClassificationDataset
+-------------------------------------
+.. autofunction:: DirectoryParsingClassificationDataset
```
Hakuyume: `autofunction` -> `autoclass`

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def _parse_classification_dataset(root, label_names,
+                                  check_img_file=_ends_with_img_ext):
+    # Use label_name_to_idx for performance.
+    label_name_to_idx = {label_names[i]: i for i in range(len(label_names))}
+
+    img_paths = []
+    labels = []
+    for label_name in os.listdir(root):
```
Hakuyume: Why don't you use `for label, label_name in enumerate(label_names)`? If we use this, we can remove `label_name_to_idx`.

tests/utils_tests/testing_tests/assertions_tests/test_assert_is_classification_dataset.py
```diff
@@ -0,0 +1,82 @@
...
+            {'dataset': InvalidImageDataset, 'valid': False},
+            {'dataset': InvalidLabelDataset, 'valid': False}
+        ],
+        [
+            {'color': False},
+            {'color': True}
+        ]
+    )
+))
+class TestAssertIsSemanticSegmentationDataset(unittest.TestCase):
```
Hakuyume: `SemanticSegmentation` -> `Classification`

chainercv/datasets/directory_parsing_classification_dataset.py
yuyu2172: I think it is better to use `numerical_sort` consistently.

yuyu2172: Thanks for your suggestion.

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+
+    The label names can be used together with
+    :class:`chainercv.datasets.DirectoryParsingClassificationDataset`.
+    An index of the label names corresponds to a label id
+    obtained from the dataset.
+
+    Args:
+        root (str): The root directory.
+        numerical_sort (bool): Label names are sorted numerically.
+            This means that :obj:`'2'` is before :obj:`10`,
```
Hakuyume: :obj:\`\`2\`\` -> :obj:\`2\`

tests/datasets_tests/test_directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,91 @@
...
+
+        for i in range(self.n_class):
+            class_dir = os.path.join(self.tmp_dir, 'class_{}'.format(i))
+            os.makedirs(class_dir)
+            for j in range(self.n_img_per_class):
+                _save_img_file(
+                    os.path.join(class_dir,
+                                 'img{}.{}'.format(j, self.suffix)),
+                    self.size, self.color)
+            open(os.path.join(class_dir, 'dummy_file.XXX'), 'a').close()
```
Hakuyume: Current code of `DirectoryParsingClassificationDataset` supports a nested directory tree. How about testing that situation?

tests/datasets_tests/test_directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def test_directory_parsing_classification_dataset(self):
+        dataset = DirectoryParsingClassificationDataset(
+            self.tmp_dir, color=self.color)
+
+        self.assertEqual(len(dataset), self.n_img_per_class * self.n_class)
+
+        assert_is_classification_dataset(
+            dataset, self.n_class, color=self.color)
```
Hakuyume: How about checking the number of items and their order?

yuyu2172: What do you mean by "the number of items"?
By the way, the length of the dataset is checked.

yuyu2172:
I stopped using `*_paths` to represent file paths.
Instead, I used `*_filenames` that is used in other part of the library.

tests/datasets_tests/test_directory_parsing_classification_dataset.py
Hakuyume: I mean "the length of the dataset". As you pointed out, it is already checked. Sorry.
How about the order of elements?

Hakuyume: > How about the order of elements?
Sorry, you have already added it. Thank you.

tests/datasets_tests/test_directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,122 @@
...
+
+        for i in range(self.n_class):
+            class_dir = os.path.join(self.tmp_dir, '{}'.format(i))
+            os.makedirs(class_dir)
+            _save_img_file(os.path.join(class_dir, 'img_0.png'),
+                           (48, 32), color=True)
+
+    def test_numerical_sort(self):
+        dataset = DirectoryParsingClassificationDataset(
+            self.tmp_dir, numerical_sort=False)
```
Hakuyume: `numerical_sort=False` -> `numerical_sort=True`
With this modification, nosetests will find the bug of `DirectoryParsingClassificationDataset`

Hakuyume:
@yuyu2172 I'm sorry, I kept you waiting so long. I found a bug in your code but I couldn't understand why nosetests passed. Now I find test code also has a bug. Please check my comment.

yuyu2172:
OK. I will look into it.
Thank you for checking.

yuyu2172:
@Hakuyume 

For strings except label names (file name and directory other than the top directory), it seems better to avoid numerical sort even if `numerical_sort=True`.

In order to carry out numerical sort, we need to assume that these values are all numbers, which is usually not the case.

Hakuyume:
> @Hakuyume
>
> For strings except label names (file name and directory other than the top directory), it seems better to avoid numerical sort even if numerical_sort=True.
>
> In order to carry out numerical sort, we need to assume that these values are all numbers, which is usually not the case.

I agree with you. This is the reason why I said applying numerical sort only to label names is not bad (https://github.com/chainer/chainercv/pull/271#discussion_r125258650)

chainercv/datasets/directory_parsing_classification_dataset.py
```diff
@@ -0,0 +1,135 @@
...
+    Args:
+        root (str): The root directory.
+        check_img_file (callable): A function to determine
+            if a file should be included in the dataset.
+        color (bool): If :obj:`True`, this dataset read images
+            as color images.
+        numerical_sort (bool): Label names are sorted numerically.
+            This means that label :obj:`2` is before label :obj:`10`,
+            which is not the case when string sort is used.
+            Regardless of this option, non-numerical sort is used for the
```
Hakuyume: You are using `string sort` and `non-numerical sort`. It sounds there are three types of sorting.

PR #278, opened 2017-06-14
Add assertions

Hakuyume:
merge after #177


yuyu2172:
I prefer `np.testing.assert_equal(X, Y)` over `assert X == Y` because it prints the actual values of X and Y when the test fails.

Hakuyume:
> I prefer np.testing.assert_equal(X, Y) over assert X == Y because it prints the actual values of X and Y when the test fails.

I think it is better to show the meaning of the test. How do you think of it?

Hakuyume:
I think the dump of images isn't helpful for debugging because it is usually a large array.

yuyu2172:
> I think it is better to show the meaning of the test. How do you think of it?

You can still print err_msg with `np.testing`.

> I think the dump of images isn't helpful for debugging because it is usually a large array.

The assertion happens not only with image-arrays, but also with integers and floats.
e.g. https://github.com/Hakuyume/chainercv/blob/b75f8ae8349cd9fdc5ddb8ecd6ee83e95ef1815b/chainercv/utils/testing/assertions/assert_is_image.py#L30

Hakuyume:
> You can still print err_msg with np.testing.

I didn't know that. Thank you. I will replace with `np.testing.~'.

Hakuyume:
> The assertion happens not only with image-arrays, but also with integers and floats.
e.g. https://github.com/Hakuyume/chainercv/blob/b75f8ae8349cd9fdc5ddb8ecd6ee83e95ef1815b/chainercv/utils/testing/assertions/assert_is_image.py#L30

How about showing actual values only for integers or floats? I think using `format` to make an error message is enough.

yuyu2172:
So you are going to use `assert` for something?

Hakuyume:
> So you are going to use assert for something?

Yes. Some tests are not provided in numpy. For example, `isinstance` and `less_or_equal`.
I need to use `assert` for them.

yuyu2172:
I learned that when run together with nosetests, the actual values are printed out with `assert`.
I think `assert` is OK.
Sorry for the confusion.

Hakuyume:
#251 and #252 can be write with `assert_is_semantic_segmentation_dataset`.

Hakuyume:
I merged #251 and #252 into this PR.
@yuyu2172 I think those two PRs are no longer needed. Could you close them?

yuyu2172:
OK!

chainercv/utils/testing/assertions/assert_is_bbox.py
```diff
@@ -0,0 +1,32 @@
+import numpy as np
+
+
+def assert_is_bbox(bbox, size=None):
+    """Checks if bounding boxes satisfy bounding box format.
+
+    This function checks if given bounding boxes satisfy bounding boxe
```
yuyu2172: typo: boxe

chainercv/utils/testing/assertions/assert_is_bbox.py
```diff
@@ -0,0 +1,32 @@
...
+
+    Args:
+        bbox (~numpy.ndarray): Bounding boxes to be checked.
+        size (tuple of ints): The size of canvas.
+            If this argument is specified,
+            Each bounding box should be within the canvas.
+    """
+
+    assert isinstance(bbox, np.ndarray), \
+        'bbox must be a numpy.ndarray.'
```
yuyu2172: I always had this feeling that "should" is better than "must".
When there is "must" in an error message, I feel like I am scolded.

chainercv/utils/testing/assertions/assert_is_bbox.py
```diff
@@ -0,0 +1,32 @@
...
+    """Checks if bounding boxes satisfy bounding box format.
+
+    This function checks if given bounding boxes satisfy bounding boxe
+    format or not.
+    If the bounding boxes do not satifiy the format, this function raises an
+    :class:`AssertionError`.
+
+    Args:
+        bbox (~numpy.ndarray): Bounding boxes to be checked.
+        size (tuple of ints): The size of canvas.
```
yuyu2172: How about using "image".

So this sentence should be "The size of an image.".

chainercv/utils/testing/assertions/assert_is_bbox.py
```diff
@@ -0,0 +1,32 @@
...
+    assert bbox.shape[1:] == (4,), \
+        'The shape of bbox must be (*, 4).'
+    assert (bbox[:, 0] < bbox[:, 2]).all(), \
+        'The coordinate of top must be less than that of bottom.'
+    assert (bbox[:, 1] < bbox[:, 3]).all(), \
+        'The coordinate of left must be less than that of right.'
+
+    if size is not None:
+        assert (bbox[:, :2] >= 0).all() and (bbox[:, 2:] < size).all(),\
+            'The coordinates of bbox should not exceed the size of canvas.'
```
yuyu2172: of bounding boxes

chainercv/utils/testing/assertions/assert_is_bbox.py
yuyu2172: Nevermind.
I found that "must" is used in NumPy doc.

chainercv/utils/testing/assertions/assert_is_image.py
```diff
@@ -0,0 +1,36 @@
...
+def assert_is_image(img, color=True, check_range=True):
+    """Checks if an image satisfies image format.
+
+    This function checks if a given image satisfies image format or not.
+    If the image does not satifiy the format, this function raises an
+    :class:`AssertionError`.
+
+    Args:
+        img (~numpy.ndarray): An image to be checked.
+        color (bool): A boolean that determines the behaviour of
```
yuyu2172: determines the expected channel size.

tests/datasets_tests/voc_tests/test_voc_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,29 @@
...
+    {'split': 'val'},
+    {'split': 'trainval'}
+)
+class TestVOCSemanticSegmentationDataset(unittest.TestCase):
+
+    def setUp(self):
+        self.dataset = VOCSemanticSegmentationDataset(split=self.split)
+
+    @attr.slow
+    def test_camvid_dataset(self):
```
yuyu2172: sorry. method name is wrong.

chainercv/utils/testing/assertions/assert_is_bbox.py
yuyu2172: Although it is different from the current implementation of generate_random_bbox, I think it is ok to use inclusive upper bound instead of exclusive.

yuyu2172:
Nice work. I finished my comments.

Hakuyume:
@yuyu2172 Thank you for reviewing. I reflected your comments.

yuyu2172:
Are you planning to use these functions for testing Links? (i.e. use it for a test to check DetectionLink).
It is possible to leave it as a future work.

Hakuyume:
> Are you planning to use these functions for testing Links? (i.e. use it for a test to check DetectionLink).
It is possible to leave it as a future work.

Actually, I am writing `assert_is_detection_link`. But I will make another PR for it.

yuyu2172:
OK nice.
LGTM

PR #287, opened 2017-06-16
Show dialog before download

### LINKED ISSUE #286: Notify user about total dataset size and download location
Currently, e.g. when downloading VOC (~4GB) for examples/faster_rcnn/ there is no notice how big is the total download and where the file will be stored on disk ($HOME/.chainer/dataset/pfnet/chainercv/hoge). It might be more user-friendly to include such information to help users avoid full hdds.

Hakuyume:
Fix #286 

```
Fetching the size of http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar ...
File will be saved to /home/admin/.chainer/dataset/_dl_cache/92029705a99338d0932803388148a725.
It will be use 1907.00 MiB of the disk space.
Proceed? (y/N): y
Downloading from http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar ...
...
```

### DIFF
--- a/chainercv/utils/download.py
+++ b/chainercv/utils/download.py
@@ -32,7 +32,7 @@ def _reporthook(count, block_size, total_size):
     percent = progress_size / total_size * 100
     sys.stdout.write(
         '\r... {:.0f} %, {:.2f} MiB / {:.2f} MiB, '
-        '{:.2f} KiB/s, {:.1f} seconds passed, {:.1f} seconds remains'
+        '{:.2f} KiB/s, {:.1f} seconds passed, {:.1f} seconds remain'
         .format(
             percent, progress_size / (1 << 20), total_size / (1 << 20),
             speed / (1 << 10), duration, (total_size - progress_size) / speed))
--- a/chainercv/utils/download.py
+++ b/chainercv/utils/download.py
@@ -1,4 +1,6 @@
+from __future__ import division
 from __future__ import print_function
+
 import hashlib
 import os
 import shutil
@@ -67,10 +69,28 @@ def cached_download(url):
         if os.path.exists(cache_path):
             return cache_path
 
+    print('Fetching the file size from {:s} ...'.format(url))
+    req = request.Request(url=url, method='HEAD')
+    with request.urlopen(req) as res:
+        total = int(res.getheader('Content-Length'))
+
+    print('Downloaded file will be saved to {:s} .'.format(cache_path))
+    print('This will be use {:.2f} MiB of the disk space.'.format(
+        total / (1 << 20)))
+
+    while True:
+        b = input('Proceed? (y/N): ')
+        if b in {'y', 'Y'}:
+            break
+        elif b in {'', 'n', 'N'}:
+            raise KeyboardInterrupt
+        else:
+            print('Please answer \'y\' or \'n\'.')
+
     temp_root = tempfile.mkdtemp(dir=cache_root)
     try:
         temp_path = os.path.join(temp_root, 'dl')
-        print('Downloading from {}...'.format(url))
+        print('Downloading from {} ...'.format(url))
         request.urlretrieve(url, temp_path, _reporthook)
         with filelock.FileLock(lock_path):
             shutil.move(temp_path, cache_path)
--- a/chainercv/utils/download.py
+++ b/chainercv/utils/download.py
@@ -70,13 +70,13 @@ def cached_download(url):
         if os.path.exists(cache_path):
             return cache_path
 
-    print('Fetching the file size from {:s} ...'.format(url))
+    print('Fetching the file size of {:s} ...'.format(url))
     req = request.Request(url=url, method='HEAD')
     with request.urlopen(req) as res:
         total = int(res.getheader('Content-Length'))
 
-    print('Downloaded file will be saved to {:s}.'.format(cache_path))
-    print('This will be use {:.2f} MiB of the disk space.'.format(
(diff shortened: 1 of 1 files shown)


yuyu2172:
`request.Request` does not have `method` argument for Python2.

Also, I would be against asking users to confirm download.
The user has already understood that download will happen when necessary.
If unintended downloading starts, the user can just ctrl-c.

chainercv/utils/download.py
```diff
@@ -51,12 +55,12 @@ def cached_download(url):
         str: Path to the downloaded file.
 
     """
-    cache_root = get_dataset_directory('_dl_cache')
+    cache_root = os.path.join(get_dataset_root(), '_dl_cache')
```
yuyu2172: Why is this change necessary?

yuyu2172: Oh, the following try, except were not necessary.

Hakuyume: > Why is this change necessary?

https://github.com/chainer/chainer/pull/2839#issuecomment-308057262

Hakuyume:
> Also, I would be against asking users to confirm download.
The user has already understood that download will happen when necessary.
If unintended downloading starts, the user can just ctrl-c.

We can divide this problem into two.
1. `cache_download` does not tell the size of file and the location where the file will be downloaded. I guess this is the issue reported in #286. I think there is no bad point about showing these information.
2. The download process starts without confirmation. As you suggested, users can stop this by `Ctrl-C`. However, I thought it is more user-freindly to ask explicitly (I'm not sure).

chainercv/utils/download.py
yuyu2172: I see.

yuyu2172:
> However, I feel it is more user friendly if it asks explicitly

I find it irritating to find my process hang because of this feature.
This is especially the case because this behavior is different from Chainer's download.

Hakuyume:
> I find it irritating to find my process hang because of this feature.
This is especially the case because this behavior is different from Chainer's download.

OK, I removed confirmation step and changed the progress message as follows.
```
Downloading from http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar to ***/.chainer/dataset/_dl_cache/92029705a99338d0932803388148a725 ...
... 0 %, 0.16 MiB / 1907.00 MiB, 195.06 KiB/s, 0.8 seconds passed
```

yuyu2172:
OK. Thanks.
Perhaps, we can add the remaining time.

Hakuyume:
> Perhaps, we can add the remaining time.

I added.
```
Downloading from http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar to ***/.chainer/dataset/_dl_cache/92029705a99338d0932803388148a725 ...
... 9 %, 172.07 MiB / 1907.00 MiB, 1145.80 KiB/s, 153.8 seconds passed, 1550.5 seconds remains
```

yuyu2172:
On downloading message, I think it is too verbose.

Four suggestions

+ Decimal numbers are not necessary (224.1MB/s --> 224MB/s)
+ I don't think elapsed time is necessary.
+ How about using "eta <TIME LEFT>" instead of <TIME LEFT> seconds remain. "eta" stands for expected time of arrival, and this is abbreviation used by wget.
+ How about using minutes and seconds instead of seconds for "eta"?

Here is a code I suggest to use from line 32.
```python
    eta = (total_size - progress_size) / speed
    sys.stdout.write(
        '\r... {:.0f}%  {:.0f}MiB/{:.0f}MiB '
        '{:.0f}KiB/s eta {:.0f}m {:.0f}s'
        .format(
            percent, progress_size / (1 << 20), total_size / (1 << 20),
            speed / (1 << 10), eta // 60, eta % 60))
```

yuyu2172:
This is something different from what I have said in the last comment, but how about switching eta to `hours:minutes:seconds`, which can be done with the string below.
Also, I didn't like how outputs length changes from frame to frame.
What do you think about the new output?

```python
    sys.stdout.write(
        '\r... {:3.0f}%  Total {:7.0f}Mib    Current {:7.0f}MiB '
        '{:8.0f}KiB/s    eta {:5.0f}:{:02.0f}:{:02.0f}'    
        .format(
            percent, total_size / (1 << 20), progress_size / (1 << 20),
            speed / (1 << 10), eta // 3600, eta // 60, eta % 60))
```


```...   2%  Total     178Mib    Current       3MiB      570KiB/s    eta     0:05:15
```

Hakuyume:
> I tweaked more, and I found that it looks nice to report only percentage, speed and eta.

From my understanding, the original issue is that users can not know the total dataset size #286. I think it is better to show the total size and downloaded size.

Hakuyume:
> Also, I didn't like how outputs length changes from frame to frame.

I don't  like either. I accept your modification about the length.

Hakuyume:
@yuyu2172 How about this output?
```
Downloading ...
From: http://host.robots.ox.ac.uk/pascal/VOC/voc2007/VOCtest_06-Nov-2007.tar
To: ***/.chainer/dataset/_dl_cache/c320d1dc0cd031efc99e256bf21d57a6
  %   Total    Recv       Speed  Time left
  1  430MiB    5MiB   1634KiB/s    0:04:26
```
This format is inspired by curl command.

yuyu2172:
LGTM

PR #289, opened 2017-06-19
Add cupy as a dependency for chainer2

zori:
Chainer 2 requires cupy as a separate dependency for CUDA support. It can work without CUDA support in CPU mode, so I'm not updating `environment_minimum.yml`.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -11,4 +11,5 @@ dependencies:
   - numpy
   - Pillow
   - pip:
+    - cupy
     - chainer==2.0


Hakuyume:
`environment.yml` should not contain `cupy` because it is used in Travis CI, which does not have CUDA. If we add cupy as dependency, Travis CI may fail.

zori:
I see. Then how about mentioning it in the ChainerCV build instructions http://chainercv.readthedocs.io/en/stable/#anaconda? I assume the pip installation also correctly fetches cupy, so probably that's the only place that would be needed.

Hakuyume:
`pip` doesn't install `cupy`, either (see below). So, we also need a note for `pip`. Could you add them (for pip and anaconda)?

```
$ pip list
numpy (1.13.1)
pip (9.0.1)
setuptools (36.2.7)
$ pip install chainercv
$ pip list
chainer (2.0.2)                                                                                                                                                                                                                              
chainercv (0.6.0)                                                                                                                                                                                                                            
filelock (2.0.11)                                                                                                                                                                                                                            
mock (2.0.0)                                                                                                                                                                                                                                 
nose (1.3.7)                                                                                                                                                                                                                                 
numpy (1.13.1)                                                                                                                                                                                                                               
olefile (0.44)                                                                                                                                                                                                                               
pbr (3.1.1)
Pillow (4.2.1)
pip (9.0.1)
protobuf (3.3.0)
setuptools (36.2.7)
six (1.10.0)
```

zori:
Just to confirm, you'd rather only add the info in documentation (and leave the default installation possibly for CPU-only use case), correct? I'll modify this PR if that's the case

Hakuyume:
Yes. Please restore the original `environment.yml` and add the info to docs.

PR #293, opened 2017-06-21
Add GPU mode for resize transform

yuyu2172:
### Demo

```python
from chainercv.transforms import resize
import numpy as np
import chainer
from chainercv.visualizations import vis_image

src = chainer.cuda.to_gpu(astronaut().transpose(2, 0, 1).astype(np.float32))
size = tuple(np.random.randint(224, 480, size=(2,)))
dst = resize(src, size)
vis_image(dst.get())
plt.show()
```

### benchmark

```python
from chainercv.transforms import resize
import numpy as np
import chainer


src = chainer.cuda.to_gpu(astronaut().transpose(2, 0, 1).astype(np.float32))
src_cpu = src.get()
size = (224, 280)

def benchmark(func, args, n_run):
    import cupy as cp
    times = []
    for _ in range(n_run):
        start = cp.cuda.Event()
        end = cp.cuda.Event()
        start.record()
        func(*args)
        end.record()
        end.synchronize()
        times.append(cp.cuda.get_elapsed_time(start, end))  # milliseconds
    return times

# dry runs
for _ in range(3):
    dst = resize(src, size)

print('gpu {}ms'.format(np.mean(benchmark(resize, (src, size), 10))))


import time
start = time.time()
resize(src_cpu, size)
print('cpu {}ms'.format((time.time() - start) * 1000))
```

#### benchmark result
```
gpu 0.0970143988729ms 
cpu 28.1879901886ms   
```


Hakuyume:
How about making another PR for `cupy -> cp`?

yuyu2172:
OK #316

yuyu2172:
I changed `resize_gpu` to launch a thread for each channel.
On my setting, the kernel execution time improved slightly.
```
Previously (Quadro M1000M)
 42.21%  725.38us        13  55.798us  52.833us  65.889us  resize_linear
New (Quadro M1000M)
 34.86%  710.09us        13  54.622us  52.960us  58.497us  resize_linear

Previous (TitanX)
 10.98%  133.99us        13  10.306us  9.2800us  12.768us  resize_linear
New (TitanX)
 10.30%  127.14us        13  9.7800us  8.9610us  11.744us  resize_linear

```
I used the benchmark in the previous comments.

yuyu2172:
I will temporarily close this because we do not have a plan to include this in ChainerCV.

PR #306, opened 2017-06-30
Support (C, 1, 1) for fill argument

Hakuyume:


chainercv/transforms/image/random_expand.py
```diff
@@ -24,6 +24,9 @@ def random_expand(img, max_ratio=4, fill=0, return_param=False):
             paper, this value is 4.
         fill (float, tuple or ~numpy.ndarray): The value of padded pixels.
```
yuyu2172: float or ~numpy.ndarray

chainercv/transforms/image/resize_contain.py
```diff
@@ -19,6 +19,9 @@ def resize_contain(img, size, fill=0, return_param=False):
         size (tuple of two ints): A tuple of two elements:
             :obj:`height, width`. The size of the image after resizing.
         fill (float, tuple or ~numpy.ndarray): The value of padded pixels.
```
yuyu2172: ditto

chainercv/transforms/image/random_expand.py
Hakuyume: I think supporting tuple is convenient. Why do you want to drop it?

yuyu2172: Let me summarize.

`fill` can be one of `float, tuple or (C, 1, 1)`.
`mean` is `(C, 1, 1)`.

Hakuyume: > fill can be one of float, tuple or (C, 1, 1).
mean is (C, 1, 1).

Yes. And I want to pass a tuple value that is never broadcasted to an image.

yuyu2172:
LGTM

PR #312, opened 2017-07-04
Use cupy.ndarray.sort

Hakuyume:
Merge after #206


Hakuyume:
@yuyu2172 In Faster R-CNN code, why do you send all arrays to cpu once? Can I rewrite this code to conduct calculations on GPU?
https://github.com/chainer/chainercv/blob/master/chainercv/links/model/faster_rcnn/utils/proposal_creator.py#L109-L111

yuyu2172:
@Hakuyume 

For some reason, I thought that the code will be significantly slower on GPU.
I can't find the benchmark I used, and I think the speed difference will be very small compared to time it takes to do convolutions.

Hakuyume:
I merged this PR into #314

PR #317, opened 2017-07-05
Fix divide by zero

tkerola:
This PR fixes two divide by zero errors.

### DIFF
--- a/chainercv/evaluations/eval_detection_voc.py
+++ b/chainercv/evaluations/eval_detection_voc.py
@@ -231,7 +231,7 @@ def calc_detection_voc_prec_rec(
         fp = np.cumsum(match_l == 0)
 
         prec[l] = tp / np.maximum(fp + tp, np.finfo(np.float64).eps)
-        rec[l] = tp / n_pos[l]
+        rec[l] = tp / max(1, n_pos[l])
 
     return prec, rec
 
--- a/chainercv/utils/bbox/bbox_iou.py
+++ b/chainercv/utils/bbox/bbox_iou.py
@@ -37,7 +37,9 @@ def bbox_iou(bbox_a, bbox_b):
     # bottom right
     br = xp.minimum(bbox_a[:, None, 2:], bbox_b[:, 2:])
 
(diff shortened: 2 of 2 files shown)


yuyu2172:
Thanks for PR.
Can you remove changes other than `eval_detection_voc.py` and `bbox_iou.py`?

tkerola:
Hi, I removed irrelevant commits.

Hakuyume:
@tkerola I think we should not hide zero division error by using `eps` or `1` as a denominator. For example, if the area of the union of two boxes is zero, one of the boxes is invalid (`invalid` means the area is zero or negative). I think it is better to notify users that there is an invalid box.

tkerola:
@Hakuyume I see. I agree with your point. Then it would be better to display a warning when this happens in `bbox_iou`.

As for the `max(1, n_pos[l])` fix, `n_pos[l]` will be zero if all gt bboxes for a certain class are marked as difficult, which will result in the map score being NaN. Do you think that we can keep the fix in this case?

Hakuyume:
> As for the max(1, n_pos[l]) fix, n_pos[l] will be zero if all gt bboxes for a certain class are marked as difficult, which will result in the map score being NaN. Do you think that we can keep the fix in this case?

As you pointed out, `n_pos[l]` can be zero if all ground truth boxes are marked as difficult and this is a valid situation. How about setting recall of the class to `None`? In this case, we need modify `calc_detection_voc_ap` to accept `None`.
Also, I think the computation of `prec` can be improved. How about setting a element of `prec` to `np.nan` if the corresponding element of `tp + fp` is zero?

Hakuyume:
I made another PR for this issue because my request was too complicated #322.

yuyu2172:
@tkerola 
Hakuyume and I had a discussion privately, and concluded that more change is required on top of this PR.
Since the change involves code outside of the scope you modified, so we thought it was easier to make a new PR #322.
Is it OK to close this?

yuyu2172:
The #322 has been merged.
Thank you for shedding light on this issue.

PR #321, opened 2017-07-06
Add ResNet

yuyu2172:
Merge after #265  (DONE)

EDIT:
merge after #427 (DONE)


yuyu2172:
I did a quick survey on various architectures that are called ResNet.
There seems to be three.

1. Original architecture (https://arxiv.org/pdf/1512.03385.pdf)
2. Facebook's ResNet.
+ The difference between the original architecture is with "strided convolution" ([link](https://github.com/facebook/fb.resnet.torch#notes)).
+ TensorFlow calls this ResNet v1. They have no implementation for the original architecture in the official repository (?) [link](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/slim/python/slim/nets/resnet_v1.py)
3. ResNet v2
+ Introduced in https://arxiv.org/pdf/1603.05027.pdf. This architecture uses BNActivConv2D instead of Conv2DBNActiv as a basic building block.
+  I am not sure if the name "ResNet v2" is common, but it is used by people using [Tensorflow](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/slim/python/slim/nets/resnet_v2.py)  and [MXNet](https://github.com/tornadomeet/ResNet).
+ The performance is not always better than the counterparts. https://github.com/tornadomeet/ResNet#resnet-v2-vs-resnet-v1

# Names
I think naming the ResNet class as `ResNet***` (e.g. `ResNet50`) is good.
+ We do not need to support ResNet v2 because it seems to be unpopular compared to the other variants.
+ We can support both the original architecture and FB ResNet in one class by switching between the two using a variable. The logic is relatively simple.

Hakuyume:
The third model is called `Pre-ResNet` in torch (https://github.com/facebook/fb.resnet.torch/blob/master/models/preresnet.lua).

yuyu2172:
Thanks. I added that information to the summary.

yuyu2172:
@Hakuyume 

Can you briefly take a look at `resnet.py`?
https://github.com/yuyu2172/chainercv/blob/9288eb869a64f6f17aaf8dbf90f4206faf6cfc42/chainercv/links/model/resnet/resnet.py

Note that the uploaded pretrained weight would not work with the current organization of weights.

yuyu2172:
Based on the discussion conducted offline, here are the changes to be made.

+ Change class of BuildingBlock to `chainer.Chain`. There is no clear guideline to decide whether to use `chainer.Chain` or `chainer.ChainList`.
+ Change a constructor option of `ResNet` to take `mode (str)` instead of `fb_resnet (bool)`. This is better because there are at least three ways to define structure of bottleneck: `kaiming`, `fb` and `preres`.
+ BN initial parameters should follow the default values instead of the custom values proposed by ResNet 1 hour paper. The custom initialization scheme can be implemented independently in a training script for reproducing that paper.

yuyu2172:
@Hakuyume 
Can you check this?

yuyu2172:
`mode` sounds like switching between inference and training.
`arch` is better.

yuyu2172:
@Hakuyume 
Could you check this?

Hakuyume:
@yuyu2172 OK, I'll check. Please fix the style first.

yuyu2172:
Oh. Sorry about that.

chainercv/links/model/resnet/resblock.py
```diff
@@ -0,0 +1,107 @@
...
+        with self.init_scope():
+            self.conv1 = Conv2DBNActiv(in_channels, mid_channels, 1,
+                                       first_stride, 0, initialW=initialW,
+                                       nobias=True)
+            self.conv2 = Conv2DBNActiv(mid_channels, mid_channels, 3,
+                                       second_stride, 1, initialW=initialW,
+                                       nobias=True)
+            self.conv3 = Conv2DBNActiv(mid_channels, out_channels, 1, 1, 0,
+                                       initialW=initialW, nobias=True,
+                                       activ=lambda x: x)
```
Hakuyume: Calling `Caonv2DBActiv` without activ looks tricky. How about adding `Conv2DBN`? (It is ok to make it a private class)

chainercv/links/model/resnet/resblock.py
```diff
@@ -0,0 +1,107 @@
...
+    """A bottleneck layer.
+
+    Args:
+        in_channels (int): The number of channels of input arrays.
+        mid_channels (int): The number of channels of intermediate arrays.
+        out_channels (int): The number of channels of output arrays.
+        stride (int or tuple of ints): Stride of filter application.
+        initialW (4-D array): Initial weight value used in
+            the convolutional layers.
+        conv_shortcut (bool): If :obj:`True`, apply a 1x1 convolution
```
Hakuyume: How about `residual_conv`? I prefer `residual` to `shortcut`. Using both `residual` and `shortcut` is confusing and `residual` is more common.

chainercv/links/model/resnet/resnet.py
```diff
@@ -0,0 +1,281 @@
...
+# This is channel wise mean of mean image distributed at
+# https://github.com/KaimingHe/deep-residual-networks
+_imagenet_mean = np.array(
+    [123.15163084, 115.90288257, 103.0626238],
+    dtype=np.float32)[:, np.newaxis, np.newaxis]
+
+
+class ResNet(PickableSequentialChain):
+
+    """Base class for ResNet Network.
```
Hakuyume: `ResNet architecture` is better because `ResNet Network` is `Residual Network Network`.

chainercv/links/model/resnet/resnet.py
```diff
@@ -0,0 +1,281 @@
...
+
+    This is a feature extraction link.
```
Hakuyume: `pickable sequential link`? We have not defined `feature extraction link` officially.

chainercv/links/model/resnet/resnet.py
```diff
@@ -0,0 +1,281 @@
...
+    loaded from weights distributed on the Internet.
+    The list of pretrained models supported are as follows:
+
+    * :obj:`imagenet`: Loads weights trained with ImageNet and distributed \
+        at `Model Zoo \
+        <https://github.com/BVLC/caffe/wiki/Model-Zoo>`_.
+        This is only supported when :obj:`arch=='he'`.
+
+    Args:
+        model_name (str): Name of the resnet model to instantiate.
```
Hakuyume: How about `n_layer`? It takes one of `50, 101, 152` as integer. `model_name` is difficult to understand.

chainercv/links/model/resnet/resnet.py
```diff
@@ -0,0 +1,281 @@
...
+            where :obj:`$CHAINER_DATASET_ROOT` is set as
+            :obj:`$HOME/.chainer/dataset` unless you specify another value
+            by modifying the environment variable.
+        mean (numpy.ndarray): A mean value. If :obj:`None`,
+            the default values are used.
+            If a supported pretrained model is used,
+            the mean value used to train the pretrained model is used.
+            Otherwise, the mean value calculated from ILSVRC 2012 dataset
+            is used.
+        initialW (callable): Initializer for the weights.
```
Hakuyume: `weights for convolution kernels`?

docs/source/reference/links.rst
```diff
@@ -20,6 +20,7 @@ Feature extraction links extract feature(s) from given images.
 .. toctree::
 
    links/vgg
+   links/resnet
```
Hakuyume: alphabetical order

chainercv/links/model/resnet/resblock.py
yuyu2172: Hmm...
I think the current version is fine, but alternatively I can explicitly use `conv` and `bn`.

I do not like the idea of adding a private class. This usages come up a lot.
I am less hesitant on the idea of adding `Conv2DBN`, but personally it looks redundant given that this is just a `Conv2DBNActiv` with no activation....

Hakuyume: If you don't like adding a private class, using `Conv2D` + `BatchNormalization` is better.
Another solution is adding `active='no'` option. to `Conv2DBNActiv`.

Hakuyume: Since the default value of `active` is `chainer.functions.relu` (https://github.com/chainer/chainercv/blob/master/chainercv/links/connection/conv_2d_bn_activ.py#L65), we can use `active=None` for no activation.

yuyu2172: OK

yuyu2172: updated

yuyu2172:
@Hakuyume

chainercv/links/model/resnet/resblock.py
```diff
@@ -0,0 +1,107 @@
...
+            x = l(x)
+        return x
+
+
+class Bottleneck(chainer.Chain):
+
+    """A bottleneck layer.
+
+    Args:
+        in_channels (int): The number of channels of input arrays.
```
Hakuyume: `input arrays` -> `(the) input array`? From my understanding, this link takes only one array.

chainercv/links/model/resnet/resblock.py
```diff
@@ -0,0 +1,107 @@
...
+        mid_channels (int): The number of channels of intermediate arrays.
+        out_channels (int): The number of channels of output arrays.
```
Hakuyume: ditto.

examples/classification/README.md
```diff
@@ -2,15 +2,30 @@
...
+| ResNet152 | 23.2 % | 23.0 % [2] |
+
+Ten crop error rate.
+
+| Model | Top 1 | Reference Top 1 |
+|:-:|:-:|:-:|
+| VGG16 | 27.1 % |   |
+| ResNet50 | 23.0 % | 22.9 % [2] |
+| ResNet101 |21.8 % | 21.8 % [2] |
+| ResNet152 |21.4 % | 21.4 % [2] |
```
Hakuyume: `| 21.4 %` (space after `|`).

yuyu2172:
Thanks.

chainercv/links/model/resnet/resblock.py
```diff
@@ -0,0 +1,107 @@
...
+                    out_channels, mid_channels, out_channels, stride=1,
+                    initialW=initialW, residual_conv=False)
+                self.add_link(name, bottleneck)
+                self._forward.append(name)
+
+    def __call__(self, x):
+        for name in self._forward:
+            l = getattr(self, name)
+            x = l(x)
+        return x
```
Hakuyume: I prefer using `PickableSequentialChain` to managing `_forward` manually. Although the pickable feature is not used, it will be more simple.

knorth55:
@yuyu2172 I have a question. Some repositories like mxnet and https://github.com/KaimingHe/deep-residual-networks, `he` resnet is `nobias=True` on `conv1`.
Which one is correct?

https://github.com/KaimingHe/deep-residual-networks/blob/master/prototxt/ResNet-101-deploy.prototxt#L18
https://github.com/apache/incubator-mxnet/blob/master/example/image-classification/symbols/resnet-v1.py#L123-L124

PR #322, opened 2017-07-06
fix zero division

Hakuyume:
Original PR: #317


yuyu2172:
Good PR.
Can you add a unittest for `calc_detection_voc_prec_rec`?

Hakuyume:
> Can you add a unittest for calc_detection_voc_prec_rec?

https://github.com/chainer/chainercv/blob/master/tests/evaluations_tests/test_eval_detection_voc.py#L119-L150

yuyu2172:
I mean, could you add a test case that is relavant to this change?
(There is gt bbox, but all of them are difficult).

Hakuyume:
> I mean, could you add a test case that is relavant to this change?
(There is gt bbox, but all of them are difficult).

I see. OK, I will add it.

yuyu2172:
LGTM

PR #345, opened 2017-07-19
Show in legend only classes that appear in an image for vis_label

yuyu2172:
There is not much point in showing legends for all the classes in a dataset.
Similar to `vis_bbox`, `vis_label` should limit information to the one that is contained in an image.

## Example
This dataset contains 80 classes.

#### Before
figure_1-2

#### After
figure_1

### DIFF
--- a/chainercv/visualizations/vis_label.py
+++ b/chainercv/visualizations/vis_label.py
@@ -67,8 +67,9 @@ def vis_label(
             overlaying the label on the source image.
         all_label_names_in_legend (bool): Determines whether to include
             all label names in a legend. If this is :obj:`False`,
-            only the names of labels
-            that are included in the label image are included.
+            the legend does not contain the names of unused labels.
+            An unused label is defined as a label that does not appear in
+            :obj:`label`.
             The default value is :obj:`False`.
         ax (matplotlib.axes.Axis): The visualization is displayed on this
             axis. If this is :obj:`None` (default), a new axis is created.
--- a/chainercv/visualizations/vis_label.py
+++ b/chainercv/visualizations/vis_label.py
@@ -116,6 +116,8 @@ def vis_label(
 
     legend_handles = list()
     for l in np.unique(label):
+        if l < 0:
+            continue
         legend_handles.append(
             Patch(color=cmap(l / (n_class - 1)), label=label_names[l]))
 
--- a/chainercv/visualizations/vis_label.py
+++ b/chainercv/visualizations/vis_label.py
@@ -115,8 +115,8 @@ def vis_label(
     ax.imshow(img)
 
     legend_handles = list()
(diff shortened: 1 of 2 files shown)


Hakuyume:
I prefer showing all legends because it looks better to me when I plot more than one images. Of course, this is up to the situation. How about adding an option `unused_regend` or `all_regend`? It is better to set the default behaviour to show only used legends.

yuyu2172:
> I prefer showing all legends because it looks better to me when I plot more than one images. Of course, this is up to the situation. How about adding an option unused_regend or all_regend? It is better to set the default behaviour to show only used legends.

Nice suggestion!

chainercv/visualizations/vis_label.py
```diff
@@ -115,8 +115,8 @@ def vis_label(
     ax.imshow(img)
 
     legend_handles = list()
-    for l, label_name in enumerate(label_names):
+    for l in np.unique(label):
```
Hakuyume: We should skip ignored label.
```
if l < 0:
    continue
```

chainercv/visualizations/vis_label.py
```diff
@@ -115,7 +121,11 @@ def vis_label(
     ax.imshow(img)
 
     legend_handles = list()
-    for l in np.unique(label):
+    if all_label_names_in_legend:
+        legend_labels = [l for l in np.unique(label) if l >= 0]
+    else:
+        legend_labels = list(range(len(label_names)))
```
Hakuyume: You don't have to convert to list. `range(...)` can work.

chainercv/visualizations/vis_label.py
```diff
@@ -64,6 +65,11 @@ def vis_label(
             value is :obj:`0`, the figure will be completely transparent.
             The default value is :obj:`1`. This option is useful for
             overlaying the label on the source image.
+        all_label_names_in_legend (bool): Determines whether to include
+            all label names in a legend. If this is :obj:`False`,
+            only the names of labels
+            that are included in the label image are included.
```
Hakuyume: In this sentence, ` are included` is used twice. It is hard to read. How about `If this is :obj:False, the legend does not contain the names of unused labels in legend. A unused label is defined as a label that does not appear in :obj:label.`

chainercv/visualizations/vis_label.py
```diff
@@ -115,8 +122,12 @@ def vis_label(
     ax.imshow(img)
 
     legend_handles = list()
-    for l, label_name in enumerate(label_names):
+    if all_label_names_in_legend:
+        legend_labels = [l for l in np.unique(label) if l >= 0]
+    else:
+        legend_labels = range(len(label_names))
```
Hakuyume: `len(label_names)` can be replaced with `n_class`.

PR #349, opened 2017-07-21
Fix Chainer version in travis

yuyu2172:
Currently in Travis, Chainer version 2.0.0 is installed.

You can check that here.
https://travis-ci.org/chainer/chainercv/jobs/255711216


yuyu2172:
Deprecated #351

PR #355, opened 2017-07-25
Remove matplotlib backend specification line from example code

mitmul:
`matplotlib.use('Agg')` sometimes causes problems in some environments. (ref. https://github.com/chainer/chainer/issues/2146)
Just using the environmental variable in this way:
`MPLBACKEND=Agg python train.py`
is preferred. This PR updates the faster-rcnn example in this way.

### DIFF
--- a/examples/faster_rcnn/README.md
+++ b/examples/faster_rcnn/README.md
@@ -51,7 +51,7 @@ You can train the model with the following code.
 $ python train.py [--gpu <gpu>]
 ```
 
-If you got `RuntimeError: Invalid DISPLAY variable` error on Linux environment, adding an environment variable specification is recommended:
+PlotReport extension uses matplotlib. If you got `RuntimeError: Invalid DISPLAY variable` error on Linux environment, adding an environment variable specification is recommended:
 
 ```
 $ MPLBACKEND=Agg python train.py [--gpu <gpu>]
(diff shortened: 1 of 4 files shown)


examples/faster_rcnn/README.md
```diff
@@ -51,6 +51,12 @@ You can train the model with the following code.
 $ python train.py [--gpu <gpu>]
 ```
 
+If you got `RuntimeError: Invalid DISPLAY variable` error on Linux environment, adding an environmental variable specification is recommended:
```
Hakuyume: I think `environment variable` is common (not `environmental variable`).

mitmul: Thanks! I fixed it.

Hakuyume:
The example code of SegNet also has this line.
https://github.com/chainer/chainercv/blob/master/examples/segnet/train.py#L3

mitmul:
@Hakuyume Thank you for pointing out it. I did the same fixation for the SegNet example.

Hakuyume:
In both examples, `matplotlib` is imported but not used. We can remove entire of `try: ...` block.

mitmul:
oh...

mitmul:
@Hakuyume Removed the part. I modifed README to explain `PlotReport` uses matplotlib, so it still suffer from RuntimeError.

Hakuyume:
One of Travis CI builds failed. I restarted it.

PR #362, opened 2017-07-29
[WIP] Add device option to evaluator extensions

mitmul:
I think specifying device option might be needed when users run the training with multiple GPUs.
This PR adds the option to two evaluator extensions.

- [ ] Write tests


Hakuyume:
This PR does not have any effects because those evaluators do not send arrays to GPU. They assume that `target.predict` sends arrays to GPU if needed.

mitmul:
Right. Sorry for my misunderstanding. I'm closing this.

PR #366, opened 2017-08-01
Add training result of SSD512

Hakuyume:
I added a training result of SSD512. Note that the batchsize is different from the original because of memory limitation.

### DIFF
--- a/examples/ssd/README.md
+++ b/examples/ssd/README.md
@@ -6,9 +6,11 @@ PASCAL VOC2007 Test
 | Model | Original | ChainerCV (weight conversion) | ChainerCV (train) |
 |:-:|:-:|:-:|:-:|
 | SSD300 | 77.5 % [2] | 77.8 % | 77.5 % |
-| SSD512 | 79.5 % [2] | 79.7 % | - |
+| SSD512 | 79.5 % [2] | 79.7 % | 80.1 % * |
 
 Scores are mean Average Precision (mAP) with PASCAL VOC2007 metric.
+
+\*: We set batchsize to 24 because of memory limitation. The original paper used 32.
 
 ## Demo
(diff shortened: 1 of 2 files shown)


yuyu2172:
I think it is better to upload a weight trained with ChainerCV and add a link to README. (same for SSD300)

By the way, I think that just adding a link on README is enough. I do not think that it is necessary to modify `pretrained_model` option to download trained weight.

Hakuyume:
> I think it is better to upload a weight trained with ChainerCV and add a link to README. (same for SSD300)

OK. Is the file name `ssd300_voc0712_trained_<year>_<month>_<data>.npz`?
I have more than one trained model. Is it better to upload all of them? (Is there anyone who want to check the reproducibility?)

> By the way, I think that just adding a link on README is enough. I do not think that it is necessary to modify pretrained_model option to download trained weight.

OK. I will leave the code as is.

yuyu2172:
I think the name is OK.

I think it is better to have all trials uploaded so that we can verify scores in the future.
Also, the weight that produced the median score should be emphasized.
It would be nice to clearly state 
+ its score
+ the fact that it is the median model.

How about putting a sentence like below in README.

```
The SSD300 weights that are trained using ChainerCV:
The weight with the median score among five trials: XXmAP   LINK
Rest of the weights:  [trial_1](LINK1) [trial_2](LINK2) [trial_3](LINK3)  [trial_4](LINK_4)

yuyu2172:
I updated README for SegNet and Faster R-CNN with scores calculated from multiple trials. #376

Hakuyume:
To reduce maintenance cost, I put only one model file for each architecture.

yuyu2172:
LGTM

PR #370, opened 2017-08-01
Support Chainer 2.0.1 and 3.0.0a1

kkk669:
`chainer.links.caffe.CaffeFunction.add_link` has been [deprecated](https://docs.chainer.org/en/latest/reference/generated/chainer.links.caffe.CaffeFunction.html#chainer.links.caffe.CaffeFunction.add_link) since 2.0.0, and `chainer.links.caffe.CaffeFunction` no longer calls `self.add_link` since [this commit](https://github.com/chainer/chainer/pull/2947/files) was merged.

I switched from overriding `add_link` to overriding `__setattr__` so that it can run in Chainer 2.0.1 or later.

This PR is almost the same as [this](https://github.com/Hakuyume/chainer-ssd/pull/10).

### DIFF
--- a/examples/ssd/caffe2npz.py
+++ b/examples/ssd/caffe2npz.py
@@ -47,17 +47,17 @@ class SSDCaffeFunction(caffe.CaffeFunction):
         print('loading weights from {:s} ... '.format(model_path))
         super(SSDCaffeFunction, self).__init__(model_path)
 
-    def __setattr__(self, name, link):
-        if self.within_init_scope and isinstance(link, Link):
+    def __setattr__(self, name, value):
+        if self.within_init_scope and isinstance(value, Link):
             new_name = rename(name)
 
             if new_name == 'extractor/conv1_1':
                 # BGR -> RGB
-                link.W.data[:, ::-1] = link.W.data
+                value.W.data[:, ::-1] = value.W.data
                 print('{:s} -> {:s} (BGR -> RGB)'.format(name, new_name))
             elif new_name.startswith('multibox/loc/'):
                 # xy -> yx
-                for data in (link.W.data, link.b.data):
+                for data in (value.W.data, value.b.data):
                     data = data.reshape((-1, 4) + data.shape[1:])
                     data[:, [1, 0, 3, 2]] = data.copy()
                 print('{:s} -> {:s} (xy -> yx)'.format(name, new_name))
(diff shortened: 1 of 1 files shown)


examples/ssd/caffe2npz.py
```diff
@@ -59,17 +61,18 @@ def add_link(self, name, link):
                 data = data.reshape((-1, 4) + data.shape[1:])
                 data[:, [1, 0, 3, 2]] = data.copy()
             print('{:s} -> {:s} (xy -> yx)'.format(name, new_name))
-        else:
+        elif match:
```
Hakuyume: Could you remove `match`? It is better to show all layers even if their names are not changed. And this modification is not related to supporting v2 and v3.

kkk669: Yes, but the reason why I added `match` is because it shows all attributes even if they are not layers (e.g. [self.verbose](https://github.com/Hakuyume/chainer-ssd/blob/master/lib/caffe.py#L45)).

Hakuyume: I see. How about checking `self.within_init_scope` (https://github.com/chainer/chainer/blob/master/chainer/link.py#L153)? This property becomes `True` if the attribute is a layer (a child link).

Hakuyume: Using `within_init_scope`, the code will be more simple.
```
def __setattr__(self, name, link):
    if self.within_init_scope:
        # link is a layer (a child link)
    else:
        # link is a regular attribute
```

kkk669: I didn't know it. Thank you. I'll remove match and use it.

Hakuyume:
@kkk669 Thank you for updating this PR. It looks good to me. Travis CI is not working well now (This is not only for this PR but for all branches). After it is fixed, I will merge  this PR.

kkk669:
@Hakuyume I understand. My final commit is `81d1147` and I'm not planning to push future commits. Thank you for all your advices.

examples/ssd/caffe2npz.py
```diff
@@ -47,7 +48,7 @@ def __init__(self, model_path):
         super(SSDCaffeFunction, self).__init__(model_path)
 
     def __setattr__(self, name, link):
-        if self.within_init_scope and name != '_within_init_scoep':
+        if self.within_init_scope and instanceof(link, Link):
```
Hakuyume: `instanceof` -> `isinstance`
https://github.com/chainer/chainer/blob/master/chainer/link.py#L655

Hakuyume: Could you rename `link` -> `value`? `Chain.__setattr__` uses this name. (https://github.com/chainer/chainer/blob/master/chainer/link.py#L655)

kkk669: I'm sorry for my typo.

kkk669: I've done.

Hakuyume: Thank you!

PR #371, opened 2017-08-03
Fix use_gpu in examples

Hakuyume:
`examples/segnet/eval_camvid.py` fails if `args.gpu > 0`. I fixed this bug by adding `chainer.cuda.get_device(args.gpu).use()`. Other modifications are only for consistency.


yuyu2172:
LGTM

PR #382, opened 2017-08-08
use write_image in tests

Hakuyume:


yuyu2172:
I am not sure if it is a good idea to introduce an extra dependency.

Hakuyume:
It is better to use well-tested function than to write slicing and transposing directly.

yuyu2172:
OK.
If so, it would be better to change tests for `read_image`.

Hakuyume:
> If so, it would be better to change tests for read_image.

OK, I will fix it.

yuyu2172:
LGTM

PR #383, opened 2017-08-08
Add FeaturePredictor

yuyu2172:
Split from #265.

`FeatureExtractionPredictor` will be added to the documentation in #265.

### DIFF
--- a/chainercv/links/model/feature_extraction_predictor.py
+++ b/chainercv/links/model/feature_extraction_predictor.py
@@ -143,18 +143,18 @@ class FeatureExtractionPredictor(chainer.Chain):
 
         with chainer.function.no_backprop_mode():
             imgs = chainer.Variable(imgs)
-            activations = self.extractor(imgs)
+            features = self.extractor(imgs)
 
-        if isinstance(activations, tuple):
+        if isinstance(features, tuple):
             output = []
-            for activation in activations:
+            for activation in features:
                 activation = activation.data
                 if n_crop > 1:
                     activation = self._average_crops(activation, n_crop)
                 output.append(cuda.to_cpu(activation))
             output = tuple(output)
         else:
-            output = cuda.to_cpu(activations.data)
+            output = cuda.to_cpu(features.data)
             if n_crop > 1:
                 output = self._average_crops(output, n_crop)
 
--- a/chainercv/links/model/feature_extraction_predictor.py
+++ b/chainercv/links/model/feature_extraction_predictor.py
@@ -49,7 +49,8 @@ class FeatureExtractionPredictor(chainer.Chain):
             If this is an integer, the image is cropped to
             :math:`(crop_size, crop_size)`.
         scale_size (int or tuple): If :obj:`scale_size` is :obj:`None`,
-            neighther scaling nor resizing is conducted during preprocessing.
+            neither scaling nor resizing is conducted during preprocessing.
+            This is the default behavior.
             If this is an integer, an image is resized so that the length of
             the shorter edge is equal to :obj:`scale_size`. If this is a tuple
             :obj:`(height, width)`, the image is resized to
@@ -92,7 +93,7 @@ class FeatureExtractionPredictor(chainer.Chain):
         Last, the image is mean subtracted by an array :obj:`mean`.
 
         Args:
-            img (~numpy.ndarray): An image. This is in CHW and RGB format.
+            img (~numpy.ndarray): An image. This is in CHW format.
                 The range of its value is :math:`[0, 255]`.
 
         Returns:
--- a/chainercv/links/model/feature_extraction_predictor.py
+++ b/chainercv/links/model/feature_extraction_predictor.py
@@ -25,7 +25,7 @@ class FeatureExtractionPredictor(chainer.Chain):
         >>> from chainercv.links import VGG16
         >>> from chainercv.links import FeatureExtractionPredictor
         >>> base_model = VGG16()
-        >>> model = FeatureExtractionPredictor(base_model)
+        >>> model = FeatureExtractionPredictor(base_model, 224, 256)
         >>> prob = model.predict([img])
         # Predicting multiple features
         >>> model.extractor.feature_names = ['conv5_3', 'fc7']
@@ -48,31 +48,38 @@ class FeatureExtractionPredictor(chainer.Chain):
             cropping in preprocessing.
             If this is an integer, the image is cropped to
             :math:`(crop_size, crop_size)`.
-        scale_size (int or tuple): If :obj:`scale_size` is an integer, during
-            preprocessing, an image is resized so that the length of the
-            shorter edge is equal to :obj:`scale_size`. If this is a tuple
+        scale_size (int or tuple): If :obj:`scale_size` is :obj:`None`,
+            neighther scaling nor resizing is conducted during preprocessing.
+            If this is an integer, an image is resized so that the length of
+            the shorter edge is equal to :obj:`scale_size`. If this is a tuple
             :obj:`(height, width)`, the image is resized to
             :math:`(height, width)`.
         crop ({'center', '10'}): Determines the style of cropping.
+        mean (numpy.ndarray): A mean value. If this is :obj:`None`,
+            :obj:`extractor.mean` is used as the mean value.
 
     """
 
     def __init__(self, extractor,
-                 crop_size=224, scale_size=256,
-                 crop='center'):
+                 crop_size, scale_size=None,
+                 crop='center', mean=None):
         super(FeatureExtractionPredictor, self).__init__()
         self.scale_size = scale_size
         if isinstance(crop_size, int):
             crop_size = (crop_size, crop_size)
         self.crop_size = crop_size
         self.crop = crop
+        self._mean = mean
 
         with self.init_scope():
             self.extractor = extractor
 
     @property
     def mean(self):
-        return self.extractor.mean
+        if self._mean is None:
+            return self.extractor.mean
+        else:
+            return self._mean
 
     def _prepare(self, img):
         """Prepare an image for feeding it to a model.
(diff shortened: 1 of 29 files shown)


chainercv/links/model/feature_extraction_predictor.py
```diff
@@ -0,0 +1,161 @@
...
+            tuple of variables.
+        crop_size (int or tuple): The height and the width of an image after
+            cropping in preprocessing.
+            If this is an integer, the image is cropped to
+            :math:`(crop_size, crop_size)`.
+        scale_size (int or tuple): If :obj:`scale_size` is an integer, during
+            preprocessing, an image is resized so that the length of the
+            shorter edge is equal to :obj:`scale_size`. If this is a tuple
+            :obj:`(height, width)`, the image is resized to
+            :math:`(height, width)`.
```
Hakuyume: How about supporting `do not scale`? Some user may want this feature if all images are larger than `crop_size`.

chainercv/links/model/feature_extraction_predictor.py
```diff
@@ -0,0 +1,161 @@
...
+        crop ({'center', '10'}): Determines the style of cropping.
+
+    """
+
+    def __init__(self, extractor,
+                 crop_size=224, scale_size=256,
```
Hakuyume: Are these default sizes reasonable?

chainercv/links/model/feature_extraction_predictor.py
```diff
@@ -0,0 +1,161 @@
...
+    patches that are ten-cropped from the input images.
+
+    When extracting more than one crops from an image, the output of
+    :meth:`predict` returns the average of the features computed from the
+    crops.
+
+    Args:
+        extractor: A feature extraction model. This is a callable chain
+            that takes a batch of images and returns a variable or a
+            tuple of variables.
```
Hakuyume: This code assumes that `extractor` has an attribute `mean`. This assumption should be noted. For example, a vanilla `SequentialFeatureExtractor` does not satisfy this.

chainercv/links/model/feature_extraction_predictor.py
```diff
@@ -0,0 +1,161 @@
...
+        """Prepare an image for feeding it to a model.
+
+        This is a standard preprocessing scheme used by feature extraction
+        models.
+        First, the image is scaled or resized according to :math:`scale_size`.
+        Next, the image is cropped to :math:`crop_size`.
+        Last, the image is mean subtracted by an array :obj:`mean`.
+
+        Args:
+            img (~numpy.ndarray): An image. This is in CHW and RGB format.
```
Hakuyume: Can't we use a grayscale image as input?

chainercv/links/model/feature_extraction_predictor.py
```diff
@@ -0,0 +1,161 @@
...
+
+        """
+        imgs = self.xp.asarray([self._prepare(img) for img in imgs])
+        n_crop = imgs.shape[-4]
+        shape = (-1, imgs.shape[-3]) + self.crop_size
+        imgs = imgs.reshape(shape)
+
+        with chainer.function.no_backprop_mode():
+            imgs = chainer.Variable(imgs)
+            activations = self.extractor(imgs)
```
Hakuyume: I prefer `features` to `activations`.

chainercv/links/model/feature_extraction_predictor.py
yuyu2172: Maybe no.
For crop size, 224 is common, but not all networks use this.
For example, ResNet and DenseNet families use 224, but InceptionV3 use 299.

For scale size, I think 256 is arbitrary.
Perhaps, it should not conduct scaling by default.

chainercv/links/model/feature_extraction_predictor.py
yuyu2172: How about adding an argument `mean`?
If `mean` is None, `FeatureExtractorPredictor` tries to use extractor's attribute.
If `mean` is not None, it uses the argument.

chainercv/links/model/feature_extraction_predictor.py
yuyu2172: I see.
How about not scaling an image when `scale_size=None`?

chainercv/links/model/feature_extraction_predictor.py
yuyu2172: Yes we can. I will modify it.

chainercv/links/model/feature_extraction_predictor.py
yuyu2172: I agree.

chainercv/links/model/feature_extraction_predictor.py
yuyu2172: For the first four arguments, how about making them like this.
`__init__(self, extractor, crop_size, scale_size=None, ...)`.

chainercv/links/model/feature_extraction_predictor.py
Hakuyume: Nice idea.

chainercv/links/model/feature_extraction_predictor.py
Hakuyume: Good. And I think it should be the default behaviour.

chainercv/links/model/feature_extraction_predictor.py
Hakuyume: How about setting no default value for `crop_size`? It is hard to set a good value. (FYI. AlexNet uses 227)

> Perhaps, it should not conduct scaling by default.

I think this is a good idea.

yuyu2172:
Updated

yuyu2172:
Perhaps, this can have more generic name like Predictir similar to Classifier.
This can be used with SemanticSegmentation model as well.

Hakuyume:
I think `Predictor` is too general. This class doesn't fit to object detection models.

yuyu2172:
The preprocessing scheme is specific to classification (crop).
I think it is useful to have a predictor with this kind of functionality.
Since it is not a good idea to make it too generic, the current name and design seems to be good.

Hakuyume:
I agree with you that current design is good. To shorten the name, how about `FeaturePredictor`?

yuyu2172:
Forget what I said first...
Semantic segmentation and classification are very different when It comes to preprocessing.

yuyu2172:
I think that name is better.

yuyu2172:
Updated.

yuyu2172:
@Hakuyume

chainercv/links/model/feature_predictor.py
```diff
@@ -0,0 +1,171 @@
...
+
+        with self.init_scope():
+            self.extractor = extractor
+
+    @property
+    def mean(self):
+        if self._mean is None:
+            return self.extractor.mean
+        else:
+            return self._mean
```
Hakuyume: How about checking existence of `mean` in `__init__`?
```python
    def __init__(...):
        ...
        if mean is None:
            self.mean = self.extractor.mean
        else:
            self.mean = mean
```
I prefer this version because it can notify the absence of `mean` before `predict`.

In the case of current version:
```python
extractor = ExtractorWithoutMean()
predictor = FeaturePredictor(extractor, crop_size=224)  # mean should be specified but no errors are shown here
...
predictor.predict(imgs)  # An exception will be raised
```

chainercv/links/model/feature_predictor.py
```diff
@@ -0,0 +1,171 @@
...
+    def _average_crops(self, y, n_crop):
+        if y.ndim == 4:
+            warnings.warn(
+                'Four dimensional features are averaged. '
+                'If these are batch of 2D spatial features, '
+                'their spatial information would be lost.')
+
+        xp = chainer.cuda.get_array_module(y)
+        n = y.shape[0] // n_crop
+        y = y.reshape((n, n_crop) + y.shape[1:])
```
Hakuyume: How about `y = y.reshape((-1, n_crop) + y.shape[1:])`?

chainercv/links/model/feature_predictor.py
```diff
@@ -0,0 +1,171 @@
...
+        y = xp.sum(y, axis=1) / n_crop
```
Hakuyume: How about using `xp.mean`?

chainercv/links/model/feature_predictor.py
```diff
@@ -0,0 +1,171 @@
...
+
+        Returns:
+            numpy.ndarray or tuple of numpy.ndarray:
+            A batch of features or a tuple of them.
+
+        """
+        imgs = self.xp.asarray([self._prepare(img) for img in imgs])
+        n_crop = imgs.shape[-4]
+        shape = (-1, imgs.shape[-3]) + self.crop_size
+        imgs = imgs.reshape(shape)
```
Hakuyume: How about adding comments about shape?
```python
        # [(C, H_0, W_0), (C, H_1, W_1), ... (C, H_{B-1}, W_{B-1})] -> (B, N, C, H, W)
        imgs = self.xp.asarray([self._prepare(img) for img in imgs])
        n_crop = imgs.shape[-4]
        shape = (-1, imgs.shape[-3]) + self.crop_size
        # (B, N, C, H, W) -> (B * N, C, H, W)
        imgs = imgs.reshape(shape)
```

chainercv/links/model/feature_predictor.py
```diff
@@ -0,0 +1,171 @@
...
+
+        with chainer.function.no_backprop_mode():
+            imgs = chainer.Variable(imgs)
+            features = self.extractor(imgs)
+
+        if isinstance(features, tuple):
+            output = []
+            for activation in features:
```
Hakuyume: `activation` -> `feature` https://github.com/chainer/chainercv/pull/383#discussion_r131885521

Hakuyume:
@yuyu2172 I'm sorry for letting you wait. I added some comments.
Could you merge master branch to this PR to enable style checking we added?

chainercv/links/model/feature_predictor.py
yuyu2172: Nice suggestion!

yuyu2172:
Thanks for review. I merged the master branch.

PR #384, opened 2017-08-08
Add Conv2DActiv

yuyu2172:
Split from #265.

### DIFF
--- a/chainercv/links/__init__.py
+++ b/chainercv/links/__init__.py
@@ -1,4 +1,4 @@
-from chainercv.links.connection.conv_2d_activ import Conv2DActiv
+from chainercv.links.connection.conv_2d_activ import Conv2DActiv  # NOQA
 
 from chainercv.links.model.pixelwise_softmax_classifier import PixelwiseSoftmaxClassifier  # NOQA
 from chainercv.links.model.sequential_feature_extractor import SequentialFeatureExtractor  # NOQA
--- a/chainercv/links/__init__.py
+++ b/chainercv/links/__init__.py
@@ -1,3 +1,5 @@
+from chainercv.links.connection.conv_2d_activ import Conv2DActiv
+
 from chainercv.links.model.pixelwise_softmax_classifier import PixelwiseSoftmaxClassifier  # NOQA
 from chainercv.links.model.sequential_feature_extractor import SequentialFeatureExtractor  # NOQA
 
--- a/chainercv/links/connection/conv_2d_activ.py
+++ b/chainercv/links/connection/conv_2d_activ.py
@@ -1,0 +1,73 @@
+import chainer
+from chainer.functions import relu
+from chainer.links import Convolution2D
+
+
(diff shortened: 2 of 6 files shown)


chainercv/links/connection/conv_2d_activ.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Example:
+
+        There are sevaral ways to initialize a :class:`Conv2DActiv`.
+
+        1. Give the first three arguments explicitly:
+
+            >>> l = Conv2DActiv(5, 10, 3)
+
+        2. Omit :obj:`in_channels` or fill it with :obj:`None`:
```
Hakuyume: The order of descriptions should be same as that of examples.
I mean, `Fill :obj:in_channels with :obj:None: or omit it:` is better. (Changing the order of examples is also OK)

chainercv/links/__init__.py
```diff
@@ -1,3 +1,5 @@
+from chainercv.links.connection.conv_2d_activ import Conv2DActiv
```
Hakuyume: How about `block` instead of `connection`? `connection` sounds too ambiguous.

tests/links_tests/connection_tests/test_conv_2d_activ.py
```diff
@@ -0,0 +1,120 @@
...
+def _add_one(x):
+    return x + 1
+
+
+@testing.parameterize(
+    {'args_style': 'explicit'},
+    {'args_style': 'None'},
+    {'args_style': 'omit'}
+)
+class TestConv2DActivForward(unittest.TestCase):
```
Hakuyume: Can you merge `TestConv2DActivForward` and `TestConv2DActivBarckward` into one class?

tests/links_tests/connection_tests/test_conv_2d_activ.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _add_one(x):
+    return x + 1
+
+
+@testing.parameterize(
+    {'args_style': 'explicit'},
+    {'args_style': 'None'},
+    {'args_style': 'omit'}
```
Hakuyume: How about this ?
```
 'args_style': ['explicit', 'None', 'omit'],
 'activ: [None, _add_one]
```

chainercv/links/__init__.py
yuyu2172: I made it to be consistent with the Chainer's directory structure.
`connection` seems to include links with learnable parameters (e.g. Convolution2D) and a block of them (e.g. Inception).
If we change the directory structure, we need to come up with a new name for group of links that are not blocks, but categorized under `connection`.

Hakuyume: > I made it to be consistent with the Chainer's directory structure.

I see. `connection` is OK.

yuyu2172:
I reflected your comments.

PR #388, opened 2017-08-10
Add PSPNet model

mitmul:
Please merge #494 and #392 first.
This PR uses `Conv2DBNActiv` with a ChainerMN communicator and `CityscapesSemanticSegmentationDataset`.

This PR adds PSPNet model definition and it's inference demo code.
This also includes convert.py to convert the official caffemodel to Chainer model.

- [x] Ensure the convert.py correctly works
- [x] Add tests
- [x] Add evaluation scripts to ensure the conversion correctness


chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+            chainer.config.initialW = chainer.initializers.HeNormal()
+        else:
+            chainer.config.initialW = initialW
+
+        if not isinstance(input_size, (list, tuple)):
+            input_size = (int(input_size), int(input_size))
+
+        with self.init_scope():
+            self.input_size = input_size
+            self.trunk = DilatedFCN(n_blocks=n_blocks)
```
yuyu2172: Please use `extractor` instead of `trunk`.
This is consistent with links such as `FasterRCNN`.

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+        return F.concat(ys, axis=1)
+
+
+class BottleneckConv(chainer.Chain):
+
+    def __init__(self, in_ch, mid_ch, out_ch, stride=2, dilate=False):
+        mid_stride = chainer.config.mid_stride
+        super(BottleneckConv, self).__init__()
+        with self.init_scope():
+            self.cbr1 = ConvBNReLU(
```
yuyu2172: I want to make the attribute name of `ConvBN*` consistent across the library.
We have `Conv2DBNActiv`, which will be used by ResNet implementations.

Since `cbr` is specific to `Conv->BN->ReLU`, I want to avoid the name.
How about `conv1`, `conv2`, ... ?

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+            self.cbr3 = ConvBNReLU(mid_ch, in_ch, 1, 1, 0)
+
+    def __call__(self, x):
+        h = self.cbr1(x)
+        h = self.cbr2(h)
+        h = self.cbr3(h, relu=False)
+        return F.relu(h + x)
+
+
+class ResBlock(chainer.ChainList):
```
yuyu2172: Can you merge `ResBlock` and `DilatedResBlock`?

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+            self.res3 = ResBlock(n_blocks[1], 256, 128, 512, 2)
+            self.res4 = DilatedResBlock(n_blocks[2], 512, 256, 1024, 2)
+            self.res5 = DilatedResBlock(n_blocks[3], 1024, 512, 2048, 4)
+
+    def __call__(self, x):
+        h = self.cbr1_3(self.cbr1_2(self.cbr1_1(x)))  # 1/2
+        h = F.max_pooling_2d(h, 3, 2, 1)  # 1/4
+        h = self.res2(h)
+        h = self.res3(h)  # 1/8
+        if chainer.config.train:
```
yuyu2172: Why not always return `h1` and `h2`?

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+                mean = self._models[pretrained_model]['mean']
+                self._use_pretrained_model = True
+
+        chainer.config.mid_stride = mid_stride
+        chainer.config.comm = comm
+
+        if initialW is None:
+            chainer.config.initialW = chainer.initializers.HeNormal()
+        else:
+            chainer.config.initialW = initialW
```
yuyu2172: I prefer not using global variables.

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+        if not isinstance(input_size, (list, tuple)):
+            input_size = (int(input_size), int(input_size))
+
+        with self.init_scope():
+            self.input_size = input_size
+            self.trunk = DilatedFCN(n_blocks=n_blocks)
+
+            # To calculate auxirally loss
+            if chainer.config.train:
+                self.cbr_aux = ConvBNReLU(None, 512, 3, 1, 1)
```
yuyu2172: I do not like to change the configuration of a model using global variable.
It would not be much problem to instantiate these links. Why not always instantiate them?

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+            Training time: it returns the outputs from auxiliary branch and the
+                main branch. So the returned value is a tuple of two Variables.
+            Inference time: it returns the output of the main branch. So the
+                returned value is a sinle Variable which forms
+                ``(N, n_class, H, W)`` where ``N`` is the batchsize and
+                ``n_class`` is the number of classes specified in the
+                constructor. ``H, W`` is the input image size.
+
+        """
+        if chainer.config.train:
```
yuyu2172: Why not add `compute_aux` option instead of changing behavior using `chainer.config.train`?
There can be a situation when a user wants to compute `aux` with `chainer.config.train=False`.

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+        Returns:
+            ~numpy.ndarray:
+            A preprocessed image.
+
+        """
+        if self.mean is not None:
+            img -= self.mean[:, None, None]
+            img = img.astype(np.float32, copy=False)
+            if self._use_pretrained_model:
+                # Pre-trained model is trained for BGR images
```
yuyu2172: Can you flip the order of in_channels of the weights in conversion code?
By doing so, this if statement can be deleted.
I want to make `pspnet.py` to be as simple as possible at the expense of more complex conversion code because people would rarely see the conversion code.

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+            'feat_size': 60,
+            'mid_stride': True,
+            'pyramids': [6, 3, 2, 1],
+            'mean': np.array([123.68, 116.779, 103.939]),
+            'url': 'https://github.com/mitmul/chainer-pspnet/releases/download'
+                   '/ChainerCV_PSPNet/pspnet50_ADE20K_473_reference.npz'
+        }
+    }
+
+    def __init__(self, n_class=None, input_size=None, n_blocks=None,
```
yuyu2172: How about making `PSPNet50` and `PSPNet101`, which inherit from `PSPNet`.
In my idea, `PSPNet.__init__` would take `extractor` as input.

I have the designs of `FasterRCNN` and `FasterRCNNVGG16` in my mind.

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+    warnings.warn('To perform batch normalization with multiple GPUs or '
+                  'multiple nodes, MultiNodeBatchNormalization link is '
+                  'needed. Please install ChainerMN: '
+                  'pip install pip install git+git://github.com/chainer/'
+                  'chainermn.git@distributed-batch-normalization')
+
+
+class ConvBNReLU(chainer.Chain):
+
+    def __init__(self, in_ch, out_ch, ksize, stride=1, pad=1, dilation=1):
```
yuyu2172: Can you make input variables/variable names similar to `Conv2DBNActiv`?

chainercv/links/model/pspnet/pspnet.py
```diff
@@ -0,0 +1,476 @@
...
+import chainer.functions as F
+import chainer.links as L
+from chainercv.utils import download_model
+import numpy as np
+import six
+
+try:
+    from chainermn.links import MultiNodeBatchNormalization
+except Exception:
+    warnings.warn('To perform batch normalization with multiple GPUs or '
```
yuyu2172: Instead of raising a warning during import, it should be raised when PSPNet is instantiated.

chainercv/links/model/pspnet/pspnet.py
mitmul: `h1` is only for calculating auxiliary loss. That is completely useless during inference.

chainercv/links/model/pspnet/pspnet.py
mitmul: Right. I'll change it.

chainercv/links/model/pspnet/pspnet.py
mitmul: OK, I'll try to do it.

chainercv/links/model/pspnet/pspnet.py
yuyu2172: `predict` should be called during inference and not `__call__`. There is not much inconvenience returning it.

yuyu2172: In general, I want to avoid using a global variable to configure behavior if possible.

yuyu2172:
I ran `predict` and an error was raised.
I used Python3.

```
  File "/home/yuyu2172/projects/chainercv/chainercv/links/model/pspnet/pspnet.py", line 420, in _tile_predict           
    for yy in six.moves.xrange(hh):                                                                                     
TypeError: integer argument expected, got float                                                                         
```

yuyu2172:
Can you make sure that `predict` does not alter the value of input images?
Currently, `prepare` changes the value because the input is not copied.

chainercv/links/model/pspnet/pspnet.py
mitmul: Thanks, I changed the attribute name to `conv1`, `conv2`, ...

chainercv/links/model/pspnet/pspnet.py
mitmul: Merged.

chainercv/links/model/pspnet/pspnet.py
mitmul: OK, I changed this part to return both `h1` and `h2` always.

chainercv/links/model/pspnet/pspnet.py
mitmul: OK, let's instantiate them always.

chainercv/links/model/pspnet/pspnet.py
mitmul: OK, I changed it.

chainercv/links/model/pspnet/pspnet.py
mitmul: I replaced all such part to use a local variable given by the argument to the class.

chainercv/links/model/pspnet/pspnet.py
yuyu2172: Please fix this. Please make sure that the error is raised only when `comm` is specified.

Also, we should install the latest release.

yuyu2172:
@Hakuyume  @mitmul 

What do you think about reusing ResBlock for `resnet` for PSPNet?
This can be possible once we merge #494.

https://github.com/chainer/chainercv/blob/master/chainercv/links/model/resnet/resblock.py

Hakuyume:
:+1: for reusing `ResBlock` in PSPNet if possible.

mitmul:
@yuyu2172 @Hakuyume Can I restart working on this? What's the status around it. Just updating this PR by following your reviews is fine?

yuyu2172:
I was actually working on it, but had not yet completed implementations.
I can finish the inference-part by this week. Is that OK?

https://github.com/yuyu2172/chainercv/tree/add-pspnet-infer-major-change

yuyu2172:
Reopening PR #610

mitmul:
@yuyu2172 Thanks, that's fine. I think it's better to close this PR. Thank you for handing over this.

PR #389, opened 2017-08-14
[links/models/faster_rcnn/rpn] format to give tuple for transpose() and resize()

knorth55:


yuyu2172:
Why do you prefer this style?

yuyu2172:
Oh. The style is inconsistent...

knorth55:
This style is common, I think, and other lines in this file follows this style.

yuyu2172:
It is better to consider other code in ChainerCV together.
I greped, and `non tuple` style is used everywhere else.

Can you change all the `reshape` in `region_proposal_network` to use `non tuple` style?
This style is preferred because less characters are used.

knorth55:
Im not a specialist of numpy, but which style is preferred?
In official doc, they ise tuple style so i thought its style is common.

yuyu2172:
I checked Chainer, and both styles are used.

As I said, I prefer shorter syntax since it is easier to read.

yuyu2172:
I gave more thoughts on this issue, and I changed my mind to support `tuple` style.
I was not really aware of the official stance to this issue.

I googled a bit more about the recommended style, and I found that `tuple` style is the preferred style. (a bit more strongly preferred than I had expected)
First, from an issue on NumPy (https://github.com/numpy/numpy/issues/570), I found a comment by one of the core devs saying that `tuple` style is better.
Also, I read the doc, and there seems to be *no* documentation on `non-tuple` style.

Although, `tuple` stlye is the recommended style, aesthetically it looked worse than `non-tuple` style to me.
It turned out that `tuple` style is not always strictly enforced.
In a case when shape is one dimensional, the official doc permits the usage of integer instead of tuple (e.g. `a.reshape(6)` is OK instead of `a.reshape((6,))`).
This situation was my main concern, so this finding mitigated the problem I was worried about a lot.

knorth55:
so, what you prefer is 
- one dimension: use `non-tuple` style
- more than two dimensions: use `tuple` style
is it correct?

yuyu2172:
Yes.

Hakuyume:
I wrote a simple script to check the style.
Here is the result.
https://gist.github.com/Hakuyume/b8f264346b130e9c9cb25af6e1aa1a35

yuyu2172:
Great. We should call the script inside Travis.

@knorth55 
Is it OK to close this PR, and reopen a PR that fixes the style and modify Travis?

Hakuyume:
Shall I make a PR to add style checking to Travis?

yuyu2172:
Yes, please.

knorth55:
You can merge this first, and refine other codes later.

Hakuyume:
I made a PR for checker #393 and Travis CI could find style problems (https://travis-ci.org/chainer/chainercv/builds/264705958).

knorth55:
move to #394

PR #379, opened 2017-08-07
Add Conv2DBNActiv

yuyu2172:
Merge after #265.

PR #390, opened 2017-08-14
Add Conv2DBNActiv

yuyu2172:
Reopend https://github.com/chainer/chainercv/pull/379

### DIFF
--- a/chainercv/links/connection/conv_2d_bn_activ.py
+++ b/chainercv/links/connection/conv_2d_bn_activ.py
@@ -7,7 +7,7 @@ from chainer.links import Convolution2D
 class Conv2DBNActiv(chainer.Chain):
     """Convolution2D --> Batch Normalization --> Activation
 
-    This is a chain that sequentially apllies two-dimensional convolution,
+    This is a chain that sequentially apllies a two-dimensional convolution,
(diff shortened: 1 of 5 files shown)


chainercv/links/connection/conv_2d_bn_activ.py
```diff
@@ -0,0 +1,78 @@
...
+            initializer is used.
+            May also be a callable that takes :obj:`numpy.ndarray` or
+            :obj:`cupy.ndarray` and edits its value.
+        initial_bias (1-D array): Initial bias value. If :obj:`None`, the bias
+            is set to 0.
+            May also be a callable that takes :obj:`numpy.ndarray` or
+            :obj:`cupy.ndarray` and edits its value.
+        activ (callable): An activation function. The default value is
+            :func:`chainer.functions.relu`.
+        bn_kwargs (dict): Key-word arguments passed to initialize
```
Hakuyume: `Key-word` -> `Keyword`?

Hakuyume:
@yuyu2172 Perhaps, it is better to set `nobias=True` by default because `nobias=True` is common when batchnorm is used.

yuyu2172:
I agree. That sounds better.

PR #230, opened 2017-05-31
[WIP] Add Cityscapes segmentation dataset

t-abe:
#138 

This is a dataset class for segmentation task of Cityscapes dataset.
Basically I implemented it by following ``datasets.camvid``'s manner.

- [x] python2
- [ ] Style
- [ ] Tests
- [ ] Docstring

PR #161, opened 2017-05-10
Add SegNetBasic model

mitmul:
Merge after #132 

- SegNetBasic model is from this Caffe impl.: https://github.com/alexgkendall/SegNet-Tutorial/blob/master/Models/segnet_basic_train.prototxt

### Training

```
cd examples/segnet
CHAINER_TYPE_CHECK=0 python -OO -W ignore train.py --gpu 0
```
`eval_semantic_segmentation` makes a lot of warnings, so I suppressed them by giving `-W ignore` option.

### Memo

- According to the original implementation, the authors seemed to perform LR flipping to the input images for data augmentation: https://github.com/alexgkendall/caffe-segnet/blob/segnet-cleaned/src/caffe/layers/dense_image_data_layer.cpp#L168-L175
- Chainer's LRN layer is different from Caffe's one in terms of the meaning of "alpha" argment: https://github.com/alexgkendall/caffe-segnet/blob/segnet-cleaned/src/caffe/layers/lrn_layer.cpp#L121

## Results

The evaluation result after 16000 iterations:
```
                    Sky : 0.8675
               Building : 0.6487
                   Pole : 0.1795
           Road_marking : 0.8493
                   Road : 0.6230
               Pavement : 0.6051
                   Tree : 0.1813
             SignSymbol : 0.1901
              Fence,Car : 0.5924
             Pedestrian : 0.2402
              Bicyclist : 0.2111
==================================
               mean IoU : 0.4716
 Class average accuracy : 0.6705
Global average accuracy : 0.8266
```

```
Global accuracy : 82.8%
Class accuracy : 62.3%
Mean IoU : 46.3%
```

The trained weights to replicate the same results as below is here: [model_iteration-16000](https://www.dropbox.com/s/exas66necaqbxyw/model_iteration-16000?dl=1)

PR #392, opened 2017-08-15
Add Cityscapes semantic segmentation dataset

mitmul:
This PR adds a dataset class for Cityscapes dataset. It replaces this PR: #230

### DIFF
--- a/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
+++ b/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
@@ -10,7 +10,9 @@ from chainercv.utils import read_image
 
 class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
 
-    """Dataset class for a semantic segmentation task on Cityscapes dataset
+    """Dataset class for a semantic segmentation task on `Cityscapes dataset`_.
+
+    .. _`Cityscapes dataset`: https://www.cityscapes-dataset.com
 
     .. note::
 
--- a/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
+++ b/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
@@ -25,7 +25,7 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
             contain at least two directories, :obj:`leftImg8bit` and either
             :obj:`gtFine` or :obj:`gtCoarse`. If :obj:`None` is given, it uses
             :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes` by default.
-        label_mode ({'fine', 'coarse'}): The resolution of the labels. It
+        label_resol ({'fine', 'coarse'}): The resolution of the labels. It
             should be either :obj:`fine` or :obj:`coarse`.
         split ({'train', 'val'}): Select from dataset splits used in
             Cityscapes dataset.
@@ -37,17 +37,17 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
 
     """
 
-    def __init__(self, data_dir=None, label_mode=None, split='train',
+    def __init__(self, data_dir=None, label_resol=None, split='train',
                  ignore_labels=True):
         if data_dir is None:
             data_dir = download.get_dataset_directory(
                 'pfnet/chainercv/cityscapes')
-        if label_mode not in ['fine', 'coarse']:
-            raise ValueError('\'label_mode\' argment should be eighter '
+        if label_resol not in ['fine', 'coarse']:
+            raise ValueError('\'label_resol\' argment should be eighter '
                              '\'fine\' or \'coarse\'.')
 
         img_dir = os.path.join(data_dir, os.path.join('leftImg8bit', split))
-        resol = 'gtFine' if label_mode == 'fine' else 'gtCoarse'
+        resol = 'gtFine' if label_resol == 'fine' else 'gtCoarse'
         label_dir = os.path.join(data_dir, resol)
         if not os.path.exists(img_dir) or not os.path.exists(label_dir):
             raise ValueError(
@@ -97,12 +97,11 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
         img = read_image(self.img_paths[i])
         label_orig = read_image(
             self.label_paths[i], dtype=np.int32, color=False)[0]
-        H, W = label_orig.shape
         if self.ignore_labels:
-            label_out = np.ones((H, W), dtype=np.int32) * -1
+            label_out = np.ones(label_orig.shape, dtype=np.int32) * -1
             for label in cityscapes_labels:
                 if not label.ignoreInEval:
-                    label_out[np.where(label_orig == label.id)] = label.trainId
+                    label_out[label_orig == label.id] = label.trainId
         else:
             label_out = label_orig
         return img, label_out
--- a/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
+++ b/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
@@ -25,15 +25,15 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
             contain at least two directories, :obj:`leftImg8bit` and either
             :obj:`gtFine` or :obj:`gtCoarse`. If :obj:`None` is given, it uses
             :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes` by default.
-        label_mode (string): The resolution of the labels. It should be either
-            :obj:`fine` or :obj:`coarse`.
+        label_mode ({'fine', 'coarse'}): The resolution of the labels. It
+            should be either :obj:`fine` or :obj:`coarse`.
         split ({'train', 'val'}): Select from dataset splits used in
             Cityscapes dataset.
         ignore_labels (bool): If True, the labels marked :obj:`ignoreInEval`
             defined in the original
             `cityscapesScripts<https://github.com/mcordts/cityscapesScripts>_`
             will be replaced with :obj:`-1` in the :meth:`get_example` method.
-            The default value is :obj:`True`
+            The default value is :obj:`True`.
 
     """
 
@@ -43,12 +43,19 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
             data_dir = download.get_dataset_directory(
                 'pfnet/chainercv/cityscapes')
         if label_mode not in ['fine', 'coarse']:
-            raise ValueError('\'label_name\' argment should be eighter '
+            raise ValueError('\'label_mode\' argment should be eighter '
                              '\'fine\' or \'coarse\'.')
 
         img_dir = os.path.join(data_dir, os.path.join('leftImg8bit', split))
         resol = 'gtFine' if label_mode == 'fine' else 'gtCoarse'
         label_dir = os.path.join(data_dir, resol)
+        if not os.path.exists(img_dir) or not os.path.exists(label_dir):
+            raise ValueError(
+                'Cityscapes dataset does not exist at the expected location.'
+                'Please download it from https://www.cityscapes-dataset.com/.'
+                'Then place directory leftImg8bit at {} and {} at {}.'.format(
+                    os.path.join(data_dir, 'leftImg8bit'), resol, label_dir))
+
         self.ignore_labels = ignore_labels
 
         self.label_paths = list()
--- a/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
+++ b/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
@@ -55,25 +55,25 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
         label_dir = os.path.join(data_dir, resol)
         self.ignore_labels = ignore_labels
 
-        self.label_fnames = list()
-        self.img_fnames = list()
+        self.label_paths = list()
+        self.img_paths = list()
         city_dnames = list()
         for dname in glob.glob(os.path.join(label_dir, '*')):
             if split in dname:
                 for city_dname in glob.glob(os.path.join(dname, '*')):
                     for label_fname in glob.glob(
                             os.path.join(city_dname, '*_labelIds.png')):
-                        self.label_fnames.append(label_fname)
+                        self.label_paths.append(label_fname)
                         city_dnames.append(os.path.basename(city_dname))
-        for city_dname, label_fname in zip(city_dnames, self.label_fnames):
+        for city_dname, label_fname in zip(city_dnames, self.label_paths):
             label_fname = os.path.basename(label_fname)
             img_fname = label_fname.replace(
                 '{}_labelIds'.format(resol), 'leftImg8bit')
             img_fname = os.path.join(img_dir, city_dname, img_fname)
-            self.img_fnames.append(img_fname)
+            self.img_paths.append(img_fname)
 
     def __len__(self):
-        return len(self.img_fnames)
+        return len(self.img_paths)
 
     def get_example(self, i):
         """Returns the i-th example.
@@ -91,9 +91,9 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
             the dtype of the label image is :obj:`numpy.int32`.
 
         """
-        img = read_image(self.img_fnames[i])
+        img = read_image(self.img_paths[i])
         label_orig = read_image(
-            self.label_fnames[i], dtype=np.int32, color=False)[0]
+            self.label_paths[i], dtype=np.int32, color=False)[0]
         H, W = label_orig.shape
         if self.ignore_labels:
             label_out = np.ones((H, W), dtype=np.int32) * -1
--- a/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
+++ b/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
@@ -25,7 +25,7 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
             contain at least two directories, :obj:`leftImg8bit` and either
             :obj:`gtFine` or :obj:`gtCoarse`. If :obj:`None` is given, it uses
             :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes` by default.
-        label_resol ({'fine', 'coarse'}): The resolution of the labels. It
+        label_resolutionution ({'fine', 'coarse'}): The resolution of the labels. It
             should be either :obj:`fine` or :obj:`coarse`.
         split ({'train', 'val'}): Select from dataset splits used in
             Cityscapes dataset.
@@ -37,17 +37,17 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
 
     """
 
-    def __init__(self, data_dir=None, label_resol=None, split='train',
+    def __init__(self, data_dir=None, label_resolution=None, split='train',
                  ignore_labels=True):
         if data_dir is None:
             data_dir = download.get_dataset_directory(
                 'pfnet/chainercv/cityscapes')
-        if label_resol not in ['fine', 'coarse']:
-            raise ValueError('\'label_resol\' argment should be eighter '
+        if label_resolution not in ['fine', 'coarse']:
+            raise ValueError('\'label_resolution\' argment should be eighter '
                              '\'fine\' or \'coarse\'.')
 
         img_dir = os.path.join(data_dir, os.path.join('leftImg8bit', split))
-        resol = 'gtFine' if label_resol == 'fine' else 'gtCoarse'
+        resol = 'gtFine' if label_resolution == 'fine' else 'gtCoarse'
         label_dir = os.path.join(data_dir, resol)
         if not os.path.exists(img_dir) or not os.path.exists(label_dir):
             raise ValueError(
--- a/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
+++ b/chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
@@ -4,6 +4,7 @@ import os
 import numpy as np
 
 from chainer import dataset
+from chainer.dataset import download
 from chainercv.datasets.cityscapes.cityscapes_utils import cityscapes_labels
 from chainercv.utils import read_image
 
@@ -16,41 +17,56 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
 
     .. note::
 
-        Please download the data by yourself because Cityscapes dataset doesn't
-        allow to re-distribute their data.
+        Please manually downalod the data because it is not allowed to
+        re-distribute Cityscapes dataset.
 
     Args:
-        img_dir (string): Path to the image dir. It should end with
-            ``leftImg8bit``.
-        label_dir (string): Path to the dir which contains labels. It should
-            end with either ``gtFine`` or ``gtCoarse``.
+        img_dir (string): Path to the image directory. It should end with
+            :obj`leftImg8bit`. If :obj:`None` is given, it uses
+            :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes/leftImg8bit`
+            as default.
+        label_dir (string): Path to the directory which contains labels. It
+            should end with either :obj`gtFine` or :obj:`gtCoarse`. If
+            :obj:`None` is given, it uses
+            :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes/gtFine`
+            as default.
         split ({'train', 'val'}): Select from dataset splits used in
             Cityscapes dataset.
-        ignore_labels (bool): If True, the labels marked ``ignoreInEval``
+        ignore_labels (bool): If True, the labels marked :obj:`ignoreInEval`
             defined in the original
             `cityscapesScripts<https://github.com/mcordts/cityscapesScripts>_`
-            will be replaced with `-1` in the `get_example` method.
+            will be replaced with :obj:`-1` in the :meth:`get_example` method.
+            The default value is :obj:`True`
 
     """
 
-    def __init__(self, img_dir, label_dir, split='train', ignore_labels=True):
+    def __init__(self, img_dir=None, label_dir=None, split='train',
+                 ignore_labels=True):
+        data_root = download.get_dataset_directory(
+            'pfnet/chainercv/cityscapes')
+        base_path = os.path.join(data_root, 'cityscapes')
+        if img_dir is None:
+            img_dir = os.path.join(base_path, 'leftImg8bit')
+        if label_dir is None:
+            label_dir = os.path.join(base_path, 'gtFine')
         img_dir = os.path.join(img_dir, split)
         self.ignore_labels = ignore_labels
 
-        self.label_fns, self.img_fns = [], []
+        self.label_fnames = list()
+        self.img_fnames = list()
         resol = os.path.basename(label_dir)
-        for dname in glob.glob('{}/*'.format(label_dir)):
+        for dname in glob.glob(os.path.join(label_dir, '*')):
             if split in dname:
-                for label_fn in glob.glob(
-                        '{}/*/*_labelIds.png'.format(dname)):
-                    self.label_fns.append(label_fn)
-        for label_fn in self.label_fns:
-            img_fn = label_fn.replace(resol, 'leftImg8bit')
-            img_fn = img_fn.replace('_labelIds', '')
-            self.img_fns.append(img_fn)
+                for label_fname in glob.glob(
+                        os.path.join(dname, '*', '*_labelIds.png')):
+                    self.label_fnames.append(label_fname)
+        for label_fname in self.label_fnames:
+            img_fname = label_fname.replace(resol, 'leftImg8bit')
+            img_fname = label_fname.replace('_labelIds', '')
+            self.img_fnames.append(img_fname)
 
     def __len__(self):
-        return len(self.img_fns)
+        return len(self.img_fnames)
 
     def get_example(self, i):
         """Returns the i-th example.
@@ -68,18 +84,15 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
             the dtype of the label image is :obj:`numpy.int32`.
 
         """
-        img = read_image(self.img_fns[i])
+        img = read_image(self.img_fnames[i])
         label_orig = read_image(
-            self.label_fns[i], dtype=np.int32, color=False)[0]
+            self.label_fnames[i], dtype=np.int32, color=False)[0]
         H, W = label_orig.shape
         if self.ignore_labels:
             label_out = np.ones((H, W), dtype=np.int32) * -1
             for label in cityscapes_labels:
-                if label.ignoreInEval:
-                    label_out[np.where(label_orig == label.id)] = -1
-                else:
+                if not label.ignoreInEval:
                     label_out[np.where(label_orig == label.id)] = label.trainId
         else:
             label_out = label
-        img = img.astype(np.float32)
         return img, label_out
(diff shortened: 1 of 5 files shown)


chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+
+    """
+
+    def __init__(self, img_dir, label_dir, split='train', ignore_labels=True):
+        img_dir = os.path.join(img_dir, split)
+        self.ignore_labels = ignore_labels
+
+        self.label_fns, self.img_fns = [], []
+        resol = os.path.basename(label_dir)
+        for dname in glob.glob('{}/*'.format(label_dir)):
```
yuyu2172: How about using `os.path.join(label_dir, '*')` instead?
This will work even if there is `/` at the end of `img_dir`.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+            if split in dname:
+                for label_fn in glob.glob(
+                        '{}/*/*_labelIds.png'.format(dname)):
```
yuyu2172: ditto.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+        split ({'train', 'val'}): Select from dataset splits used in
+            Cityscapes dataset.
+        ignore_labels (bool): If True, the labels marked ``ignoreInEval``
+            defined in the original
+            `cityscapesScripts<https://github.com/mcordts/cityscapesScripts>_`
+            will be replaced with `-1` in the `get_example` method.
+
+    """
+
+    def __init__(self, img_dir, label_dir, split='train', ignore_labels=True):
```
yuyu2172: How about setting a default directory names?
if `img_dir` is None,
how about setting the img_dir as `CHAINER_DATASET_ROOT/pfnet/chainercv/cityscapes/leftImg8bit`?
Same for `label_dir`.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+        self.ignore_labels = ignore_labels
+
+        self.label_fns, self.img_fns = [], []
+        resol = os.path.basename(label_dir)
+        for dname in glob.glob('{}/*'.format(label_dir)):
+            if split in dname:
+                for label_fn in glob.glob(
+                        '{}/*/*_labelIds.png'.format(dname)):
+                    self.label_fns.append(label_fn)
+        for label_fn in self.label_fns:
```
yuyu2172: https://github.com/chainer/chainercv/pull/161#discussion_r117882350

`fn` is not used in ChainerCV.
However, I think that `filenames` is too long. An alternative would be `fnames`.

@Hakuyume

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+            The dtype of the color image is :obj:`numpy.float32` and
+            the dtype of the label image is :obj:`numpy.int32`.
+
+        """
+        img = read_image(self.img_fns[i])
+        label_orig = read_image(
+            self.label_fns[i], dtype=np.int32, color=False)[0]
+        H, W = label_orig.shape
+        if self.ignore_labels:
+            label_out = np.ones((H, W), dtype=np.int32) * -1
```
yuyu2172: We can optimize this part by
1. not initializing `label_out`.
2. for loop only labels in ignore lists

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+            for label in cityscapes_labels:
+                if label.ignoreInEval:
+                    label_out[np.where(label_orig == label.id)] = -1
+                else:
+                    label_out[np.where(label_orig == label.id)] = label.trainId
+        else:
+            label_out = label
+        img = img.astype(np.float32)
```
yuyu2172: `read_image` guarantees that `img` is `np.int32`.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+        img_dir (string): Path to the image dir. It should end with
+            ``leftImg8bit``.
+        label_dir (string): Path to the dir which contains labels. It should
+            end with either ``gtFine`` or ``gtCoarse``.
+        split ({'train', 'val'}): Select from dataset splits used in
+            Cityscapes dataset.
+        ignore_labels (bool): If True, the labels marked ``ignoreInEval``
+            defined in the original
+            `cityscapesScripts<https://github.com/mcordts/cityscapesScripts>_`
+            will be replaced with `-1` in the `get_example` method.
```
yuyu2172:
```
`-1`  -->  :obj:`-1`
`get_example` -->  :meth:`get_example`
```

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+        allow to re-distribute their data.
+
+    Args:
+        img_dir (string): Path to the image dir. It should end with
+            ``leftImg8bit``.
+        label_dir (string): Path to the dir which contains labels. It should
+            end with either ``gtFine`` or ``gtCoarse``.
+        split ({'train', 'val'}): Select from dataset splits used in
+            Cityscapes dataset.
+        ignore_labels (bool): If True, the labels marked ``ignoreInEval``
```
yuyu2172: Please use obj.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+    .. note::
+
+        Please download the data by yourself because Cityscapes dataset doesn't
+        allow to re-distribute their data.
+
+    Args:
+        img_dir (string): Path to the image dir. It should end with
+            ``leftImg8bit``.
+        label_dir (string): Path to the dir which contains labels. It should
+            end with either ``gtFine`` or ``gtCoarse``.
```
yuyu2172: ditto

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+    .. _`Cityscapes dataset`: https://www.cityscapes-dataset.com
+
+    .. note::
+
+        Please download the data by yourself because Cityscapes dataset doesn't
+        allow to re-distribute their data.
+
+    Args:
+        img_dir (string): Path to the image dir. It should end with
+            ``leftImg8bit``.
```
yuyu2172: ditto

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+
+    .. _`Cityscapes dataset`: https://www.cityscapes-dataset.com
+
+    .. note::
+
+        Please download the data by yourself because Cityscapes dataset doesn't
+        allow to re-distribute their data.
+
+    Args:
+        img_dir (string): Path to the image dir. It should end with
```
yuyu2172: dir->directory

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+            ``leftImg8bit``.
+        label_dir (string): Path to the dir which contains labels. It should
```
yuyu2172: dir --> directory

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
yuyu2172:
```
True --> :obj:`True`
```

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
yuyu2172: Please add the following in the end.
 `The default value is :obj:`True``.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+            `cityscapesScripts<https://github.com/mcordts/cityscapesScripts>_`
+            will be replaced with `-1` in the `get_example` method.
+
+    """
+
+    def __init__(self, img_dir, label_dir, split='train', ignore_labels=True):
+        img_dir = os.path.join(img_dir, split)
+        self.ignore_labels = ignore_labels
+
+        self.label_fns, self.img_fns = [], []
```
yuyu2172: How about using two lines and use `list`.
`self.label_fnames = list()`
`self.img_fnames = list()`

tests/datasets_tests/cityscapes_tests/test_cityscapes.py
```diff
@@ -0,0 +1,59 @@
...
+
+            label = np.random.randint(0, 20, size=(128, 160)).astype(np.uint8)
+            label = Image.fromarray(np.zeros((128, 160), dtype=np.uint8))
+            label.save(os.path.join(
+                label_dir,
+                'aachen_000000_0000{:02d}_gtFine_labelIds.png'.format(i)))
+
+        img_dir = os.path.join(self.temp_dir, 'leftImg8bit')
+        label_dir = os.path.join(self.temp_dir, 'gtFine')
+        if self.split == 'test':
```
yuyu2172: This is unnecessary.

tests/datasets_tests/cityscapes_tests/test_cityscapes.py
```diff
@@ -0,0 +1,59 @@
...
+            self.temp_dir, 'leftImg8bit/{}/aachen'.format(self.split))
+        label_dir = os.path.join(
+            self.temp_dir, 'gtFine/{}/aachen'.format(self.split))
+        os.makedirs(img_dir)
+        os.makedirs(label_dir)
+
+        for i in range(10):
+            img = np.random.randint(0, 255, size=(128, 160, 3))
+            img = Image.fromarray(img.astype(np.uint8))
+            img.save(os.path.join(
```
yuyu2172: Please use `write_image`.
https://github.com/chainer/chainercv/pull/382

tests/datasets_tests/cityscapes_tests/test_cityscapes.py
```diff
@@ -0,0 +1,59 @@
...
+
+from chainer import testing
+from chainer.testing import attr
+from chainercv.datasets import cityscapes_label_names
+from chainercv.datasets import CityscapesSemanticSegmentationDataset
+from chainercv.utils import assert_is_semantic_segmentation_dataset
+
+
+@testing.parameterize(
+    {'split': 'train'},
```
yuyu2172: Please test `ignore_labels` (True, False).

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+
+class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
+
+    """Dataset class for a semantic segmentation task on `Cityscapes dataset`_.
+
+    .. _`Cityscapes dataset`: https://www.cityscapes-dataset.com
+
+    .. note::
+
+        Please download the data by yourself because Cityscapes dataset doesn't
```
yuyu2172: How about this.

Please manually downalod the data because it is not allowed to re-distribute Cityscapes dataset.

yuyu2172:
Thank you for a great PR.
One thing about the code block in ChainerCV.
We have been using 
```
:obj: `XXX`
```

instead of 
```
``XX``
```

This is because I initially thought that the red highlight emphasizes a word too much.
Could you follow this convention for consistency?

yuyu2172:
Also, could you add this to the docs?
```docs/source/reference/dataset.rst```

File: chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py (hunk 3 above)
Comment (mitmul) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: OK, but for label_dir, we can't assume which label users use.

File: chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py (hunk 4 above)
Comment (mitmul) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: Oops, sorry for the same mistake.

Comment (mitmul) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: Is it OK to use fnames?

File: chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py (hunk 5 above)
Comment (mitmul) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: When make the loop only over the ignore id list, how can I replace the label ids which are not marked as an `ignoreInEval` with `trainId`?

File: chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py (hunk 6 above)
Comment (mitmul) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: not np.float32?

Comment (yuyu2172) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: Sorry it is typo.

File: chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py (hunk 5 above)
Comment (yuyu2172) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: Ahh. I see. My bad.
It looks OK.
Is it OK to remove line 79?
Also, np.where is not necessary.

File: chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py (hunk 4 above)
Comment (Hakuyume) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: I prefer `filenames` to `fnames`. It is not so long.

Comment (yuyu2172) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: `filenames` is OK, but `label_filenames` and `img_filenames` are bit too long.

File: chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py (hunk 3 above)
Comment (yuyu2172) [inline chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py]: By having a default directory value, users do not need to specify the directory path once they set up their files properly. I think that this feature is very useful.

How about changing the options to `data_dir` and `label_mode`.
(i.e. `def __init__(self, data_dir=None, label_mode=None, split='train', ignore_labels=True):`)
The `data_dir` would be the path to the root dir whose default value is `CHAINER_DATASET_ROOT/pfnet/chainercv/cityscapes`.
Below the root directory, we expect at least two folders  (`leftImg8bit` and a label directory that is going to be used).
The `label_mode` should raise an error when unspecified. It should be either `fine` or `coarse`.

File: chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py [hunk 18]
```diff
@@ -0,0 +1,85 @@
...
+    """
+
+    def __init__(self, img_dir, label_dir, split='train', ignore_labels=True):
+        img_dir = os.path.join(img_dir, split)
+        self.ignore_labels = ignore_labels
+
+        self.label_fns, self.img_fns = [], []
+        resol = os.path.basename(label_dir)
+        for dname in glob.glob('{}/*'.format(label_dir)):
+            if split in dname:
```
yuyu2172: Does it work for Coarse dataset as well?
From the Github page, it seems that there is train_extra split.
https://github.com/mcordts/cityscapesScripts

I have not yet looked at the dataset by myself, so please tell me if I am wrong.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
Hakuyume: How about `label_paths` and `img_paths`?

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
Hakuyume: How about adding a check of `[]/list()` to our coding style chceker?

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
mitmul: That's why I used `if split in dname` at L:59

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
Hakuyume: #396

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
yuyu2172: Can you add that to the doc and test?

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
yuyu2172: The difference between `*_filenames` and `*_paths` is unclear.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
mitmul: OK, I'll reflect your suggestion and add a test for it.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
Hakuyume: `*_paths` is shorter than `*_filenames` and `_fnames`. If the problem of `*_filenames` is its length, `*_paths` will be a good solution.

yuyu2172: I agree.
I think changing all `*_filenames` to `*_paths` is good. Leaving both is bad.

I will update other datasets accordingly.

@mitmul 
Can you use `*_paths`?

mitmul:
Fixed. Replaced *_fnames with *_paths.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,105 @@
...
+    .. note::
+
+        Please manually downalod the data because it is not allowed to
+        re-distribute Cityscapes dataset.
+
+    Args:
+        data_dir (string): Path to the dataset directory. The directory should
+            contain at least two directories, :obj:`leftImg8bit` and either
+            :obj:`gtFine` or :obj:`gtCoarse`. If :obj:`None` is given, it uses
+            :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes` as default.
```
yuyu2172: by default.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,105 @@
...
+                 ignore_labels=True):
+        if data_dir is None:
+            data_dir = download.get_dataset_directory(
+                'pfnet/chainercv/cityscapes')
+        if label_mode is None:
+            raise ValueError('You need to give some value to \'label_mode\' '
+                             'argment.')
+        elif label_mode != 'fine' and label_mode != 'coarse':
+            raise ValueError('\'label_name\' argment should be eighter '
+                             '\'fine\' or \'coarse\'. But {} was '
```
yuyu2172: I think it is OK to simplify this. (delete line 45 and line 46)

```
if label_mode not in ['fine', 'coarse']:
    raise ValueError('\'label_name\' argment should be eighter '
                               '\'fine\' or \'coarse\'.')
```

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,105 @@
...
+        label_dir = os.path.join(data_dir, resol)
+        self.ignore_labels = ignore_labels
+
+        self.label_paths = list()
+        self.img_paths = list()
+        city_dnames = list()
+        for dname in glob.glob(os.path.join(label_dir, '*')):
+            if split in dname:
+                for city_dname in glob.glob(os.path.join(dname, '*')):
+                    for label_fname in glob.glob(
```
yuyu2172: fname -> path

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,105 @@
...
+                            os.path.join(city_dname, '*_labelIds.png')):
+                        self.label_paths.append(label_fname)
+                        city_dnames.append(os.path.basename(city_dname))
+        for city_dname, label_fname in zip(city_dnames, self.label_paths):
```
yuyu2172: ditto

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,105 @@
...
+            label_fname = os.path.basename(label_fname)
+            img_fname = label_fname.replace(
```
yuyu2172: ditto

mitmul:
@yuyu2172 Thanks, fixed.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,101 @@
...
+        if data_dir is None:
+            data_dir = download.get_dataset_directory(
+                'pfnet/chainercv/cityscapes')
+        if label_mode not in ['fine', 'coarse']:
+            raise ValueError('\'label_name\' argment should be eighter '
+                             '\'fine\' or \'coarse\'.')
+
+        img_dir = os.path.join(data_dir, os.path.join('leftImg8bit', split))
+        resol = 'gtFine' if label_mode == 'fine' else 'gtCoarse'
+        label_dir = os.path.join(data_dir, resol)
```
yuyu2172: We can easily anticipate users to instantiate this object expecting that ChainerCV would download the dataset.
This happens with `CityscapesSemanticSegmentationDataset()` without properly setting files under the default locations.
We should give a clear guidance to the users on where they should prepare the datasets when the error occurs.
This can be done in the similar way to ResNet in Chainer. https://github.com/chainer/chainer/blob/master/chainer/links/model/vision/resnet.py#L695

Perhaps raise an error here when either of the two necessary directories do not exist?

Error message can be something like this.

```
'Cityscapes dataset does not exist at the expected location.'
'Please download it from https://www.cityscapes-dataset.com/.'
'Then place directory leftImg8bit at {} and {} at {}.'.format(
    img_dir, resol, label_dir)
```

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Please manually downalod the data because it is not allowed to
+        re-distribute Cityscapes dataset.
+
+    Args:
+        data_dir (string): Path to the dataset directory. The directory should
+            contain at least two directories, :obj:`leftImg8bit` and either
+            :obj:`gtFine` or :obj:`gtCoarse`. If :obj:`None` is given, it uses
+            :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes` by default.
+        label_mode (string): The resolution of the labels. It should be either
```
yuyu2172: string -> {'fine', 'val'}

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,101 @@
...
+            :obj:`fine` or :obj:`coarse`.
+        split ({'train', 'val'}): Select from dataset splits used in
+            Cityscapes dataset.
+        ignore_labels (bool): If True, the labels marked :obj:`ignoreInEval`
+            defined in the original
+            `cityscapesScripts<https://github.com/mcordts/cityscapesScripts>_`
+            will be replaced with :obj:`-1` in the :meth:`get_example` method.
+            The default value is :obj:`True`
```
yuyu2172: Forgetting a period.

yuyu2172:
Great. Few more comments.

mitmul:
Thanks for the comments. Updated it.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,108 @@
...
+
+        Please manually downalod the data because it is not allowed to
+        re-distribute Cityscapes dataset.
+
+    Args:
+        data_dir (string): Path to the dataset directory. The directory should
+            contain at least two directories, :obj:`leftImg8bit` and either
+            :obj:`gtFine` or :obj:`gtCoarse`. If :obj:`None` is given, it uses
+            :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes` by default.
+        label_mode ({'fine', 'coarse'}): The resolution of the labels. It
```
yuyu2172: label_mode -> label_resol

I think this is more specific and better. Also this is consistent with the variable name used inside the code.
Sorry for the extra work.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,108 @@
...
+            Cityscapes dataset.
+        ignore_labels (bool): If True, the labels marked :obj:`ignoreInEval`
+            defined in the original
+            `cityscapesScripts<https://github.com/mcordts/cityscapesScripts>_`
+            will be replaced with :obj:`-1` in the :meth:`get_example` method.
+            The default value is :obj:`True`.
+
+    """
+
+    def __init__(self, data_dir=None, label_mode=None, split='train',
```
yuyu2172: label_mode -> label_resol

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,108 @@
...
+                 ignore_labels=True):
+        if data_dir is None:
+            data_dir = download.get_dataset_directory(
+                'pfnet/chainercv/cityscapes')
+        if label_mode not in ['fine', 'coarse']:
+            raise ValueError('\'label_mode\' argment should be eighter '
```
yuyu2172: label_mode -> label_resol

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,108 @@
...
+            tuple of a color image and a label whose shapes are (3, H, W) and
+            (H, W) respectively. H and W are height and width of the image.
+            The dtype of the color image is :obj:`numpy.float32` and
+            the dtype of the label image is :obj:`numpy.int32`.
+
+        """
+        img = read_image(self.img_paths[i])
+        label_orig = read_image(
+            self.label_paths[i], dtype=np.int32, color=False)[0]
+        H, W = label_orig.shape
```
yuyu2172: I think we can delete this and use `np.ones(label_orgi.shape, dtype=np.int32) *  -1` instead for line 102.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,108 @@
...
+        if self.ignore_labels:
+            label_out = np.ones((H, W), dtype=np.int32) * -1
+            for label in cityscapes_labels:
+                if not label.ignoreInEval:
+                    label_out[np.where(label_orig == label.id)] = label.trainId
```
yuyu2172: We can remove `np.where`.

yuyu2172:
`label_resol` seems better. Sorry for the extra work. `label_mode` was my idea...

mitmul:
@yuyu2172 Thanks for the good catches. I fixed them.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
Hakuyume: I think `resol` is difficult to understand.  How about using `use_fine_label`, which takes a boolean? If you want to support the case of three resolutions, I think `label_resolution` is better.

yuyu2172: Thanks for you feedback.
+1 for `label_resolution`.

mitmul: How about using `type`? 

> `type` the type/modality of data, e.g. gtFine for fine ground truth, or leftImg8bit for left 8-bit images.

https://github.com/mcordts/cityscapesScripts/blob/8815d4643222abc0f0a41614745ffa1637734335/README.md#dataset-structure

yuyu2172: `label_resolution=`coarse` is more explicit than `use_fine_label=False`, and it makes the code more readable.
It is not obvious that the opposite of `use_fine_label` is `use_coarse_label`.

yuyu2172: OK. `label_type` is better.

Hakuyume: `label_type` sounds `np.int32` or `np.float32`. If you want shorter form, how about `label_level`?

mitmul: If it's `label_dtype`, it surely sounds like a datatype, but I think `label_type` is OK. Anyway, just keeping the consistency of naming from the official scripts is better I think.

yuyu2172: `label_level` is not intuitive to me.

`label_type` can be confused as dtype.
It seems that `label_resolution` is fine. The length does not bother me much.

mitmul: OK, `label_resolution` works for me.

yuyu2172: The problem with `label_type` is that it is not specific to labels in the original code.
It accepts  leftImg8bit.

Hakuyume: > Anyway, just keeping the consistency of naming from the official scripts is better I think.

This rule may break the consistency in ChainerCV.

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Please manually downalod the data because it is not allowed to
+        re-distribute Cityscapes dataset.
+
+    Args:
+        data_dir (string): Path to the dataset directory. The directory should
+            contain at least two directories, :obj:`leftImg8bit` and either
+            :obj:`gtFine` or :obj:`gtCoarse`. If :obj:`None` is given, it uses
+            :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes` by default.
+        label_resolutionution ({'fine', 'coarse'}): The resolution of the
```
yuyu2172: typo

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
mitmul: Ah, right. Well, `label_resolution` is OK for you?

yuyu2172:
LGTM!

PR #398, opened 2017-08-18
Change variable names: filenames to paths

yuyu2172:
Related #392

### DIFF
--- a/chainercv/datasets/camvid/camvid_dataset.py
+++ b/chainercv/datasets/camvid/camvid_dataset.py
@@ -83,12 +83,12 @@ class CamVidDataset(chainer.dataset.DatasetMixin):
             data_dir = get_camvid()
 
         img_list_filename = os.path.join(data_dir, '{}.txt'.format(split))
-        self.filenames = [
+        self.paths = [
             [os.path.join(data_dir, fn.replace('/SegNet/CamVid/', ''))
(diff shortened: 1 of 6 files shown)


chainercv/datasets/camvid/camvid_dataset.py
```diff
@@ -108,7 +108,7 @@ def get_example(self, i):
         """
         if i >= len(self):
             raise IndexError('index is too large')
-        img_filename, label_filename = self.filenames[i]
+        img_filename, label_filename = self.paths[i]
```
Hakuyume: `img_filename` -> `img_path`
`label_filename` -> `label_path`

chainercv/datasets/camvid/camvid_dataset.py
```diff
@@ -83,12 +83,12 @@ def __init__(self, data_dir='auto', split='train'):
             data_dir = get_camvid()
 
         img_list_filename = os.path.join(data_dir, '{}.txt'.format(split))
```
Hakuyume: `img_list_filename` -> `img_list_path`

Hakuyume:
I found some `filename` s in other part than `datasets`. How do you think of them? I think some of them should be kept for the consistency with chainer.

yuyu2172:
You are right.
I think it is OK to leave them as is for consistency.

PR #399, opened 2017-08-18
Add return_bb option to CUBDatasets and add a test

yuyu2172:


tests/datasets_tests/cub_tests/test_cub_label_dataset.py
```diff
@@ -0,0 +1,26 @@
...
+class TestCUBLabelDataset(unittest.TestCase):
+
+    def setUp(self):
+        self.dataset = CUBLabelDataset(crop_bbox=self.crop_bbox)
+
+    @attr.slow
+    def test_cub_label_dataset(self):
+        assert_is_classification_dataset(
+            self.dataset, len(cub_label_names), n_example=10)
+
```
Hakuyume: Is it possible to check the effect of `crop_bbox`?

yuyu2172: It is totally different thing, but what do you think about changing this dataset to return `bbox` instead of returning a cropped image.
This is more flexible because in some cases users want to crop an image by a padded bbox.

This can be done by replacing `crop_bbox` option to `return_bbox`. This style of interface is similar to `VOCDetectionDataset`, which also can return extra data.

Hakuyume: Returning `bbox` looks good to me. As you said, it will be more flexible. Perhaps, adding `transforms.image.crop_with_bbox (or simply crop)` will be helpful. With this function, we can write cropped dataset easily, 
`TransformDataset(CUBLabelDataset(return_bbox), lambda in_data: crop_with_bbox(in_data[0], in_data[2]), in_data[1])`.

Hakuyume:
I noticed we have both `LabelDataset (CUBLabelDataset)` and `ClassificationDataset (DirectoryParsingClassificationDataset)`. We should unify these names because their tasks are same.

yuyu2172:
You are right.
`ClassificationDataset` is good because `label` is overloaded too much.
Also, it is consistent with other names of datasets which contain task names.

On the other hand, `LabelDataset` is good because this can be used for tasks other than Classification.

Hakuyume:
From my understanding, `Classification` in the class name indicates the main task for which the dataset was designed. Its name does not limit the usage of the dataset.

As you pointed out, we can use annotation type instead of task type. In this case, we should change class names as follows:
- `ClassificationDataset` -> `ImagewiseLabelDataset`
- `DetectionDataset` -> `BBoxWithLabelDataset` (?)
- `SemanticSegmentationDataset` -> `PixelwiseLabeldataset`

yuyu2172:
In my opinion, there are three options. I slightly prefer 2 or 3.

1. Name by the main tasks for which the dataset was designed.
2. Name the dataset by the most distinctive annotation.

In this case, I would also suggest the following names:
+ `ClassificationDataset` -> `LabelDataset`
+ `DetectionDataset` -> `BboxDataset`
+ `SemanticSegmentationDataset` -> `SemanticSegmentationDataset`.

One advantage with this naming convention is that it only uses notations that have already been used in ChainerCV.

3. Treat `LabelDataset` as a special case. `LabelDataset` can be used in multiple tasks such as Classification and Image Retrieval. This is the primary reason why `ClassificationDataset` sounds wrong. On the other hand, there is one to one correspondence between a task and a dataset type for `Detection` and `SemanticSegmentation`. There is no problem with assigning a task name in this case.

Hakuyume:
>    ClassificationDataset -> LabelDataset
    DetectionDataset -> BboxDataset
    SemanticSegmentationDataset -> SemanticSegmentationDataset.

`SemanticSegmentationDataset` looks inconsistent. This is the name of task.

> On the other hand, there is one to one correspondence between a task and a dataset type for Detection and SemanticSegmentation. 

This is not true. For example, a detection dataset can be used for object counting task.

yuyu2172:
> This is not true. For example, a detection dataset can be used for object counting task.

I see. So it seems that option 3 is too arbitrary.

> SemanticSegmentationDataset looks inconsistent. This is the name of task.

Although it is the name of the task, it is the name of the output.

Hakuyume:
> Although it is the name of the task, it is the name of the output. This can be improved.

How about `SemanticMask`?

tests/datasets_tests/cub_tests/test_cub_label_dataset.py
Hakuyume: I think `return_bbox` is an inaccurate name because we defined `bbox` as a set of bounding boxes.

yuyu2172: I was thinking of returning a set of bbox (shape=(1,4)).

yuyu2172:
I am not sure if that name is common in the field.

Hakuyume:
Personally, I prefer `Name by the main tasks for which the dataset was designed`. Is there anyone who think "I cannot use this dataset for image retrieval task because this is named Classification"?

tests/datasets_tests/cub_tests/test_cub_label_dataset.py
Hakuyume: From my understanding, `CUB` has only one bbox for each image. Is the first axis of `(1, 4)` redundant?

yuyu2172: Yes, it is redundant.

yuyu2172: But, I think it is better to keep data type consistent with other bboxes, so that we can use tools for bboxes.

yuyu2172:
On top of the inherent inconsistency problem, the task name is longer (`Label` -> `Classification`).
I observe this phenomenon a lot.

+ `Keypoint` -> `Pose Estimation`
+ `Caption` -> `Question Answering` (Although in this case, we can abbreviate the task name to QA)
+ `Scene Graph` -> `Scene Graph Generation`

I took a quick look at COCO and Visual Genome, which are prominent datasets that cover multiple tasks.

> SemanticSegmentationDataset looks inconsistent. This is the name of task.

I think a user would not get confused about the name of `Bbox/Detection` dataset just because there is `SemanticSegmentationDataset`.

Hakuyume:
> I think a user would not get confused about the name of Bbox/Detection dataset just because there is SemanticSegmentationDataset.

What do you mean?

yuyu2172:
It is totally fine to name `SemanticSegmentationDataset` together with `BboxDataset`.
We can have a task and an annotation whose names are the same.

Hakuyume:
> We can have a task and an annotation whose names are the same.

Yes, that is not the problem. However, `Segmentation` sounds `separating a thing into some pieces`. Do we call `separated pieces` as `segmentation`? Aren't they `segments`? This is the reason why `SemanticSegmentationDataset` looks strange to me.

yuyu2172:
`segmentation` can have the same meaning as `segment`.

I googled the word, and it is used in two ways.
https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FAST

Hakuyume:
> segmentation can have the same meaning as segment.

I see. So, `SemanticSegmentationDataset` is OK.
The names of datasets will be `LabelDataset`, `BboxDataset` and `SemanticSegmentationDataset`, right?
Perhaps, it is better to rename `(pixel-wise) label` to `segmentation (seg/segm?)`. For example, `vis_label` -> `vis_segmentation`.

yuyu2172:
> Perhaps, it is better to rename (pixel-wise) label to segmentation (seg/segm?)

Looking back at our internal discussion, `label` was preferred over `segm` because there is no similar name for `score`.
`label/score` pair is used throughout the library and it was suggested to be used for Semantic Segmentation as well.

However, `vis_segmentation` is much more informative than `vis_label`.
Since the situation has changed, I think `segm` is good.

By the way, taking `instance_segmentation` into consideration, I think `vis_semantic_segmentation` is better.

Hakuyume:
> label/score pair is used throughout the library and it was suggested to be used for Semantic Segmentation as well.

I see.

> By the way, taking instance_segmentation into consideration, I think vis_semantic_segmentation is better.

I thought the functionality of `vis_instance_segmentation` is very similar to that of `vis_semantic_segmentation` and we can treat both task by one function. But, `vis_semantic_segmentation` looks better now.

Hakuyume:
Let me summarize.
- datasets -> `<Identity><Main-annoatation-type>Dataset`
- dataset assertions -> `assert_is_<main-annotation-type>_dataset`
- visualizations-> `vis_<main-annoatation-type>`
- model assertions -> `assert_is_<task>_link`

yuyu2172:
I think it is OK to change `label` to `segm`.

Hakuyume:
> I think it is OK to change label to segm.

- `LabelDataset` -> `img, label`
- `BboxDataset` -> `img, bbox, label`
- `SemanticSegmentationDataset` -> `img, segm`

yuyu2172:
Adding to that, there are following objects:

+ Evaluator:   `<metric>Evaluator`
+ VisReport  `<task>VisReport`
+ evaluations:  ` eval_<metric>`
+ transforms   `<operation>_<annotation-type>`  (e.g. `resize_bbox`)

Hakuyume:
> VisReport <task>VisReport

Considering the consistency with `vis_*`, shouldn't it be `<Main-annotation-type>VisReport`? Do you intend to specify the type of `target`?

> transforms <operation>_<annotation-type> (e.g. resize_bbox)

If the `<annotation-type>` is image, we omit it, right? (e.g. `random_crop_image` -> `random_crop`)

yuyu2172:
So there are three conventions:

1. Annotation convention:  (e.g. `img` is `CHW`)
2. Dataset convention:  Set of annotations returned by a dataset  (e.g. `img, label` is returned by `LabelDataset`))
3. Task convention: Input and output of a network (e.g. Classification task handles a network that takes `img, label` during training as input. It takes `img` and outputs `score` (or `prob`) during testing.)

3 depends on 2 and 2 depends on 1.
For example, Detection task assumes the `BboxDataset` as a dataset.

> Considering the consistency with vis_*, shouldn't it be <Main-annotation-type>VisReport? Do you intend to specify the type of target?

Since `VisReport` assumes the input and output of a network, I think its behavior is selected by a task.


> If the <annotation-type> is image, we omit it, right? (e.g. random_crop_image -> random_crop)

Yes. This is for convenience.

tests/datasets_tests/cub_tests/test_cub_label_dataset.py
Hakuyume: I think name `return_bb` and returns `(4,)` is better. User can bbox utils by `bb[np.newaxis]/bb[None]`.

Hakuyume:
Those three conventions look good.

> Since VisReport assumes the input and output of a network, I think its behavior is selected by a task.

I see. This is reasonable.

> Yes. This is for convenience.

OK.

yuyu2172:
I think we considered vision problems pretty comprehensively.

We considered tasks with
+ different dataset and different test time output
+ the same dataset and different test time output  (`Classification` and `Image Retrieval`)
+ different dataset and the same test time output  (Faster R-CNN and Fast R-CNN?)
+ the same dataset and the same test time output, but different for other reasons (is there any???)

tests/datasets_tests/cub_tests/test_cub_label_dataset.py
yuyu2172: OK.

yuyu2172:
I found that `<task>_label_names` is used.
Should it be `<annotation>_label_names`?

Hakuyume:
> I found that <task>_label_names is used.
Should it be <annotation>_label_names?

If the annotation type is `label`. `label_label_names` sounds strange...
`bbox_label_names` is OK.

yuyu2172:
Perhaps, `imagewise_label_names`?

Hakuyume:
> Perhaps, imagewise_label_names?

If so, `LabelDataset` should be `ImagewiseLabelDataset`?

yuyu2172:
Yeah. I think that is better.

One more thing.
`DirectoryParsingClassificationDataset/DirectoryParsingImagewiseLabelDataset` can be shortened to `DirectoryParsingDataset` because `Imagewise Label` is the only annotation that can be extracted by the names of subdirectories.

Hakuyume:
Is it possible to use `label_names` when the annotation type is image-wise label?

> Imagewise Label is the only annotation that can be extracted by the names of subdirectories

Perhaps, we can make semantic segmentation/image conversion dataset by parsing directory. As you pointed out, we can extract only image-wise label from the name of subdirectories. However, `DirectoryParsing` sounds more general.

yuyu2172:
> Is it possible to use label_names when the annotation type is image-wise label?

Do you mean this?

+ `label_names` (not `imagewise_label_names`)
+ `LabelDataset` instead of `ImagewiseLabelDataset`.

It looks OK. This can be better because it is shorter.

> However, DirectoryParsing sounds more general.

Yes. This may sound too general. People can expect this to work as a directory parsing semantic segmentation dataset.

yuyu2172:
@Hakuyume 
Let's finish this.

Hakuyume:
> Do you mean this?
>
>    label_names (not imagewise_label_names)
>    LabelDataset instead of ImagewiseLabelDataset.

> It looks OK. This can be better because it is shorter.

I mean both.

| dataset | values | additional info |
| :-: | :-: | :-: | 
| LabelDataset | img, label | label_names |
| BoundingboxDataset | img, bbox, label | boundingbox_label_names |
| SemanticSegmentationDataset | img, label | semantic_segmentation_label_names, semantic_segmentation_label_colors |

yuyu2172:
I would prefer `bbox_label_names` over `boundngbox_label_names`.
`bbox` is used throughout the library.

Other than that, it looks ok.

Hakuyume:
> I would prefer bbox_label_names over boundngbox_label_names.
bbox is used throughout the library.

Sorry, it's my mistake.
 
| dataset | values | additional info |
| :-: | :-: | :-: | 
| LabelDataset | img, label | label_names |
| BoundingboxDataset | img, bbox, label | bbox_label_names |
| SemanticSegmentationDataset | img, label | semantic_segmentation_label_names, semantic_segmentation_label_colors |

Hakuyume:
> If we choose to use LabelDataset for datasets with imagewise annotations, DirectoryParsingLabelDataset would be a good name.

I agree with you.

yuyu2172:
Sorry. I forgot to point out, but I prefer the dataset name for `bbox` to be `BboxDataset`.

I will start implementing these changes.

Hakuyume:
> Sorry. I forgot to point out, but I prefer the dataset name for bbox to be BboxDataset.

Sorry, I fixed it. And I added two columns 'visualizer' and 'related tasks'.

yuyu2172:
The new API (https://github.com/chainer/chainercv/pull/399#issuecomment-328745157) is implemented:

+ #405
+ #419 
+ #420

yuyu2172:
Please merge this after #405.

tests/datasets_tests/cub_tests/test_cub_label_dataset.py
yuyu2172: I reflected this change.

yuyu2172:
Good idea.

yuyu2172:
@Hakuyume 
I found that `mask` is not really an appropriate name for this data because it can be in any value between [0, 255].
I think `prob_map` is better name for this. 
Since, this is a relative big change, how about making another PR?

Hakuyume:
> I found that mask is not really an appropriate name for this data because it can be in any value between [0, 255].
I think prob_map is better name for this.

Is it a value of probability? If so, can we scale it to `[0, 1)`?

> Since, this is a relative big change, how about making another PR?

Yes. it is better?

yuyu2172:
Yes.

Please review this first.

PR #400, opened 2017-08-18
Add a test for OnlineProductsDataset

yuyu2172:


Hakuyume:
How about checking `super_label`?

yuyu2172:
Thank you your suggestion. I updated

PR #404, opened 2017-08-20
Fix resize_contain with odd margin

### LINKED ISSUE #401: Bug in transforms.resize_contain
When the target size has odd number as its element, `chainercv.transforms.resize_contain` raises an error as follows:

```
from chainercv import transforms
import numpy as np

img = np.zeros((3, 10, 10))
img = transforms.resize_contain(img, (11, 11))
```

```
site-packages/chainercv/transforms/image/resize_contain.py in resize_contain(img, size, fill, return_param)
     57     out_img = np.empty((C, out_H, out_W), dtype=img.dtype)
     58     out_img[:] = np.array(fill).reshape((-1, 1, 1))
---> 59     out_img[:, y_slice, x_slice] = img
     60
     61     if return_param:

ValueError: could not broadcast input array from shape (3,10,10) into shape (3,9,9)
```

Hakuyume:
fix #401

### DIFF
--- a/chainercv/transforms/image/resize_contain.py
+++ b/chainercv/transforms/image/resize_contain.py
@@ -1,3 +1,4 @@
+from __future__ import division
 import numpy as np
 
 from chainercv.transforms import resize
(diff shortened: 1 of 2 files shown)


yuyu2172:
LGTM. Thanks.

PR #412, opened 2017-08-31
Add interpolation options to scale

akitotakeki:
`transforms/image/resize` can select interpolation options, but `transforms/image/scale` can't.
So I add interpolation options to `scale`.


yuyu2172:
Thank you very much!
LGTM

PR #350, opened 2017-07-21
Add Faster R-CNN VOC07+12 experiment

yuyu2172:
Merge after #265

PR #418, opened 2017-09-13
Raise helpful warnings with Faster R-CNN train script

yuyu2172:
Error message
```
$ python train.py                                                                
Traceback (most recent call last):
  File "train.py", line 161, in <module>
    main()
  File "train.py", line 76, in main
    'Some hyperparameters are not set.\n'
ValueError: Some hyperparameters are not set.
Please specify the following arguments: 'dataset', 'step_size' and 'iteration'.
ex: python train.py --gpu 0 --dataset voc07 --step_size 50000 --iteration 70000
```

### DIFF
--- a/examples/detection/eval_voc07.py
+++ b/examples/detection/eval_voc07.py
@@ -48,7 +48,7 @@ def main():
                 n_fg_class=20,
                 pretrained_model=args.pretrained_model)
         else:
-            model = FasterRCNNVGG16(pretrained_model='voc0712')
+            model = FasterRCNNVGG16(pretrained_model='voc07')
     elif args.model == 'ssd300':
         if args.pretrained_model:
             model = SSD300(
--- a/examples/detection/visualize_models.py
+++ b/examples/detection/visualize_models.py
@@ -13,7 +13,7 @@ from chainercv.visualizations import vis_bbox
 def main():
     dataset = VOCDetectionDataset(year='2007', split='test')
     models = [
-        ('Faster R-CNN', FasterRCNNVGG16(pretrained_model='voc0712')),
+        ('Faster R-CNN', FasterRCNNVGG16(pretrained_model='voc07')),
         ('SSD300', SSD300(pretrained_model='voc0712')),
         ('SSD512', SSD512(pretrained_model='voc0712')),
     ]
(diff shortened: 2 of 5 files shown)


Hakuyume:
Why don't you use `required` option? https://docs.python.org/3/library/argparse.html#required

yuyu2172:
Thanks. I did not know the option.

The page says the following. 
```
Note Required options are generally considered bad form because users expect options to be optional, and thus they should be avoided when possible.
```

Maybe, it is better to set default values.
In that case, the default values should be for the simplest experiment, which is VOC07 in this case.

Hakuyume:
> Maybe, it is better to set default values.
In that case, the default values should be for the simplest experiment, which is VOC07 in this case.

I agree. The options with `--<name>` style should be optional. (at least in `argparse`). It is a good idea to use VOC07's setting as the default values.

Hakuyume:
Perhaps, it is better to use VOC0712's setting because we switched the default setting from VOC07 to VOC0712 in #350.

yuyu2172:
It depends on what you want to set as the default training parameters.
Personally, the most simple setting should be preferred for the default parameters.
"Simple" means that a script finishes quickly and the GPU memory requirement is low.
Although, the difference is not huge, VOC07 ends faster than VOC0712 and downloads less images.
Analogously, if ResNet is added to the train script, VGG feature extractor should be the default extractor.

For evaluation, the choice of pretrained weights do not matter for speed, so it can be anything.
I think it is fine to use weights trained with VOC0712, which is performing the best.

Hakuyume:
I know why you chose `VOC07` as the default of training and `VOC0712` as that of testing. However, I thought the difference between training and testing is confusing for users. Perhaps, some user wonder why the score of the trained model is lower than that of the distributed model.

Hakuyume:
> I think it is fine to use weights trained with VOC0712, which is performing the best.

Will you change the default value whenever we add a better model?

yuyu2172:
> I know why you chose VOC07 as the default of training and VOC0712 as that of testing. However, I thought the difference between training and testing is confusing for users. Perhaps, some user wonder why the score of the trained model is lower than that of the distributed model.

How about making all pretrained weight options to the simplest setting? (in this case VOC07)

Hakuyume:
> How about making all pretrained weight options to the simplest setting? (in this case VOC07)

It is OK.

yuyu2172:
Updated

PR #419, opened 2017-09-14
Use bbox* instead of detection* for datasets and label_names

yuyu2172:
Related https://github.com/chainer/chainercv/pull/399#issuecomment-328745157

+ `VOCDetectionDataset` -> `VOCBboxDataset`
+ `voc_detection_label_names` -> `voc_bbox_label_names`
+ `assert_is_detection_dataset` -> `assert_is_bbox_dataset`

### DIFF
--- a/chainercv/datasets/voc/voc_bbox_dataset.py
+++ b/chainercv/datasets/voc/voc_bbox_dataset.py
@@ -9,7 +9,7 @@ from chainercv.datasets.voc import voc_utils
 from chainercv.utils import read_image
 
 
-class VOCDetectionDataset(chainer.dataset.DatasetMixin):
+class VOCBboxDataset(chainer.dataset.DatasetMixin):
 
     """Dataset class for the detection task of PASCAL `VOC`_.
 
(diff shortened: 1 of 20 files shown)


chainercv/datasets/voc/voc_bbox_dataset.py
```diff
@@ -9,7 +9,7 @@
 from chainercv.utils import read_image
 
 
-class VOCDetectionDataset(chainer.dataset.DatasetMixin):
+class VOCBboxDataset(chainer.dataset.DatasetMixin):
 
     """Dataset class for the detection task of PASCAL `VOC`_.
```
Hakuyume: `Dataset class for the detection task` is intended? This description is true because it is designed for detection. However, this is against the concept of separating dataset and tasks.

yuyu2172: >  this is against the concept of separating dataset and tasks.

Right. I will fix this.

Hakuyume: I found `VOCSemanticSegmentationDataset` also has task-based description. https://github.com/chainer/chainercv/blob/master/chainercv/datasets/voc/voc_semantic_segmentation_dataset.py#L12
 Can you fix it?

yuyu2172: Thanks.
I made another PR.

docs/source/reference/datasets.rst
```diff
@@ -52,9 +52,9 @@ OnlineProductsDataset
 PASCAL VOC
 ----------
 
-VOCDetectionDataset
+VOCBboxDataset
 ~~~~~~~~~~~~~~~~~~~
```
Hakuyume: The length of `~`.

yuyu2172:
@Hakuyume 

Could you review this?

PR #427, opened 2017-09-20
Change SequentialFeatureExtractor to PickableSequentialChain

### LINKED ISSUE #425: Name of SequentialFeatureExtractor can be improved
Since there is no "extractor" other than "feature extractor", it is safe to name it SequentialExtractor.

yuyu2172:
Fix #425


chainercv/links/model/pickable_sequential_chain.py
```diff
@@ -0,0 +1,151 @@
...
+    When :obj:`pick` is an iterable of strings, a tuple of layers
+    is returned. The order of the layers is the same as the order of
+    the strings in :obj:`pick`.
+    When :obj:`pick` is :obj:`None`, the last layer is returned.
+
+    Examples:
+
+        >>> import chainer.functions as F
+        >>> import chainer.links as L
+        >>> model = PickableSequentialChain()
```
Hakuyume: `PickableSequentialChain` is not imported. (Is it OK?)

yuyu2172: It is a gray zone.
For instance, I left the documentation for VGG16 without import.

Do you think we should always import the class that is documented?
Even if we add the import line, the snippet still does not work because `x` is undefined.

Hakuyume: > Even if we add the import line, the snippet still does not work because x is undefined.

I see. So it's OK.

yuyu2172:
@Hakuyume 
Could you review this?

chainercv/links/model/vgg/vgg16.py
```diff
@@ -15,43 +15,41 @@
...
-    The value of :obj:`VGG16.feature_names` selects the features that are going
-    to be collected by :meth:`__call__`.
-    :obj:`self.all_feature_names` is the list of the names of features
-    that can be collected.
+    This is a feature extraction link.
+    The network can choose output layers from set of all
+    intermediate layers.
+    The attribute :obj:`pick` is the names of the layers that are going
+    to be picked by :meth:`__call__`.
+    The attribute :obj:`layer_names` is the names of layers
```
Hakuyume: How about `the names of all layers that can be picked`?

PR #436, opened 2017-09-29
Add ResNet training code

yuyu2172:
~~Merge after #652~~
~~Merge after #432, #435, #321~~

~~Aims at reproducing scheme used in https://github.com/facebook/fb.resnet.torch.~~
Edit:
The training code reproduces the scheme introduced below.
This performs better than `fb.resnet.torch` and scales to the situation with many GPUs.
https://arxiv.org/pdf/1706.02677.pdf

- [x] *Linear Scaling Rule* for lr (see the paper)
- [x] Set default value of gamma for BNs in the end of bottleneck to 0
- [x] use corrected momentum SGD
- [x] Have correct scale for weight decay (ChainerMN accomplishes this by default)
- [x] Learning rate scheduling happens at 30, 60, 80-th epoch.
- [x] stop using color related data augmentation

Edit 11/16:
Although "ResNet in 1 hour" paper is new, the training scheme is adapted by some researchers (e.g. [mixup](https://arxiv.org/abs/1710.09412)).


yuyu2172:
The result without setting `gamma` to 0 for the last BN of each ResBlock.

```
8 GPU
23.60

32 GPU
24.37
```

yuyu2172:
I updated the code with lr warmup. The behavior stays the same when 8GPUs are used.

results:
ResNet50 with 32GPUs
23.54

yuyu2172:
@Hakuyume 
Please review this

Hakuyume:
Both `a weight` and `weights` are used for "a set of parameters of a model". Which one is better?

yuyu2172:
I am checking how slow the training gets without options like `autotune`.

yuyu2172:
Thanks for reviewing.
I reflected ours discussions to the code.

examples/classification/README.md
```diff
@@ -32,15 +34,49 @@ Ten crop error rate.
...
+
+The scores of the models trained with `train_imagenet_multi.py`, which can be executed like below.
+Please consult the full list of arguments for the training script with `python train_imagenet_multi.py -h`.
+```
+$ mpiexec -n N python train_imagenet_multi.py <path_to_train_dataset> <path_to_val_dataset>
+```
+
+The training procedure carefully follows the "ResNet in 1 hour" paper [5].
+
+##### Performance tip
```
Hakuyume: Why did you use 5th level? (4th level `####` is not enough?)

examples/classification/train_imagenet_multi.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+class TrainTransform(object):
+
+    def __init__(self, mean):
+        self.mean = mean
+
+    def __call__(self, in_data):
+        img, label = in_data
+        _, H, W = img.shape
```
Hakuyume: `H` and `W` are not used?

examples/classification/README.md
```diff
@@ -1,6 +1,8 @@
 # Classification
 
-## Performance
+## ImageNet
+
+### Weight conversion
 
 Single crop error rate.
```
Hakuyume: For the consistency with `Trained model`, it would be nice to add something like `~ of the models converted from Caffe model`.

examples/classification/README.md
```diff
@@ -1,6 +1,8 @@
 # Classification
 
-## Performance
+## ImageNet
```
Hakuyume: `he` and `fb` should be distinguished from each other in score board.

PR #438, opened 2017-10-02
Refactor faster_rcnn.predict

knorth55:
- Use original image shape for bbox max
- fix maximum pixel
  - if image shape is `(H, W)`, max pixel is `(H-1, W-1)`

### DIFF
--- a/chainercv/links/model/faster_rcnn/faster_rcnn.py
+++ b/chainercv/links/model/faster_rcnn/faster_rcnn.py
@@ -270,22 +270,21 @@ class FasterRCNN(chainer.Chain):
 
         """
         prepared_imgs = list()
-        scales = list()
+        sizes = list()
         for img in imgs:
-            _, H, W = img.shape
+            size = img.shape[1:]
             img = self.prepare(img.astype(np.float32))
-            scale = img.shape[2] / W
             prepared_imgs.append(img)
-            scales.append(scale)
+            sizes.append(size)
 
         bboxes = list()
         labels = list()
         scores = list()
-        for img, scale in zip(prepared_imgs, scales):
+        for img, size in zip(prepared_imgs, sizes):
             with chainer.using_config('train', False), \
                     chainer.function.no_backprop_mode():
                 img_var = chainer.Variable(self.xp.asarray(img[None]))
-                H, W = img_var.shape[2:]
+                scale = img_var.shape[3] / size[1]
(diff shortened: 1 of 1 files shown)


yuyu2172:
Hi. Thanks for your interest.
We keep bounding box y-coordinates in range`[0, H]` and x-coordinates in range `[0, W]`, which follows float-coordinate convention.
There are two coordinate type conventions:
+ Integer convention limits coordinates in range `[0, H - 1]` and `[0, W - 1]`. In this convention, the area of a box is computed by `(x_max - x_min + 1) * (y_max - y_min + 1)`.
+ Float convention uses ranges `[0, H]` and `[0, W]`. The area is computed by `(x_max - x_min) * (y_max - y_min)`.

The float convention is adopted in newer datasets such as COCO.

The decision influenced every part where area of a bounding box is computed and values of coordinates are cropped.
For instance
 https://github.com/chainer/chainercv/blob/master/chainercv/utils/bbox/bbox_iou.py#L40
https://github.com/chainer/chainercv/blob/master/chainercv/utils/bbox/non_maximum_suppression.py#L137

knorth55:
thank you for your comment.
I remove the subtraction commit, and just clean up the code.

chainercv/links/model/faster_rcnn/faster_rcnn.py
```diff
@@ -270,22 +270,21 @@ def predict(self, imgs):
 
         """
         prepared_imgs = list()
-        scales = list()
+        shapes = list()
```
yuyu2172: Can you use `sizes` instead of `shapes`?
`size` is used specifically to refer to `H, W`, whereas `shape` is a general tuple.

Also, `shape` should be changed to `size`.

knorth55: I think `size` is not good because of `np.array.size`

yuyu2172: `shapes` sounds like `(C, H, W)`.
It is less likely that people mix up with `np.array.size`.
For consistency reasons, `size` is better.
(e.g. https://github.com/chainer/chainercv/blob/master/chainercv/transforms/image/resize.py#L44)

knorth55: ok

knorth55:
updated

yuyu2172:
LGTM

PR #443, opened 2017-10-05
Add prob_map option to CUBDataset

yuyu2172:
https://github.com/chainer/chainercv/pull/399#issuecomment-334373228

I will explain the rationale behind the name `prob_map`.
The name has not been introduced in ChainerCV, and usually the name `prob` is used instead for an pixel-wise probability.
Although this is fine in the case of semantic segmentation, it is not the case when used together with image-wise labels.
In this case, `CUBLabelDataset` uses image-wise labels, so it seemed better to assign name to the probability that can be recognized as pixel-wise probabilities easily.

`prob_map` seems to be good because it implies that this is a pixel-wise annotation and range is [0, 1].


Hakuyume:
Let me confirm.

- In semantic segmentation, we use `label` for pixel-wise data because users can know it from the context and the definition of `SemanticSegmentationDataset`.
- Without any contexts, we use `*_map` for pixel-wise data to show it is pixel-wise.

yuyu2172:
Yes. That is what I wanted to say.

Hakuyume:
In detection, we are using `score` for confidence value(s). We should define the difference between `score` and `prob`.

I think these definitions are good.

- `score`: confidence value(s) whose range is not limited.
- `prob`: probability value(s) whose range is `[0, 1]`.

yuyu2172:
I think that rule is good.
As you said, some detection functions use `scores` where `probs` should be used.

Hakuyume:
> As you said, some detection functions use scores where probs should be used.

Yes. But we should use `scores` in detection context because the API uses `scores`.

yuyu2172:
Should we change that to `probs` as well?

Hakuyume:
> Should we change that to probs as well?

I think we should not because some detection models return confidence values which exceed `1`.

yuyu2172:
OK.
Then how about using `prob` when we always want it to be in range [0, 1].
Therefore, `prob` is a subset of `score`.

Hakuyume:
> Then how about using prob when we always want it to be in range [0, 1].

It is OK. Please note that `prob` implies not only the range is `[0, 1]` but also the sum of probs is `1`.

> Therefore, prob is a subset of score.

Yes.

chainercv/datasets/cub/cub_keypoint_dataset.py
```diff
@@ -91,25 +92,38 @@ def __init__(self, data_dir='auto', return_bb=False,
...
                 return img, keypoint, kp_mask
 
-        path, _ = os.path.splitext(self.paths[i])
-        mask = utils.read_image(
-            os.path.join(self.mask_dir, path + '.png'),
-            dtype=np.uint8,
-            color=False)
+        prob_map = utils.read_image(self.prob_map_paths[i],
+                                    dtype=np.uint8, color=False)
+        prob_map = prob_map.astype(np.float32) / 255  # [0, 255] -> [0, 1]
```
Hakuyume: In python2, `from __future__ import division` is required.

yuyu2172:
> It is OK. Please note that prob implies not only the range is [0, 1] but also the sum of probs is 1.

OK.
In the case of `CUBDataset`, the probability map should be called `fg_prob_map` because it does not sum to one.

yuyu2172:
Well, I guess it is ok to call fg/bg probability object as `prob`.

PR #448, opened 2017-10-07
Fix inconsistency in 'data_dir'

Hakuyume:
`Cityscapes*Dataset` uses `None` as the default value of `data_dir` though other datasets use `auto`.


yuyu2172:
LGTM

PR #453, opened 2017-10-13
Add COCOBboxDataset

yuyu2172:
Depends on #452

I think we need some modification to the BboxDataset API.

1. bbox can be zero-sized `(0,4)`.
2. Make it easier to filter images with no ground truth bounding boxes. I suggest to make a public method `labels` for all bbox datasets. The method returns a list of bounding boxes. Also, I suggest to support similar method for all label datasets.


Hakuyume:
>  Make it easier to filter images with no ground truth bounding boxes. I suggest to make a public method labels for all bbox datasets. The method returns a list of bounding boxes. Also, I suggest to support similar method for all label datasets.

I guess the advantage of `labels` is that users can filter images without loading them. Is ii right?

yuyu2172:
Yea

Hakuyume:
I made a mix-in for this problem. Please check #454.

yuyu2172:
Areas of instance masks are needed for evaluation. So, I added an option that makes the dataset to return them.

yuyu2172:
ok

docs/source/reference/datasets.rst
```diff
@@ -51,6 +51,13 @@ CityscapesTestImageDataset
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 .. autoclass:: CityscapesTestImageDataset
 
+COCO
```
Hakuyume: I prefer `MS COCO` because we use `PASCAL VOC` in this doc.

PR #454, opened 2017-10-14
Add SliceableDataset

Hakuyume:
related to #453 

This PR adds `AnnotatedImageDatasetMixin`. 

This mix-in requires these methods.

- `__len__`: returns the length of dataset
- `get_image`: takes an index and returns an image
- `get_annotation`: takes an index and returns annotation(s)

This mix-in provides these attribute/method.

- `__getitem__`: takes an index/slice and returns example(s). An example is a tuple of image and annotations. If `get_annotation` returns `(anno0, anno1)`, an example will be `(img, anno0, anno1)`.
- `annotations`: a dataset without images. If `get_annotation` returns `(anno0, anno1)`, an example of this dataset will be `(anno0, anno1)`.


yuyu2172:
I like how you kept the design simple.
I think it is worthwhile to discuss possibility to support accessing annotations quickly without an extra abstraction.

I found two problems.
First, the dataset needs to retrieve all annotations together.
This may become annoying in a situation when annotations are `label, mask`.
The loading of `mask` will prevent users from quickly accessing `labels`.
Second, this may be less efficient in the case when annotations do not need to be fetched sequentially.
This is the case when retrieving the entire labels for ImageNet.
I ran a small benchmark, and there is a noticeable overhead.
 
```python
import time
import numpy as np

x = np.arange(1000000)   # 1M

start = time.time()
out = list()
for i in range(len(x)):
    out.append(x[i])
print(time.time() - start)  # 0.1742 s
```

Also, there is another person making similar feature in Chainer project.

Hakuyume:
> First, the dataset needs to retrieve all annotations together.
This may become annoying in a situation when annotations are label, mask.
The loading of mask will prevent users from quickly accessing labels.

Yes. If annotations contain some large data, my API is not efficient.
For example, my API is not good for semantic segmentation dataset.

> Second, this may be less efficient in the case when annotations do not need to be fetched sequentially.
This is the case when retrieving the entire labels for ImageNet.
I ran a small benchmark, and there is a noticeable overhead.

What do you mean? 

> Also, there is another person making similar feature in Chainer project.
chainer/chainer#3252

I see. That PR is more general than mine. If that PR is merged, we don't need to add `labels` to BboxDataset APIs.

yuyu2172:
> Second, this may be less efficient in the case when annotations do not need to be fetched sequentially.
This is the case when retrieving the entire labels for ImageNet.
I ran a small benchmark, and there is a noticeable overhead.

If a dataset has an array `labels` as an attribute, you can access it instantly (i.e. `labels = dataset.labels`).
In your design, you need to run a for-loop to get labels, which is slow.

Hakuyume:
> If a dataset has an array labels as an attribute, you can access it instantly (i.e. labels = dataset.labels).
In your design, you need to run a for-loop to get labels, which is slow.

I see. That makes sense. However, if users use only a subset of the dataset and labels are stored under multiple files, returning a list is not efficient. For example, If I want to filter the first 100 examples of VOCBboxDataset, loading `labels` for 101-st image is unnecessary.

yuyu2172:
> If I want to filter the first 100 examples of VOCBboxDataset, loading labels for 101-st image is unnecessary.

I see.

Since the purpose of this PR conflicts with #3252 and this is not unique to Vision, I think the design should be discussed with Chainer team.
BTW, your design seems to be better because extending it is a lot more intuitive.

yuyu2172:
> Since the purpose of this PR conflicts with #3252 and this is not unique to Vision, I think the design should be discussed with Chainer team.

Since #3252 does not seem to solve the efficiency problem that we have been discussing, I changed my mind and think that this dataset is appropriate for ChainerCV.
The solution it solves is specific to vision.
In that sense, `AnnotatedImageDatasetMixin` is different from datasets like `ConcatenatedDataset`.
Do you think this distinction is reasonable?

> First, the dataset needs to retrieve all annotations together.
This may become annoying in a situation when annotations are label, mask.
The loading of mask will prevent users from quickly accessing labels.

`img, label, mask` is quite common scenario (e.g. Instance Segmentation).
Do you have any workarounds in mind?

Hakuyume:
> Since #3252 does not seem to solve the efficiency problem that we have been discussing, I changed my mind and think that this dataset is appropriate for ChainerCV.
The solution AnnotatedImageDatasetMixin solves is specific to vision.
In that sense, it is different from datasets like ConcatenatedDataset.
Do you think this distinction is reasonable?

I think it is reasonable to keep our own mix-in. The style of https://github.com/chainer/chainer/pull/3252 is not optimized to ChainerCV. If we use https://github.com/chainer/chainer/pull/3252, we have to implement 5 methods for each dataset.
In the term of efficiency, https://github.com/chainer/chainer/pull/3252 seems to solve the efficiency problem. We can override `extract_feature` to avoid calling `get_example`.

> img, label, mask is quite common scenario (e.g. Instance Segmentation).
Do you have any workarounds in mind?

A simple solution is excluding `mask` from `get_annotation` and overriding `get_example` to return `mask`. Perhaps, `get_lightweight_annotation` is a proper name.
The code will be as follows.
```python
    def get_lightweight_annotation(self, i):
        ...
        return label

    def get_image(self, i):
        ...
        img = read_image(...)
        return img

    def get_mask(self, i):
        ...
        mask = read_image(...)
        ...
        return mask

    def get_example(self, i):
        label = self.get_lightweight_annotation(i)
        img = self.get_image(i)
        mask = self.get_mask(i)
        return img, mask, label
```
This workaround can be used for `prob_map` of CUB.

yuyu2172:
> A simple solution is excluding mask from get_annotation and overriding get_example to return mask. Perhaps, get_lightweight_annotation is a proper name.
The code will be as follows.

I have two problems.
First, there is a situation when it is more efficient and readable to compute `label` and `mask` together.
This happens in the case when logic to compute the two annotations are similar.
For instance, here https://github.com/yuyu2172/chainercv/blob/5c11243351ce7dfb95be659ee343a74e262e614d/chainercv/datasets/coco/coco_bbox_dataset.py#L151.

Second, it became more complicated than what it was.
`get_annotation` is obvious about what it returns, but `get_lightweight_annotation` is not.

How about setting an attribute that configures `get_annotation` when necessary?
For instance, set an attribute like `label_only`.
This makes the dataset to only load `labels` when `annotations` is called.


BTW, shouldn't `get_annotation` be `get_annotations`?

Hakuyume:
> First, there is a situation when it is more efficient and readable to compute label and mask together.
This happens in the case when logic to compute the two annotations are similar.
For instance, here https://github.com/yuyu2172/chainercv/blob/5c11243351ce7dfb95be659ee343a74e262e614d/chainercv/datasets/coco/coco_bbox_dataset.py#L151.

In that case, `get_annotation` should contain `mask` because it does not require heavy computation. 

> Second, it became more complicated than what it was.
get_annotation is obvious about what it returns, but get_lightweight_annotation is not.

Yes. That is a problem.

yuyu2172:
> In that case, get_annotation should contain mask because it does not require heavy computation.

What do you mean?
`mask` is still heavy to load.
It gets slightly more efficient if computed together with other annotations.
Ideally, it should be possible to load all of them together and load annotations without masks.

Hakuyume:
I'm trying to implement better APIs.

Hakuyume:
How about this API? I have implemented `PickableDataset`.

```python
class MyDataset(PickableDataset):
    def __init__(self):
        super().__init__()

        self.data_names = ('img', 'bbox', 'label', 'mask')
        self.add_getter('img', self.get_image)
        self.add_getter('mask', self.get_mask)
        self.add_getter(('bbox', 'label'), self.get_bbox_label)

    def get_image(self, i):
        print('get_image')
        return 'img_{:d}'.format(i)

    def get_mask(self, i):
        print('get_mask')
        return 'mask_{:d}'.format(i)

    def get_bbox_label(self, i):
        print('get_bbox_label')
        return 'bbox_{:d}'.format(i), 'label_{:d}'.format(i)


dataset = MyDataset()
print(dataset[0])
# get_image
# get_bbox_label
# get_mask
# ('img_0', 'bbox_0', 'label_0', 'mask_0')

picked_dataset = dataset.pick('label')
print(picked_dataset[1])
# get_bbox_label
# label_1
```

yuyu2172:
I am not sure whether we should introduce a new abstraction or remain with the minimal abstraction (`DatasetMixIn`).
Since supporting easy access to `labels` can be achieved without any new abstraction, I would like to know why this kind of abstraction is needed.

In my opinion, by the nature that the class is intended to be extended by users, this has to be perfect or should not be used.

I think there is still to be improved.
For instance, `PickableDataset` returns `PickedDataset` with method `pick`.
But, the returned object doesn't support `pick`.
Since `pick()` is somewhat equivalent to `a[:, ***]` of arrays, it feels more natural if the returned dataset supports the interface.

Hakuyume:
> Since supporting easy access to labels can be achieved without any new abstraction, I would like to know why this kind of abstraction is needed.

I don't think providing partial access only for `labels` is not enough. For example, some user wants to filter images by the size of bounding boxes. In this case, partial access for `bboxes` is required. With this class, we can implement partial access for all data easily. Especially, `return_***` options will get more simple. Please check #457.

> In my opinion, by the nature that the class is intended to be extended by users, this has to be perfect or should not be used.

I agree. I would like to make this class perfect as much as possible.

> I think there is still to be improved.
For instance, PickableDataset returns PickedDataset with method pick.
But, the returned object doesn't support pick.
Since pick() is somewhat equivalent to a[:, ***] of arrays, it feels more natural if the returned dataset supports the interface.

Good idea. I'll fix it.

yuyu2172:
> I agree. I would like to make this class perfect as much as possible.

Ideally, it is beneficial to have a dataset abstraction that never changes its functionality by modifying it.
For instance, currently, an attribute becomes private when a dataset is used together with `SubDataset` or `TransformDataset`. The problem with `pick` is similar to this problem.

On top of that, we have another demand that subset of dataset annotations should be easily and efficiently accessible. This is the problem that we have been discussing from the beginning.

With that in mind, I suggest to introduce a new dataset class possibly different from `DatasetMixIn`.
In short, I suggest to make two styles of datasets for Chainer. The first one involves thin abstractions (e.g. `DatasetMixIn` and `TransformDataset`). The second one is more complex, but combines all the functionalities together. Users can pick based on their preference on the level of abstraction they want from a base dataset class.

If we take this strategy, https://github.com/chainer/chainer/pull/3343 and https://github.com/chainer/chainer/pull/3252 should be replaced.


> I don't think providing partial access only for labels is not enough. For example, some user wants to filter images by the size of bounding boxes. In this case, partial access for bboxes is required. With this class, we can implement partial access for all data easily. Especially, return_*** options will get more simple. Please check #457.

We can add `bboxes` as a property as well.

Hakuyume:
> We can add bboxes as a property as well.

Will you add `labels`, `bboxes`, `labels_bboxes`, `difficulties`, `labels_difficulties`, `bboxes_difficulties` ... ?

yuyu2172:
> Will you add labels, bboxes, labels_bboxes, difficulties, labels_difficulties, bboxes_difficulties ... ?

I see what you want to say. If all of them are going to be supported, we need to take more general approach with base dataset class.
If ad hoc approach is taken, I would probably start with `labels` only or `labales and `bboxes`.

Hakuyume:
https://github.com/chainer/chainer/pull/3976

Hakuyume:
@yuyu2172 Please review #522 first.

Hakuyume:
import from https://github.com/chainer/chainer/pull/3976

yuyu2172:
Can you add at least one code example to the doc string?
Also, it would be better to include in the docstrings that the entire module is planned to be removed.

yuyu2172:
@Hakuyume 
Can you work on this?

chainercv/chainer_experimental/datasets/sliceable/concatenated_dataset.py
```diff
@@ -0,0 +1,45 @@
+from chainercv.chainer_experimental.datasets.sliceable import SliceableDataset
+
+
+class ConcatenatedDataset(SliceableDataset):
+    """A sliceable version of :class:`chainer.datasets.ConcatenatedDataset`.
+
+    Hew is an example.
```
yuyu2172: Here

chainercv/chainer_experimental/datasets/sliceable/getter_dataset.py
```diff
@@ -0,0 +1,91 @@
...
+class GetterDataset(SliceableDataset):
+    """A sliceable dataset class that defined by getters.
+
+    This ia a dataset class with getters.
+
+    Here is an example.
+
+    >>> class SliceableLabeledImageDataset(GetterDataset):
+    >>>     def __init__(self, pairs, root='.'):
+    >>>         super().__init__()
```
yuyu2172: Python 2 would not work.

chainercv/chainer_experimental/datasets/sliceable/getter_dataset.py
```diff
@@ -0,0 +1,91 @@
...
+    >>>     def get_label(self, i):
+    >>>         _, label = self._pairs[i]
+    >>>         return np.int32(label)
+    >>
+    >>> dataset = SliceableLabeledImageDataset('list.txt')
+    >>>
+    >>> # get a subset with label = 0, 1, 2
+    >>> # no images are loaded
+    >>> indices = [i for i, label in
+    >>> enumerate(dataset.slice[:, 'label']) if label in {0, 1, 2}]
```
yuyu2172: indentation is wiered

chainercv/chainer_experimental/datasets/sliceable/sliceable_dataset.py
```diff
@@ -0,0 +1,137 @@
...
+            return self.get_example_by_keys(
+                index, tuple(range(len(self.keys))))
+        else:
+            return self.get_example_by_keys(index, (0,))[0]
+
+    @property
+    def slice(self):
+        return SliceHelper(self)
+
+    def __iter__(self):
```
yuyu2172: Do we need this?

chainercv/chainer_experimental/datasets/sliceable/sliceable_dataset.py
```diff
@@ -0,0 +1,137 @@
...
+        return t,
+
+
+class SliceableDataset(chainer.dataset.DatasetMixin):
+    """An abstract dataset class that supports slicing.
+
+    This ia a dataset class that supports slicing.
+    A dataset class inheriting this class should implement
+    three methods: :meth:`__len__`, :meth:`keys`, and
+    :meth:`get_example_by_keys`.
```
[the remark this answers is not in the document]
yuyu2172: I mean the relationship between `SlicableDataset` and `GetterDataset` should be clear.

From users perspective,
they would first come and read `SlicableDataset`. Since this is not intended to be directly touched by users, we should guide them to `GetterDataset`.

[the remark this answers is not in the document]
Hakuyume: Nice suggestion. I agree with you.

docs/source/tutorial/sliceable.rst
```diff
@@ -0,0 +1,209 @@
...
+            bbox = generate_random_bbox(10, (224, 224), 10, 224)
+            label = np.random.randint(0, 9, size=10).astype(np.int32)
+            return bbox, label
+
+
+    dataset = SampleBboxdataset()
+    img, bbox, label = dataset[0]  # get_image(0) and get_annotation(0)
+
+    view = dataset.slice[:, 'label']
+    label = view[1]  # get_annotation(0)
```
yuyu2172: `get_annotaion(1)`

docs/source/tutorial/sliceable.rst
```diff
@@ -0,0 +1,209 @@
...
+
+        def get_annotation(self, i):
+            print('get_annotation({})'.format(i))
+            # generate dummy annotations
+            bbox = generate_random_bbox(10, (224, 224), 10, 224)
+            label = np.random.randint(0, 9, size=10).astype(np.int32)
+            return bbox, label
+
+
+    dataset = SampleBboxdataset()
```
yuyu2172: typo

docs/source/tutorial/sliceable.rst
```diff
@@ -0,0 +1,209 @@
...
+
+
+Usage: slice along with both axes
+---------------------------------
+
+.. code-block:: python
+    from chainercv.datasets import VOCBboxDataset
+    dataset = VOCBboxDataset()
+
+    # the view of label of the first 100 examples
```
yuyu2172: the view of the labels of the first 100 examples

docs/source/tutorial/sliceable.rst
```diff
@@ -0,0 +1,209 @@
...
+:method:`slice` takes indices of examples as its first argument.
+
+.. code-block:: python
+    from chainercv.datasets import VOCBboxDataset
+    dataset = VOCBboxDataset()
+
+    # the view of first 100 examples
+    view = dataset.slice[:100]
+
+    # the view of last 100 examples
```
yuyu2172: the last

docs/source/tutorial/sliceable.rst
```diff
@@ -0,0 +1,209 @@
...
+
+Usage: slice along with the axis of examples
+--------------------------------------------
+:method:`slice` takes indices of examples as its first argument.
+
+.. code-block:: python
+    from chainercv.datasets import VOCBboxDataset
+    dataset = VOCBboxDataset()
+
+    # the view of first 100 examples
```
yuyu2172: the first

docs/source/tutorial/sliceable.rst
```diff
@@ -0,0 +1,209 @@
...
+    img, label = view[0]
+
+
+Motivation
+----------
+:method:`slice` returns a view of the dataset without conducting data loading,
+where :method:`DatasetMixin.__getitem__` conducts :method:`get_example` for all required examples.
+Users can write efficient code by this view.
+
+This example counts the number of images that contain dogs.
```
yuyu2172: ... contain dogs without loading any images.

docs/source/tutorial/sliceable.rst
```diff
@@ -0,0 +1,209 @@
+Sliceable Dataset
+=================
+
+This tutorial will walk you through the features related to sliceable dataset.
+We assume that readers have a basic understanding of Chainer dataset (e.g. understand :class:`chainer.dataset.DatasetMixin`).
+
+In ChainerCV, we introduce `sliceable` feature to datasets.
+SliceableT datasets support :method:`slice` that returns a sub view of the dataset.
```
yuyu2172: typo

yuyu2172: a sub view --> a view

docs/source/tutorial/sliceable.rst
```diff
@@ -0,0 +1,209 @@
...
+
+This example that shows the basic usage.
```
yuyu2172: This example shows the basic usage.

docs/source/tutorial/sliceable.rst
```diff
@@ -0,0 +1,207 @@
...
+            # generate dummy annotations
+            bbox = generate_random_bbox(10, (224, 224), 10, 224)
+            label = np.random.randint(0, 9, size=10).astype(np.int32)
+            return bbox, label
+
+    dataset = SampleBboxDataset()
+    img, bbox, label = dataset[0]  # get_image(0) and get_annotation(0)
+
+    view = dataset.slice[:, 'label']
+    label = view[1]  # get_annotation(0)
```
yuyu2172: `get_annotation(1)`

chainercv/chainer_experimental/datasets/sliceable/getter_dataset.py
```diff
@@ -0,0 +1,91 @@
...
+    if isinstance(t, tuple):
+        return t
+    else:
+        return t,
+
+
+class GetterDataset(SliceableDataset):
+    """A sliceable dataset class that is defined with getters.
+
+    This ia a dataset class with getters.
```
yuyu2172: This ia --> This is a

yuyu2172: How about adding a comment that lets users know about the tutorial?

```
Please refer to the tutorial for more detailed explanation.
```

chainercv/chainer_experimental/datasets/sliceable/tuple_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+    >>> named_dataset.keys)  # ('feat0', None)
+    >>> # slice takes both key and index (or their mixture)
+    >>> named_dataset.slice[:, ['feat0', 1]][:])  # [(0, 0), (1, 1), (2, 4)]
+
+    Args:
+        datasets: The underlying datasets.
+            The following datasets are acceptable.
+
+            * An inheritance of \
+                :class:~chainer.datasets.sliceable.SliceableDataset`.
```
yuyu2172: does not compile properly

chainercv/chainer_experimental/datasets/sliceable/concatenated_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+    >>> dataset_a = TupleDataset([0, 1, 2], [0, 1, 4])
+    >>> dataset_b = TupleDataset([3, 4, 5], [9, 16, 25])
+    >>>
+    >>> dataset = ConcatenatedDataset(dataset_a, dataset_b)
+    >>> dataset.slice[:, 0][:]  # [0, 1, 2, 3, 4, 5]
+
+    Args:
+        datasets: The underlying datasets.
+            Each dataset should inherit
+            :class:~chainercv.chainer_experimental.datasets.sliceable.Sliceabledataset`
```
yuyu2172: does not compile properly

PR #458, opened 2017-10-19
fix bbox format in flip_bbox

knorth55:
bbox is in range of [0, H] or [0, W]


knorth55:
Test failing because test is also wrong because `bbox` can be `H` or `W`.
https://github.com/chainer/chainercv/blob/master/tests/transforms_tests/bbox_tests/test_flip_bbox.py#L17

yuyu2172:
Thanks.
LGTM.

PR #459, opened 2017-10-20
avoid zero division in bbox2loc

knorth55:


yuyu2172:
How about `np.minimum(..., 1e14)`?

Hakuyume:
How about `eps` which can get from `numpy.finfo`(https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.finfo.html) instead of magic value `1e-14`?

Hakuyume:
This is just a question. Are there any cases that `height` or `width` becomes `0`? According to `assert_is_bbox`, both `height` and `width` should be a positive number (https://github.com/chainer/chainercv/blob/master/chainercv/utils/testing/assertions/assert_is_bbox.py#L25-L28).

knorth55:
`rpn` output `rois` can be `height =0` and `width = 0` at initial state in training.
and this `rois` can be used `proposal_target_creator`

knorth55:
thank you for review.
I updated

Hakuyume:
> rpn output rois can be height =0 and width = 0 at initial state in training.
and this rois can be used proposal_target_creator

I see. That makes sense.

PR #461, opened 2017-10-23
fix setting parameters of random_distort

peisuke:
The parameters of random_distort doesn't seems work because the value is set with hard coded.
I fixed it.

### DIFF
--- a/chainercv/links/model/ssd/transforms.py
+++ b/chainercv/links/model/ssd/transforms.py
@@ -66,38 +66,38 @@ def random_distort(
         img[img > 255] = 255
         return img.astype(np.uint8)
 
-    def brightness(cv_img, brightness_delta):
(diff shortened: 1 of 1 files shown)


chainercv/links/model/ssd/transforms.py
```diff
@@ -66,46 +66,52 @@ def convert(img, alpha=1, beta=0):
         img[img > 255] = 255
         return img.astype(np.uint8)
 
-    def brightness(cv_img):
+    def brightness(cv_img, brightness_delta):
```
Hakuyume: `delta` is enough unique in the function `brightness`. Can you rename `brightness_delta` -> `delta`?

peisuke:
Thank you for review.
I updated.

Hakuyume:
Thank you!

PR #475, opened 2017-10-27
Improve appearence of y_offset and x_offset in doc

yuyu2172:
Before

transforms_ _chainercv_0_7_0_documentation

After

transforms_ _chainercv_0_7_0_documentation


Hakuyume:
It looks inconsistent with  `y_min, x_min, y_max, x_max`. Is it OK?

yuyu2172:
I generally prefer math expression over code block for readability in most cases.
How about changing `y_min, x_min, y_max, x_max` to math expressions instead?

Hakuyume:
> How about changing y_min, x_min, y_max, x_max to math expressions instead?

It is OK.

PR #476, opened 2017-10-27
Fix refs to NMS in SSD doc

yuyu2172:


chainercv/links/model/ssd/multibox_coder.py
```diff
@@ -198,7 +198,7 @@ def decode(self, mb_loc, mb_conf, nms_thresh=0.45, score_thresh=0.6):
             mb_conf (array): A float array whose shape is
                 :math:`(K, n\_fg\_class + 1)`.
             nms_thresh (float): The threshold value
-                for :meth:`chainercv.transfroms.non_maximum_suppression`.
+                for :meth:`~chainercv.utils.non_maximum_suppression`.
```
Hakuyume: I prefer `:func:` to `:meth:` because it is a regular function.

yuyu2172: OK

PR #485, opened 2017-10-28
Use math notations for bounding box coordinates

yuyu2172:
https://github.com/chainer/chainercv/pull/475#issuecomment-340109339


yuyu2172:
@Hakuyume

PR #488, opened 2017-11-13
fix IOError: image file truncated” with big image

jinjiren:
If the image size is too large, then such error will occur: 
`OSError: image file is truncated (13 bytes not processed)`.

The solution I found from stackoverflow: https://stackoverflow.com/questions/12984426/python-pil-ioerror-image-file-truncated-with-big-images.

I tested it and it actually works.


Hakuyume:
Thank you.
Could you add tests for such large images?

PR #497, opened 2017-12-17
Support MixUpSoftLabelDataset.

fukatani:
Between class learning is simple and effective technique, I think it is valuable to merge to chainercv.
https://arxiv.org/abs/1711.10284


yuyu2172:
Hi thanks for a timely PR. I am interested in supporting this feature.
I checked BC paper, and I found that it is slightly different from a more popular MixUp paper https://arxiv.org/abs/1710.09412.

1. MixUp takes positive pairs, but BC only takes negative pairs.
2. BC proposes two formulas for mixing two samples. The first one is the same as MixUp. The second one is different.

Since MixUp is simpler, it is easier to include in a library.
I would suggest to only add MixUp, and not BC to ChainerCV.
However, if there is a strong advantage for BC (note: I would like to know this!), it can be integrated into ChainerCV *together* with MixUp.

Having said that, I have several design comments.

1. Can you change the name to `MixUpSoftLabelDataset`? We call a dataset `*LabelDataset` when it returns an image and an integer. Since, `MixUp` returns an array of size `(n_class,)` as label, it needs to be differentiated. Thus, I would name it `*SoftLabelDataset`. Also, MixUp is easier to recognize, and I prefer the name better.
2. I found that it is common in CV to fetch two samples from two datasets. Thus, I made a dataset class 
 called `SiameseDataset` that supports this functionality in a different PR https://github.com/chainer/chainercv/pull/505. I ask you to wait until that one is merged to ChainerCV, then change your code to use `SiameseDataset`. Sorry for inconvenience with this point.

chainercv/datasets/between_class_dataset.py
```diff
@@ -0,0 +1,104 @@
...
+            two image mixture is `method(image1, image2, mix_ratio)`. Else,
+            Perform calculations according to the original paper. Default is
+            'bc_plus'.
+
+    See the papers for details: [1]`Between-class Learning for Image \
+    Classification <https://arxiv.org/abs/1711.10284>`_.
+
+    """
+
+    def __init__(self, dataset, dtype=numpy.float32, label_dtype=numpy.float32,
```
yuyu2172: Could you remove `dtype` and `label_dtype` from arguments?
By default, images and soft-labels are `dtype=np.float32`.
If different dtypes are needed, transforms can be used.

fukatani:
Thank you for your quick and attentive response.

OK, I  will change `BetweenClassLabeledDataset` to `MixUpSoftLabelDataset` and implement it as MixUp paper.

After merging this PR and if I can confirm advantage fo BC, I may suggest BC-like mixing as option.

> Sorry for inconvenience with this point.

Don't mind. Actually, I will take time to restart this PR.

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
+import numpy
```
yuyu2172: We use `np` in ChainerCV. Could you change it?

chainercv/datasets/__init__.py
```diff
@@ -15,6 +15,7 @@
 from chainercv.datasets.cub.cub_utils import cub_label_names  # NOQA
 from chainercv.datasets.directory_parsing_label_dataset import directory_parsing_label_names  # NOQA
 from chainercv.datasets.directory_parsing_label_dataset import DirectoryParsingLabelDataset  # NOQA
+from chainercv.datasets.mixup_dataset import MixUpSoftLabelDataset  # NOQA
```
yuyu2172: Can you change the name of the file to `mixup_soft_label_dataset`?
Also, could you change the name of the test file as well?

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+            More over, each element of each dataset should have same shape.
+            dataset also needs to support `__len__` as described above.
+        max_label: Max label index of base datasets.
+
+    See the papers for details: [1]`mixup: Beyond Empirical Risk Minimization
+    <https://arxiv.org/abs/1710.09412>`_.
+
+    """
+
+    def __init__(self, dataset, max_label):
```
yuyu2172: How about using `n_class` instead of `max_label`?
`n_class` is widely used in other parts of ChainerCV.

In that case the doc would be

```
n_class (int): The number of classes in the base dataset.
```

yuyu2172: You can remove `+1 in line 57 if you do this.

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+            >>> from chainer.datasets import get_mnist
+            >>> from chainercv.datasets import SiameseDataset
+            >>> from chainercv.datasets import MixUpSoftLabelDataset
+            >>> mnist, _ = get_mnist()
+            >>> base_dataset = SiameseDataset(mnist, mnist)
+            >>> dataset = MixUpSoftLabelDataset(base_dataset)
+            >>> mixed_image, mixed_label = dataset[0]
+
+    Args:
+        dataset: The underlying dataset. dataset should returns two image
```
yuyu2172: `daset should return ...` can be written more explicitly.

```
The dataset returns :obj:`img_0, label_0, img_1, label_1`, which is a tuple containing two pairs of an image and a label.
```

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+            and their label. Typically, dataset is `SiameseDataset`.
+            More over, each element of each dataset should have same shape.
```
yuyu2172: `Moreover, ...` can be rewritten as...

```
The shapes of images and labels should be constant.
```

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+            dataset also needs to support `__len__` as described above.
```
yuyu2172: I think `__len__` is not mentioned anywhere else. We can remove this line.

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+
+        .. code::
+
+            >>> from chainer.datasets import get_mnist
+            >>> from chainercv.datasets import SiameseDataset
+            >>> from chainercv.datasets import MixUpSoftLabelDataset
+            >>> mnist, _ = get_mnist()
+            >>> base_dataset = SiameseDataset(mnist, mnist)
+            >>> dataset = MixUpSoftLabelDataset(base_dataset)
+            >>> mixed_image, mixed_label = dataset[0]
```
yuyu2172: How about printing the shape? This is an important details.

```
>>> print(mixed_label.shape, mixed_label.dtype)   # ((10,), numpy.float32)
```

tests/datasets_tests/test_mixup_dataset.py
```diff
@@ -0,0 +1,63 @@
...
+        self.n_class = np.max((self.labels_0, self.labels_1)) + 1
+        self.siamese_dataset = SiameseDataset(dataset_0, dataset_1)
+
+    def _check_example(self, example):
+        assert_is_image(example[0])
+        self.assertEqual(example[0].shape, self.img_shape)
+        self.assertEqual(example[1].dtype, np.float32)
+        self.assertEqual(example[1].ndim, 1)
+        self.assertEqual(example[1].shape[0], self.n_class + 1)
+        self.assertAlmostEqual(example[1].sum(), 1.0)
```
yuyu2172: Can you check that the label is nonnegative?

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+
+class MixUpSoftLabelDataset(dataset_mixin.DatasetMixin):
+
+    """Dataset which returns mixed images and labels for mixup learning[1].
+
+    `MixUpSoftLabelDataset` mix two images and labels which is chosen by base
+    dataset respectively by weighted average.
+
+    Unlike `LabeledImageDatasets`, label is a one-dimensional float array with
+    at most two nonzero weights (i.e. soft label). The summed weights is one.
```
yuyu2172: The sum of the two weights is one

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+
+    The base dataset `__getitem__` should return image and label. Please see
```
yuyu2172: This is incorrect. We can remove this.

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
+import numpy
+
+from chainer.dataset import dataset_mixin
+
+
+class MixUpSoftLabelDataset(dataset_mixin.DatasetMixin):
+
+    """Dataset which returns mixed images and labels for mixup learning[1].
+
+    `MixUpSoftLabelDataset` mix two images and labels which is chosen by base
```
yuyu2172:
```
:class:`MixUpSoftLabelDataset` mixes two pairs of labeled images fetched from the base dataset.
```

yuyu2172:
Thank you for a nice PR!
I am so sorry for the late response.

chainercv/datasets/mixup_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+        We construct a mixup dataset from MNIST.
+
+        .. code::
+
+            >>> from chainer.datasets import get_mnist
+            >>> from chainercv.datasets import SiameseDataset
+            >>> from chainercv.datasets import MixUpSoftLabelDataset
+            >>> mnist, _ = get_mnist()
+            >>> base_dataset = SiameseDataset(mnist, mnist)
+            >>> dataset = MixUpSoftLabelDataset(base_dataset)
```
yuyu2172: You need to supply the number of class to `MixupSoftLabelDataset`

yuyu2172:
The link to the paper is not compiled properly.

screenshot from 2018-03-21 11 33 06

fukatani:
Thank you for your attentive review.
(And sorry for my poor doc)

I addressed your comments.

fukatani:
mix
I confirmed document was compiled successfully.

docs/source/reference/datasets.rst
```diff
@@ -18,6 +18,10 @@ SiameseDataset
 ~~~~~~~~~~~~~~
 .. autoclass:: SiameseDataset
 
+MixUpSoftLabelDataset
```
yuyu2172: Can you move this before `SiameseDataset`?
We are ordering docs in alphabetical order.

chainercv/datasets/mixup_soft_label_dataset.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, dataset, n_class):
+        self._dataset = dataset
+        self._n_class = n_class
+
+    def __len__(self):
+        return len(self._dataset)
+
+    def get_example(self, i):
+        image1, label1, image2, label2 = self._dataset[i]
```
yuyu2172: How about changing the names of the variables to `img_0, label_0, img_1, label_1` so that they are consistent with the documentation?

fukatani:
Thanks! I addressed your comments.

yuyu2172:
LGTM. Thank you for your contribution!

PR #503, opened 2017-12-20
[WIP] Add dataset class for OpenImages dataset

furushchev:
This PR adds dataset class for OpenImages dataset https://github.com/openimages/dataset
This is the first 'working' version and not elaborated at all, so any comments are welcome. :)


yuyu2172:
Very nice PR.

PR #504, opened 2017-12-20
SSD train example with multiple GPUs

Hakuyume:


examples/ssd/README.md
```diff
@@ -34,6 +34,11 @@ Note that this code requires `cv2` module.
 $ python train.py [--model ssd300|ssd512] [--batchsize <batchsize>] [--gpu <gpu>]
 ```
 
+If you want to use multiple GPUs, use `train_mulit.py`.
```
fukatani: `train_multi.py` is right.
(Sorry, I am reading this PR because it is interesting though this is WIP.)

Hakuyume: Thank you for pointing out. I will remove `[WIP]` after I confirm this script can reproduce the same performance.

examples/ssd/train_multi.py
```diff
@@ -0,0 +1,234 @@
...
+        mb_locs, mb_confs = self.model(imgs)
+        loc_loss, conf_loss = multibox_loss(
+            mb_locs, mb_confs, gt_mb_locs, gt_mb_labels, self.k)
+        loss = loc_loss * self.alpha + conf_loss
+
+        chainer.reporter.report(
+            {'loss': loss, 'loss/loc': loc_loss, 'loss/conf': conf_loss},
+            self)
+
+        # use the number of positives in all GPUs instead of that in this GPU
```
yuyu2172: use the total number of positives in all GPUs instead of the number of positives in each GPU

examples/ssd/train_multi.py
```diff
@@ -0,0 +1,234 @@
...
+        with chainer.cuda.get_device_from_array(gt_mb_labels):
+            if n_positive[0] > 0:
+                loss *= (gt_mb_labels > 0).sum() / n_positive[0]
+        return loss
+
+
+class Transform(object):
+
```
yuyu2172: If this class is completely same as the training script for single GPU, there should be a comment saying it.

examples/ssd/train_multi.py
```diff
@@ -0,0 +1,234 @@
...
+    for param in train_chain.params():
+        if param.name == 'b':
+            param.update_rule.add_hook(GradientScaling(2))
+        else:
+            param.update_rule.add_hook(WeightDecay(0.0005))
+
+    devices = {'main': args.gpu[0]}
+    devices.update(
+        ('other{:d}'.format(i + 1), gpu) for i, gpu in enumerate(args.gpu[1:]))
+    updater = training.ParallelUpdater(train_iter, optimizer, devices=devices)
```
yuyu2172: Do we need to make `train_multi.py` to behave the same way as `train.py`?
Does training get unstable if make changes?

Also, we should document some changes made from `train.py`, especially ones involving `n_positive`.

Hakuyume: > Do we need to make train_multi.py to behave the same way as train.py?
Does training get unstable if make changes?

I think it is bad that the number of GPUs affects the behaviour. We should keep the behaviour unchanged as much as possible. (I think it will converge even if we change the behaviour).
Do you want to change the behaviour according to the number of GPUs?

yuyu2172: > I think it is bad that the number of GPUs affects the behaviour.

Should we follow this rule for all multi-GPU implementations (e.g. ImageNet)?

> Do you want to change the behaviour according to the number of GPUs?

Well, I am wondering if we should use `ParallelUpdater`. Since this is the only option to implement the current behavior, we need to change the behavior if we use other updaters, such as ChainerMN and `MultiprocessParallelUpdater`.
In my opinion, we should always use the same updater for multi-GPU implementation, so that the community would not get confused about which one to use.

Is `ParallelUpdater` fast enough in this training script?
It can be slow because it does not use NCCL.

Hakuyume: > Should we follow this rule for all multi-GPU implementations (e.g. ImageNet)?

From my understanding, there is no concrete rule for multi-GPU implementations.
In addition, this implementation is not against to that of classification (softmax cross entropy does not change it's behaviour).

> Well, I am wondering if we should use ParallelUpdater. Since this is the only option to implement the current behavior, we need to change the behavior if we use other updaters, such as ChainerMN and MultiprocessParallelUpdater.

I agree that we should consider the cases of  ChainerMN or MultiprocessParallelUpdater.
I will check the APIs of them.

> Is ParallelUpdater fast enough in this training script?

It is not slow when it is compared with Caffe implementation. The biggest motivation of this implementation is to reduce the memory consumption per GPU (please refer Chainer-jp slack).

yuyu2172: > From my understanding, there is no concrete rule for multi-GPU implementations.

Yes. I would like to discuss if we are going to have any.

> In addition, this implementation is not against to that of classification (softmax cross entropy does not change it's behaviour).

Yeah, but BN needs to be modified if we want to make them the same.

Hakuyume: BN in ChainerMN keeps the behaviour, doesn't it?

yuyu2172: I think `MultiNodeBatchNormalization` is not common in ImageNet training protocols, such as the one in 1-hour paper.
So the behavior would change.

Hakuyume: I heard that PSPNet requires `MultiNodeBatchNormalization` for good performance. We cannot follow the protocol of ImageNet in that case. We don't need to treat `ImageNet in 1 hour` as a standard of all tasks/models.

Hakuyume: I think the best way is to follow the original papers as much as possible. Changing the behaviour is inconvenient for academic researches because we/users cannot estimate the amount of side-effect of modification.

examples/ssd/train_multi.py
```diff
@@ -0,0 +1,142 @@
...
+
+        return loss
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--model', choices=('ssd300', 'ssd512'), default='ssd300')
+    parser.add_argument('--batchsize', type=int, default=32)
+    parser.add_argument('--test_batchsize', type=int, default=32)
```
yuyu2172: How about making it `16`.

yuyu2172:
LGTM

PR #505, opened 2017-12-21
Add SiameseDataset

yuyu2172:


yuyu2172:
Updated

chainercv/datasets/siamese_dataset.py
```diff
@@ -0,0 +1,120 @@
+import numpy as np
+
+import chainer
+
+
+def _construct_label_to_key(labels):
+    d = dict()
+    for i, label in enumerate(labels):
+        if label not in d:
+            d[label] = list()
```
Hakuyume: How about using `collections.defaultdict`?

yuyu2172: Nice

docs/source/reference/datasets.rst
```diff
@@ -14,6 +14,10 @@ directory_parsing_label_names
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 .. autofunction:: directory_parsing_label_names
 
+SiameseDataset
+~~~~~~~~~~~~~~
+.. autofunction:: SiameseDataset
```
Hakuyume: `autoclass`?

yuyu2172: Thanks

chainercv/datasets/siamese_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Args:
+        dataset_0: The first base dataset.
+        dataset_1: The second base dataset.
+        pos_ratio (None or float): If this is not :obj:`None`,
+            this dataset tries to construct positive pairs at the
+            given rate. If :obj:`None`,
+            this dataset randomly samples examples from the base
+            datasets. The default value is :obj:`None`.
+        length (None): The length of this dataset. If :obj:`None`,
```
Hakuyume: `length (int)`
I think we can omit default `None` form the type description. i.e. `pos_ratio (None or float)` -> `pos_ratio (float)`

chainercv/datasets/siamese_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+            the length of the first base dataset is the length of this
+            dataset.
+        labels_0 (None or numpy.ndarray): The labels associated to
+            the first base dataset. The length should be the same as
+            the length of the first dataset. If this is :obj:`None`,
+            the labels are automatically fetched using the following
+            line of code: :obj:`[ex[1] for ex in dataset_0]`.
+            Also, if :obj:`pos_ratio` is :obj:`None`, this value
+            is ignored. The default value is :obj:`None`.
```
Hakuyume: How about describing the intention of this argument?
```
...
the labels are automatically fetched using the following line of code: :obj:`[ex[1] for ex in dataset_0]`.
If an array of labels already exists, users can skip this fetching iteration and reduce computation cost.
...
```

chainercv/datasets/siamese_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+        self._pos_ratio = pos_ratio
+        if length is None:
+            length = len(self._dataset_0)
+        self._length = length
+
+        if pos_ratio is not None:
+            if labels_0 is None:
+                labels_0 = [example[1] for example in dataset_0]
+            if labels_1 is None:
+                labels_1 = [example[1] for example in dataset_1]
```
Hakuyume: If `dataset_1 is dataset_0`, we can use `labels_0` as `labels1`. I think using same dataset for both `dataset_0` and `dataset_1` is very common and this optimization is effective.

chainercv/datasets/siamese_dataset.py
Hakuyume: Even if `labels_0` and `labels_1` are specified, `_construct_label_to_key` iterates these arrays. So, I think the effect of this argument is limited. Is this argument necessary?

yuyu2172: `_construct_label_to_key` would not iterate over the dataset, so no images would be loaded if `labels_*` are specified.

Hakuyume: I see.

chainercv/datasets/siamese_dataset.py
```diff
@@ -57,6 +61,8 @@ def __init__(self, dataset_0, dataset_1,
         self._length = length
 
         if pos_ratio is not None:
+            if labels_1 is None and  dataset_0 is dataset_1:
+                labels_1 = labels_0
             if labels_0 is None:
                 labels_0 = [example[1] for example in dataset_0]
             if labels_1 is None:
```
Hakuyume: In the case of `labels_0 is None and labels_1 is None`, we can use `[example[1] for example in dataset_0]` for `lablels_1`. Also, the computation of `self._label_to_index_1` can be skipped.

chainercv/datasets/siamese_dataset.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if pos_ratio is not None:
+            if labels_1 is None and dataset_0 is dataset_1:
+                labels_1 = labels_0
+            if labels_0 is None and dataset_0 is dataset_1:
+                labels_0 = labels_1
+            if labels_0 is None:
+                labels_0 = [example[1] for example in dataset_0]
+            if labels_1 is None:
+                labels_1 = [example[1] for example in dataset_1]
```
Hakuyume: In the case of `labels_0 is None`, `labels_1 is None` and `dataset_0 is dataset_1`,
```python
        if pos_ratio is not None:
            if labels_1 is None and dataset_0 is dataset_1:
                labels_1 = labels_0  # labels_1 = None
            if labels_0 is None and dataset_0 is dataset_1:
                labels_0 = labels_1  # labels_0 = None
            if labels_0 is None:
                labels_0 = [example[1] for example in dataset_0]  # fetch labels from dataset_0 <- OK
            if labels_1 is None:
                labels_1 = [example[1] for example in dataset_1]  # fetch labels from dataset_1 <- redundant
```

chainercv/datasets/siamese_dataset.py
```diff
@@ -0,0 +1,137 @@
...
+            labels_1 = np.array(labels_1)
+            if not (labels_0.dtype == np.int32 and labels_0.ndim == 1
+                    and len(labels_0) == len(dataset_0) and
+                    labels_1.dtype == np.int32 and labels_1.ndim == 1
+                    and len(labels_1) == len(dataset_1)):
+                raise ValueError('the labels are invalid.')
+
+            # Construct mapping label->idx
+            self._label_to_index_0 = _construct_label_to_key(labels_0)
+            self._label_to_index_1 = _construct_label_to_key(labels_1)
```
Hakuyume:
```python
if dataset_0 is dataset_1:
    self,_label_to_index_1 = self._label_to_index0
else:
    self._label_to_index_1 = _construct_label_to_key(labels_1)
```

chainercv/datasets/siamese_dataset.py
```diff
@@ -0,0 +1,137 @@
...
+                elif labels_1 is None:
+                    labels_1 = labels_0
+            else:
+                if labels_0 is None:
+                    labels_0 = [example[1] for example in dataset_0]
+                if labels_1 is None:
+                    labels_1 = [example[1] for example in dataset_1]
+
+            labels_0 = np.array(labels_0)
+            labels_1 = np.array(labels_1)
```
Hakuyume: The following code can reduce call of `np.array`.
```python
            if dataset_0 is dataset_1:
                if labels_0 is None and labels_1 is None:
                    labels_0 = np.array([example[1] for example in dataset_0])
                    labels_1 = labels_0
                elif labels_0 is None:
                    labels_0 = labels_1
                elif labels_1 is None:
                    labels_1 = labels_0
            else:
                if labels_0 is None:
                    labels_0 = np.array([example[1] for example in dataset_0])
                if labels_1 is None:
                    labels_1 = np.array([example[1] for example in dataset_1])
```

PR #508, opened 2017-12-25
calculate loc2bbox as float64 in proposal_creator

knorth55:
in order to avoid `overflow` error in `np.exp`, we can calculate with `np.float64` because it is on CPU.
in `np.float32`, maximum `x` of `np.exp(x)` is around `88`, and it might be exceeded.

### DIFF
--- a/chainercv/links/model/faster_rcnn/utils/proposal_creator.py
+++ b/chainercv/links/model/faster_rcnn/utils/proposal_creator.py
@@ -111,7 +111,8 @@ class ProposalCreator(object):
         anchor = cuda.to_cpu(anchor)
 
         # Convert anchors into proposal via bbox transformations.
-        roi = loc2bbox(anchor, loc)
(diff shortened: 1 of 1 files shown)


yuyu2172:
There is something wrong when `x > 88`. The network should have learned that boxes can never be that huge. Can you check if there is any problem in your code?

Also, even if the output of `loc2bbox` is infinitely large, it should be clipped to image size.
https://github.com/knorth55/chainercv/blob/f8d1a037269f01a2242bb892fdefd6718dd04394/chainercv/links/model/faster_rcnn/utils/proposal_creator.py#L121

```
>>> import numpy as np
>>> x = np.exp(np.float32(90))
>>> print(np.clip(x, 0, 10))
10
```

knorth55:
that's true.
thank you for your comment.
I think my model training was overfitted and loss got large extremely.

yuyu2172:
ok

PR #519, opened 2018-02-06
Dynamically import matplotlib.pyplot

yuyu2172:
Related https://github.com/chainer/chainer/pull/2740

`matplotlib` backend can now be configured after importing `chainercv`.

### DIFF
--- a/chainercv/extensions/vis_report/detection_vis_report.py
+++ b/chainercv/extensions/vis_report/detection_vis_report.py
@@ -7,7 +7,7 @@ import chainer
 from chainercv.visualizations.vis_bbox import vis_bbox
 
 try:
-    import matplotlib
+    import matplotlib  # NOQA
     _available = True
 
 except (ImportError, TypeError):
--- a/chainercv/extensions/vis_report/detection_vis_report.py
+++ b/chainercv/extensions/vis_report/detection_vis_report.py
@@ -7,10 +7,10 @@ import chainer
 from chainercv.visualizations.vis_bbox import vis_bbox
 
 try:
-    from matplotlib import pyplot as plot
+    import matplotlib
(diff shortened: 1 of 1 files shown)


knorth55:
@yuyu2172 How about also importing matplotlib in the position of `_check_available()`?

yuyu2172:
Why do we need to have a module `matplotlib.pyplot` as an attribute?

knorth55:
i thought the problem is importing the module multiple times.
In that case, i thought its better keeping the module as attribute and reusing it

chainercv/extensions/vis_report/detection_vis_report.py
```diff
@@ -7,10 +7,10 @@
 from chainercv.visualizations.vis_bbox import vis_bbox
 
 try:
-    from matplotlib import pyplot as plot
+    import matplotlib  # NOQA
     _available = True
 
-except ImportError:
+except (ImportError, TypeError):
```
Hakuyume: Are there any cases that `TypeError` is raised?

Hakuyume:
I don't think keeping a module as an attribute is good. @yuyu2172's solution looks good and enough.

chainercv/extensions/vis_report/detection_vis_report.py
yuyu2172: I copied it from Chainer.

`TypeError` is sometimes called when backend is not imported properly.

###### related

https://github.com/chainer/chainer/pull/2482

https://stackoverflow.com/questions/31328436/typeerror-constructor-returned-null-while-importing-pyplot-in-ssh

Hakuyume: I see.

PR #396, opened 2017-08-17
Improve style checker

Hakuyume:
New checks
- `[]` or `list()` => `list()`
- `{}` or `dict()` => `dict()`

PR #522, opened 2018-02-20
list()/dict() -> []/{}

Hakuyume:
In #396, I proposed using `list()/dict()`. However, this is against the rule of chainer/cupy (https://github.com/cupy/cupy/issues/700).


yuyu2172:
LGTM

PR #523, opened 2018-02-27
Improve apply_prediction

### LINKED ISSUE #468: Consider multiple inputs into model
Currently, the additional tools in chainercv expect that a model takes only one input (an image).
So tools such as `apply_prediction_to_iterator.py` are only designed to handle single input models. In some cases, a vision model will take multiple inputs (e.g. VQA or Bi-Linear models) and it would be nice to natively support that.

### LINKED ISSUE #516: apply_prediction_to_iterator is too lengthy
I prefer shorter name, but I have no good idea now (`apply_prediction`?).

Hakuyume:
fix #468, fix #516 
- renamed `imgs, pred_values, gt_values` -> `in_values, out_values, rest_values`
- `in_values` are the first `n_input` elements of the batch
- renamed `apply_prediction_to_iterator` -> `apply_to_iterator`

### DIFF
--- a/chainercv/utils/iterator/apply_to_batch.py
+++ b/chainercv/utils/iterator/apply_to_batch.py
@@ -1,32 +1,31 @@
 from chainercv.utils.iterator.unzip import unzip
 
 
-def apply_prediction_to_iterator(func, iterator, n_input=1, hook=None):
-    """Apply a prediction function/method to an iterator.
+def apply_to_batch(func, iterator, n_input=1, hook=None):
+    """Apply a function/method to an iterator of batches.
 
-    This function applies a prediction function/method to an iterator.
-    It assumes that the iterator returns a batch of images or
-    a batch of tuples whose first element is an image. In the case that
-    it returns a batch of tuples, the rests are treated as ground truth
-    values.
+    This function applies a function/method to an iterator of batches.
+    It assumes that the iterator returns a batch of input data or
+    a batch of tuples whose first elements ara input data.
 
-    >>> imgs = next(iterator)
-    >>> # imgs: [img]
-    or
     >>> batch = next(iterator)
-    >>> # batch: [(img, gt_val0, gt_val1)]
-
-    This function applys :func:`predict` to a batch of images and gets
-    predicted value(s). :func:`predict` should take a batch of images and
-    return a batch of prediction values
-    or a tuple of batches of prediction values.
-
-    >>> pred_vals0 = predict(imgs)
-    >>> # pred_vals0: [pred_val0]
+    >>> # batch: [in_val]
     or
-    >>> pred_vals0, pred_vals1 = predict(imgs)
-    >>> # pred_vals0: [pred_val0]
-    >>> # pred_vals1: [pred_val1]
+    >>> # batch: [(in_val0, in_val1, ...)]
+    or
+    >>> # batch: [(in_val0, in_val1, ..., rest_val0, rest_val1, ...)]
+
+    This function applies :func:`func` to batch(es) of input data and gets
+    computed value(s). :func:`func` should take a batch of data and
+    return a batch of computed values
+    or a tuple of batches of computed values.
+
+    >>> out_vals = func(in_val0, in_val1, ...)
+    >>> # out_vals: [out_val]
+    or
+    >>> out_vals0, out_vals1, ... = func(in_val0, in_val1)
+    >>> # out_vals0: [out_val0]
+    >>> # out_vals1: [out_val1]
 
     Here is an exmple, which applies a pretrained Faster R-CNN to
     PASCAL VOC dataset.
@@ -35,7 +34,7 @@ def apply_prediction_to_iterator(func, iterator, n_input=1, hook=None):
     >>>
     >>> from chainercv.datasets import VOCDetectionDataset
     >>> from chainercv.links import FasterRCNNVGG16
-    >>> from chainercv.utils import apply_prediction_to_iterator
+    >>> from chainercv.utils import apply_to_batch
     >>>
     >>> dataset = VOCDetectionDataset(year='2007', split='test')
     >>> # next(iterator) -> [(img, gt_bbox, gt_label)]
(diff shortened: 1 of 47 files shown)


chainercv/utils/iterator/apply_to_batch.py
```diff
@@ -0,0 +1,153 @@
...
+    This function applies a function/method to an iterator of batches.
+    It assumes that the iterator returns a batch of input data or
+    a batch of tuples whose first elements ara input data.
+
+    >>> batch = next(iterator)
+    >>> # batch: [in_val]
+    or
+    >>> # batch: [(in_val0, in_val1, ...)]
+    or
+    >>> # batch: [(in_val0, in_val1, ..., rest_val0, rest_val1, ...)]
```
yuyu2172: `n_input` should appear at this part of the documentation.

chainercv/utils/iterator/apply_to_batch.py
```diff
@@ -0,0 +1,153 @@
...
+
+    This function applies :func:`func` to batch(es) of input data and gets
+    computed value(s). :func:`func` should take a batch of data and
+    return a batch of computed values
+    or a tuple of batches of computed values.
+
+    >>> out_vals = func(in_val0, in_val1, ...)
+    >>> # out_vals: [out_val]
+    or
+    >>> out_vals0, out_vals1, ... = func(in_val0, in_val1)
```
yuyu2172: `in_val0, in_val1` --> `in_val0, in_val1, ...` for consistency

yuyu2172: I think these examples are not necessary.

chainercv/utils/iterator/apply_to_batch.py
```diff
@@ -0,0 +1,153 @@
+from chainercv.utils.iterator.unzip import unzip
+
+
+def apply_to_batch(func, iterator, n_input=1, hook=None):
+    """Apply a function/method to an iterator of batches.
+
+    This function applies a function/method to an iterator of batches.
+    It assumes that the iterator returns a batch of input data or
+    a batch of tuples whose first elements ara input data.
```
yuyu2172: I thought it was better to describe the internals of this function more in the beginning.
I wrote an alternative docstring. What do you think? 

---

This function assumes that the iterator iterates over a collection of tuples that contain inputs to `func`.
Additionally, the tuples may contain values that are not used by `func`.
For convenience, we allow the iterator to iterate over a collection of inputs that are not tuple.
t
Conceptually, this function does two operations.
First, it splits the iterator into two iterators. The first iterator contains the inputs to `func` and the second iterator contains the rest of the data from `iterator`.
Second, it creates another iterator that iterates over a collection of the outputs of the `func`.

Here is an illustration of the expected behavior of `iterator`.
`M` is the number of arguments for `func`, which is defined by an argument `n_input`.  `N` is the number of elements in a tuple from `iterator` that are not used by `func`. Note that `N` can be 0.

```
>>> batch = next(iterator)
>>> # batch:  [(in_val0, ..., in_val{M-1}, rest_val0, ..., rest_val{N-1})]
or
>>> # batch:  [in_val]
```

Hakuyume: > First, it splits the iterator into two iterators. The first iterator contains the inputs to func and the second iterator contains the rest of the data from iterator.
> Second, it creates another iterator that iterates over a collection of the outputs of the func.

This explanation sounds this function returns three iterators. However, it is not correct.

yuyu2172: Right. That is my mistake.

Hakuyume: OK, I will try to merge your solution into docstring.

chainercv/utils/iterator/apply_to_batch.py
Hakuyume: I would like to show the returned values should be `([out_val0], [out_val1], ...)` and not `[(out_val0, out_val1, ...)]`. I feel it is unclear without examples.

yuyu2172: Ah. I see. So you wanted to emphasize that the function returns tuple of lists.

BTW, shouldn't function take list of values?
So ` out_vals0, out_vals1, ... = func(in_val0, in_val1)` --> ` out_vals0, out_vals1, ... = func(in_vals0, in_vals1)`.

chainercv/utils/iterator/apply_to_iterator.py
```diff
@@ -0,0 +1,167 @@
...
+    >>> # out_vals0: [out_val0]
+    >>> # out_vals1: [out_val1]
+
+    With :func:`apply_to_iterator`, users can get iterator(s) of values
+    returned by :func:`func`. It also returns iterator(s) of input values and
+    values that are not used for computation.
+
+    >>> in_values, out_values, rest_values = apply_to_iterator(
+    >>>     func, iterator, n_input)
+    >>> # in_values: (iter of in_val0, ..., iter of in_val{n_input})
```
yuyu2172: n_input --> n_input-1

chainercv/utils/iterator/apply_to_batch.py
Hakuyume: You are right. That's my mistake.

chainercv/utils/iterator/apply_to_iterator.py
```diff
@@ -0,0 +1,167 @@
+from chainercv.utils.iterator.unzip import unzip
+
+
+def apply_to_iterator(func, iterator, n_input=1, hook=None):
+    """Apply a function/method to batches from an iterator.
+
+    This function applies a function/method to an iterator of batches.
```
yuyu2172: This function applies a function/method to batches from an iterator of type `chainer.Iterator`.

This makes clear that we are talking about Chainer's iterator.

chainercv/utils/iterator/apply_to_iterator.py
```diff
@@ -0,0 +1,167 @@
...
+
+    It assumes that the iterator iterates over a collection of tuples
+    that contain inputs to :func:`func`.
+    Additionally, the tuples may contain values that are not used by func.
```
yuyu2172: s/by func/by the funciton

chainercv/utils/iterator/apply_to_iterator.py
Hakuyume: This is not only for `chainer.Iterator`. It can work with other iterators.  `an iterator of type chainer.Iterator` sounds it requires some special features.

yuyu2172: OK. Then we should change args doc.
https://github.com/Hakuyume/chainercv/blob/0bd2f6d970aeaca436e6858b0e53661e364c1fd6/chainercv/utils/iterator/apply_to_iterator.py#L75

Hakuyume: Thank you for pointing out.

chainercv/utils/iterator/apply_to_iterator.py
```diff
@@ -0,0 +1,168 @@
...
+
+    This function applies a function/method to an iterator of batches.
+
+    It assumes that the iterator iterates over a collection of tuples
+    that contain inputs to :func:`func`.
+    Additionally, the tuples may contain values
+    that are not used by :func:`func`.
+    For convenience, we allow the iterator to iterate over a collection of
+    inputs that are not tuple.
+    Here is an illustration of the expected behavior of the iterator.
```
yuyu2172: I think this comment is helpful for Chainer users.

---

Here is an ....
In short, we require `iterator` to satisfy the interface of `chainer.Iterator`.

Hakuyume: `chainer.Iterator` requires many properties, such as `batch_size`, `epoch_detail`, etc..
`apply_to_iterator` does not require these properties.
https://docs.chainer.org/en/stable/reference/generated/chainer.iterators.SerialIterator.html

Hakuyume: How about `chainer.Iteratror satisfies this requirement` or  `chainer.Iteratror satisfies this behaviour`? These are correct.

Hakuyume: `This behaviour is same as chainer.Iterator` is also OK.

yuyu2172: This behaviour is the same as chainer.Iterator 

This one looks good! (“the” was missing)

Hakuyume:
@yuyu2172

PR #531, opened 2018-03-02
Add a simple example for SiameseDataset

yuyu2172:


chainercv/datasets/siamese_dataset.py
```diff
@@ -23,6 +23,21 @@ class SiameseDataset(chainer.dataset.DatasetMixin):
...
+
+        .. code::
+
+            >>> from chainer.datasets import get_mnist
+            >>> from chainercv.datasets import SiameseDataset
+            >>> mnist, _ = get_mnist()
+            >>> dataset = SiameseDataset(mnist, mnist, pos_ratio=0.3)
+            # The probability of the two samples having the same label
+            # is 0.3 as specified by pos_ratio.
+            >>> img_0, label_0, img_1, label_1 = dataset[0]
```
Hakuyume: It would be helpful to show that `SiameseDataset` returns examples randomly.
How about adding following lines?
```python
...
>>> img_0, label_0, img_1, label_1 = dataset[0]
Note that the returned examples may change in the next call because :class:`SiameseDataset` picks examples randomly.
>>> img_0_new, label_0_new, img_1_new, label_1_new = dataset[0]
In this case, :obj:`img_0_new` may differ :obj:`img_0`.
```

yuyu2172: Nice!

Hakuyume: Sorry, `differ` -> `differ from`.

PR #532, opened 2018-03-04
[WIP] SSD residual multibox

fukatani:
Multibox using residual box is used in ESSD (https://arxiv.org/pdf/1801.05918.pdf) and DSSD (https://arxiv.org/abs/1701.06659).
Only with replacing with Multibox in SSD, mAP can be increased from 76.4% to 77.1%.
I confirmed performance increasing in my local environment.

Anyone interested in supporting this feature?

update:
I found ESSD prediction module is little different (Simply added 1x1x512 layer only).


fukatani:
Sorry, I found more strong paper (https://github.com/sfzhang15/RefineDet) and, I think this PR value is not so high. I close this PR.

PR #533, opened 2018-03-07
Fix pretrained weight for ResNet

yuyu2172:
Some names of parameters are not updated during the code review. I am sorry about that.

### DIFF
--- a/chainercv/links/model/resnet/resnet.py
+++ b/chainercv/links/model/resnet/resnet.py
@@ -201,9 +201,9 @@ class ResNet(PickableSequentialChain):
             if not _available:
                 warnings.warn('cv2 is not installed on your environment. '
                               'The scores of ResNets reported in the '
-                              'READMEs in ChainerCV\'s examples are '
-                              'calculated using OpenCV as the backend. '
-                              'With Pillow as the '
+                              'README of the ChainerCV\'s classification '
+                              'example are calculated using OpenCV as the '
+                              'backend. With Pillow as the '
                               'backend, the scores would change.',
                               RuntimeWarning)
             path = download_model(_models[pretrained_model]['url'])
--- a/chainercv/links/model/resnet/resnet.py
+++ b/chainercv/links/model/resnet/resnet.py
@@ -110,8 +110,8 @@ class ResNet(PickableSequentialChain):
                 'imagenet': {
                     'n_class': 1000,
                     'url': 'https://github.com/yuyu2172/share-weights/'
-                    'releases/download/0.0.5/'
-                    'resnet50_imagenet_convert_2017_12_18.npz',
+                    'releases/download/0.0.6/'
+                    'resnet50_imagenet_convert_2018_03_07.npz',
                     'mean': _imagenet_mean
                 },
             },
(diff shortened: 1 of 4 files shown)


yuyu2172:
I am checking if `eval_imagenet.py` produces the same score as before.

yuyu2172:
@Hakuyume 
Please review this.

Hakuyume:
If the backend affects the performance, how about raising a warning?
In SSD, which is also affected by the backend, it raises a warning when user try to use the pretrained model without OpenCV.
https://github.com/chainer/chainercv/blob/master/chainercv/links/model/ssd/ssd_vgg16.py#L240-L245

yuyu2172:
Thanks

Hakuyume:
It is better to raise a warning from ResNet class and not from sample script. (please refer the example of SSD).

yuyu2172:
OK

yuyu2172:
@Hakuyume

chainercv/links/model/resnet/resnet.py
```diff
@@ -191,6 +198,13 @@ def __init__(self, n_layer,
             self.prob = F.softmax
 
         if pretrained_model in _models:
+            if not _available:
+                warnings.warn('cv2 is not installed on your environment. '
+                              'The scores of ResNets reported in the '
+                              'ChainerCV\'s Github page are calculated using '
```
Hakuyume: `Github` -> `GitHub`.

Hakuyume: `Github page` sounds that we have `***.github.io` page.

Hakuyume: Why did you change the warning message from that of SSD? It is better to use the same message because they say the same thing.

examples/classification/eval_imagenet.py
```diff
@@ -42,13 +42,13 @@ def main():
         extractor = VGG16(n_class, args.pretrained_model)
     elif args.model == 'resnet50':
         extractor = ResNet50(
-            n_class, args.pretrained_model, mode=args.resnet_mode)
+            n_class, args.pretrained_model, arch=args.resnet_mode)
```
Hakuyume: `resnet_mode` -> `resnet_arch`

chainercv/links/model/resnet/resnet.py
Hakuyume: I see. I understand the difference.

chainercv/links/model/resnet/resnet.py
```diff
@@ -201,8 +201,9 @@ def __init__(self, n_layer,
             if not _available:
                 warnings.warn('cv2 is not installed on your environment. '
                               'The scores of ResNets reported in the '
-                              'ChainerCV\'s Github page are calculated using '
-                              'OpenCV as the backend. With Pillow as the '
+                              'READMEs in ChainerCV\'s examples are '
```
Hakuyume: The scores are reported only in https://github.com/chainer/chainercv/tree/master/examples/classification. So `READMEs` -> `README`? Will you add the scores to another README?

PR #534, opened 2018-03-08
Fix the range of offset in random_crop

akitotakeki:


Hakuyume:
Thank you for PR. How about using `random.randint` (https://docs.python.org/3/library/random.html#random.randint)? It returns an integer from `[a, b]` (both boundaries are included).

And I find that `img.shape[1] == H` and `img.shape[1] > H` can be merged.
```python
    if img.shape[1] >= H:
         y_offset = random.randint(0, img.shape[1] - H)
    else:
         ...
```

akitotakeki:
@Hakuyume  Thanks for the review. I've fixed the points.

PR #535, opened 2018-03-12
Fix multi-thread hang for cv2

koki0702:
`cv2` sometimes hangs when used in sub-processes.
For example `chainercv.links.model.ssd.resize_with_random_interpolation()` does not work with `chainer.iterators.MultiprocessIterator`.

So I set single-thread using `cv2.setNumThreads(0)`.

### DIFF
--- a/chainercv/links/model/ssd/transforms.py
+++ b/chainercv/links/model/ssd/transforms.py
@@ -270,6 +270,9 @@ def resize_with_random_interpolation(img, size, return_param=False):
 
     import cv2
 
+    # cv2 sometimes hangs when used in sub-processes, so we set single-thread.
+    cv2.setNumThreads(0)
+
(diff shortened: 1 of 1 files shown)


furushchev:
I found the documentation is added now in the documentation for `chainer`.
https://github.com/chainer/chainer/pull/4589

But I would still like to clarify the workaround from the point of practical use.

yuyu2172:
> I mean the code is in chainercv library not in user-side and I think it is not easy to be modified even if there is documentation.

How about `Add cv2.setNumThreads(0) right after import cv2 in your training script`

https://docs.chainer.org/en/stable/tips.html#my-training-process-gets-stuck-when-using-multiprocessiterator

Also, this problem has been fixed in #572 .

Related: https://github.com/chainer/chainer/pull/4607

PR #537, opened 2018-03-19
Missing definition of plt

iphyer:
should add `import matplotlib.pyplot as plt` otherwise report `NameError: name 'plt' is not defined`


yuyu2172:
Hi. Thank your for you contribution.
Actually, we did not import it in purpose for brevity in purpose.

In the first code example, there is a comment saying that we omit them.

```
    # In the rest of the tutorial, we assume that the `plt`
    # is imported before every code snippet.
```

I think it is ok to leave the example as it is, so I would close this PR.

PR #540, opened 2018-03-20
Add VOC and SBD instance segmentation dataset

knorth55:
[EDIT] This PR is splitted into #540 and #541 

- [x] add `VOCInstanceSegmentationDataset`
- [x] add `SBDInstanceSegmenationDataset`
- [x] add `voc_instance_segmentation_label_names`
- [x] add `sbd_instance_segmentation_label_names`
- [x] add `assert_is_instance_segmentation_dataset`
- [x] add `test_voc_instance_segmentation_dataset.py`
- [x] add `test_sbd_instance_segmentation_dataset.py`
- [x] add `test_assert_is_instance_segmentation_dataset.py`

PR #541, opened 2018-03-21
Add vis instance segmentation func

knorth55:
Merge after #540 

- [x] add `vis_instance_segmentation`
- [x] add `test_vis_instance_segmentation_dataset.py`

```python
from chainercv.datasets import SBDInstanceSegmentationDataset
from chainercv.datasets import sbd_instance_segmentation_label_names
from chainercv.visualizations import vis_instance_segmentation

dataset = SBDInstanceSegmentationDataset()
img, bbox, mask, label = dataset[0]
vis_instance_segmentation(
    img, bbox, mask, label, label_names=sbd_instance_segmentation_label_names,
    alpha=0.7)
plt.show()
```
ins_dataset_example

### DIFF
--- a/chainercv/visualizations/vis_instance_segmentation.py
+++ b/chainercv/visualizations/vis_instance_segmentation.py
@@ -1,0 +1,97 @@
+from __future__ import division
+
+import numpy as np
+
+from chainercv.visualizations.vis_image import vis_image
+from chainercv.visualizations.vis_semantic_segmentation import _default_cmap
+
+
+def vis_instance_segmentation(
+        img, bbox, mask, label=None, score=None, label_names=None,
+        alpha=0.7, ax=None):
+    """Visualize instance segmentation.
+
+    Example:
+
+        >>> from chainercv.datasets import SBDInstanceSegmentationDataset
+        >>> from chainercv.datasets \
+        ...     import sbd_instance_segmentation_label_names
+        >>> from chainercv.visualizations import vis_instance_segmentation
+        >>> import matplotlib.pyplot as plot
+        >>> dataset = SBDSegmentationDataset()
+        >>> img, bbox, mask, label = dataset[0]
+        >>> vis_instance_segmentation(
+        ...     img, bbox, mask, label,
+        ...     label_names=sbd_instance_segmentation_label_names)
+        >>> plot.show()
+
+    Args:
+        img (~numpy.ndarray): An array of shape :math:`(3, H, W)`.
+            This is in RGB format and the range of its value is
+            :math:`[0, 255]`.
+        bbox (~numpy.ndarray): A float array of shape :math:`(R, 4)`.
+            :math:`R` is the number of objects in the image, and each
+            vector represents a bounding box of an object.
+            The bounding box is :math:`(y_min, x_min, y_max, x_max)`.
+        mask (~numpy.ndarray): A bool array of shape
+            :math`(R, H, W)`.
+            If there is an object, the value of the pixel is :obj:`True`,
+            and otherwise, it is :obj:`False`.
+        label (~numpy.ndarray): An integer array of shape :math:`(R, )`.
+            The values correspond to id for label names stored in
+            :obj:`label_names`.
+        label_names (iterable of strings): Name of labels ordered according
+            to label ids.
+        alpha (float): The value which determines transparency of the figure.
+            The range of this value is :math:`[0, 1]`. If this
+            value is :obj:`0`, the figure will be completely transparent.
+            The default value is :obj:`0.7`. This option is useful for
+            overlaying the label on the source image.
+        ax (matplotlib.axes.Axis): The visualization is displayed on this
+            axis. If this is :obj:`None` (default), a new axis is created.
+
+    Returns:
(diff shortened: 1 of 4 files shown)


chainercv/visualizations/vis_instance_segmentation.py
```diff
@@ -0,0 +1,98 @@
...
+        matploblib.axes.Axes: Returns :obj:`ax`.
+        :obj:`ax` is an :class:`matploblib.axes.Axes` with the plot.
+
+    """
+    if label is not None and not len(bbox) == len(label):
+        raise ValueError('The length of label must be same as that of bbox')
+    if score is not None and not len(bbox) == len(score):
+        raise ValueError('The length of score must be same as that of bbox')
+
+    n_ins_class = len(bbox)
```
yuyu2172: s/n_ins_class/n_inst

`n_ins_class` sounds like this is related to class categories.

chainercv/visualizations/vis_instance_segmentation.py
```diff
@@ -0,0 +1,98 @@
...
+    colors = [_default_cmap(l) for l in range(n_ins_class + 1)]
```
yuyu2172: How about `range(1, n_inst)`?
By doing so, you can remove one line.
Also, `np.array` is optional in this case, so we can omit it.

chainercv/visualizations/vis_instance_segmentation.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Returns:
+        matploblib.axes.Axes: Returns :obj:`ax`.
+        :obj:`ax` is an :class:`matploblib.axes.Axes` with the plot.
+
+    """
+    if label is not None and not len(bbox) == len(label):
+        raise ValueError('The length of label must be same as that of bbox')
+    if score is not None and not len(bbox) == len(score):
+        raise ValueError('The length of score must be same as that of bbox')
```
yuyu2172: How about checking `mask` as well?

chainercv/visualizations/vis_instance_segmentation.py
```diff
@@ -0,0 +1,98 @@
...
+
+    n_ins_class = len(bbox)
+    colors = [_default_cmap(l) for l in range(n_ins_class + 1)]
+    colors = np.array(colors)[1:]
+
+    # Returns newly instantiated matplotlib.axes.Axes object if ax is None
+    ax = vis_image(img, ax=ax)
+
+    vis_img = np.zeros((mask.shape[1], mask.shape[2], 4), dtype=np.uint8)
```
yuyu2172: `vis_img` has very similar name to `vis_image`.
How about `canvas`?

chainercv/visualizations/vis_instance_segmentation.py
```diff
@@ -0,0 +1,98 @@
...
+        ...     label_names=sbd_instance_segmentation_label_names,
+        ...     alpha=0.7)
+        >>> plot.show()
+
+    Args:
+        img (~numpy.ndarray): An array of shape :math:`(3, H, W)`.
+            This is in RGB format and the range of its value is
+            :math:`[0, 255]`.
+        bbox (~numpy.ndarray): A float array of shape :math:`(R, 4)`.
+            R is the number of objects in the image, and each vector
```
yuyu2172:
```
R --> `:math:`R``
```

chainercv/visualizations/vis_instance_segmentation.py
```diff
@@ -0,0 +1,98 @@
...
+        img, bbox, mask, label=None, score=None, label_names=None,
+        alpha=1, ax=None):
+    """Visualize instance segmentation.
+
+    Example:
+
+        >>> from chainercv.datasets import SBDInstanceSegmentationDataset
+        >>> from chainercv.datasets \
+        ...     import sbd_instance_segmentation_label_names
+        >>> from chainercv.visualizations import vis_image
```
yuyu2172: We do not need `vis_image`

yuyu2172:
Please rebase master

knorth55:
i rebased and updated

chainercv/visualizations/vis_instance_segmentation.py
```diff
@@ -0,0 +1,98 @@
...
+        mask (~numpy.ndarray): A bool array of shape
+            :math`(R, H, W)`.
+            If there is an object, the value of the pixel is :obj:`True`,
+            and otherwise, it is :obj:`False`.
+        label (~numpy.ndarray): An integer array of shape :math:`(R, )`.
+            The values correspond to id for label names stored in
+            :obj:`label_names`.
+        label_names (iterable of strings): Name of labels ordered according
+            to label ids.
+        alpha (float): The value which determines transparency of the figure.
```
yuyu2172: Is `alpha=1` a reasonable default value?
To me, `alpha=0.7` looks better.
Also, the doc needs to be changed if we are changing this.

knorth55: I prefer alpha=0.7, so I will change it

knorth55: I updated.

PR #542, opened 2018-03-25
fix doc

fukatani:


chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
```diff
@@ -31,7 +31,7 @@ class CityscapesSemanticSegmentationDataset(dataset.DatasetMixin):
             Cityscapes dataset.
         ignore_labels (bool): If :obj:`True`, the labels marked
             :obj:`ignoreInEval` defined in the original
-            `cityscapesScripts<https://github.com/mcordts/cityscapesScripts>_`
+            cityscapesScripts<https://github.com/mcordts/cityscapesScripts>
```
fukatani: I tried 
```
cityscapesScripts<https://github.com/mcordts/cityscapesScripts>`_
```
here, but it doesn't work.

yuyu2172:
```
    """Semantic segmentation dataset for `Cityscapes dataset`_.

    .. _`Cityscapes dataset`: https://www.cityscapes-dataset.com

    .. note::

        Please manually download the data because it is not allowed to
        re-distribute Cityscapes dataset.

    Args:
        data_dir (string): Path to the dataset directory. The directory should
            contain at least two directories, :obj:`leftImg8bit` and either
            :obj:`gtFine` or :obj:`gtCoarse`. If :obj:`auto` is given, it uses
            :obj:`$CHAINER_DATSET_ROOT/pfnet/chainercv/cityscapes` by default.
        label_resolution ({'fine', 'coarse'}): The resolution of the labels. It
            should be either :obj:`fine` or :obj:`coarse`.
        split ({'train', 'val'}): Select from dataset splits used in
            Cityscapes dataset.
        ignore_labels (bool): If :obj:`True`, the labels marked
            :obj:`ignoreInEval` defined in the original `cityscapesScripts`_
            will be replaced with :obj:`-1` in the :meth:`get_example` method.
            The default value is :obj:`True`.

    .. _`cityscapesScripts`: https://github.com/mcordts/cityscapesScripts

    """
```

How about this?
screenshot from 2018-03-27 16 07 59

yuyu2172:
Thank you for your contribution!

chainercv/datasets/cityscapes/cityscapes_semantic_segmentation_dataset.py
fukatani: Cool! I did it!

yuyu2172:
LGTM

PR #546, opened 2018-03-27
Update interface of vis_instance_segmentation and add color option

yuyu2172:
I changed `vis_instance_segmentation` to support custom `colors`.
This works like `vis_semantic_segmentation`.

Also, I added `show_bbox` option that visualizes bounding box together with masks.
This style of visualization is used by Facebook's detectron.

Finally, I changed `vis_instance_segmentation` from depending on `vis_semantic_segmentation`.

### DIFF
--- a/chainercv/visualizations/vis_instance_segmentation.py
+++ b/chainercv/visualizations/vis_instance_segmentation.py
@@ -2,12 +2,12 @@ from __future__ import division
 
 import numpy as np
 
-from chainercv.visualizations.vis_semantic_segmentation import _default_cmap
+from chainercv.visualizations.colormap import voc_colormap
 
 
 def vis_instance_segmentation(
         img, bbox, mask, label=None, score=None, label_names=None,
-        alpha=0.7, ax=None):
+        colors=None, alpha=0.7, show_bbox=False, ax=None):
     """Visualize instance segmentation.
 
     Example:
@@ -41,11 +41,18 @@ def vis_instance_segmentation(
             :obj:`label_names`.
         label_names (iterable of strings): Name of labels ordered according
             to label ids.
+        colors: (iterable of tuple): List of colors.
+            Each color is RGB format and the range of its values is
+            :math:`[0, 255]`. The :obj:`i`-th element is the color used
+            to visualize the :obj:`i`-th instance.
+            If :obj:`colors` is :obj:`None`, the default color map is used.
         alpha (float): The value which determines transparency of the figure.
             The range of this value is :math:`[0, 1]`. If this
             value is :obj:`0`, the figure will be completely transparent.
             The default value is :obj:`0.7`. This option is useful for
             overlaying the label on the source image.
+        show_bbox (bool): Decide whether to show bounding boxes.
+            The default value is :obj:`False`.
         ax (matplotlib.axes.Axis): The visualization is displayed on this
             axis. If this is :obj:`None` (default), a new axis is created.
 
@@ -67,14 +74,26 @@ def vis_instance_segmentation(
         raise ValueError('The length of score must be same as that of bbox')
 
     n_inst = len(bbox)
-    colors = np.array([_default_cmap(l) for l in range(1, n_inst + 1)])
+    if colors is None:
+        colors = [voc_colormap(l) for l in range(1, n_inst + 1)]
+    colors = np.array(colors)
 
     canvas_img = img.transpose((1, 2, 0)).copy()
-    for i, (color, bb, msk) in enumerate(zip(colors, bbox, mask)):
+    for i, (bb, msk) in enumerate(zip(bbox, mask)):
+        # The length of `colors` can be smaller than the number of instances
+        # if a non-default `colors` is used.
+        color = colors[i % len(colors)]
         bb = np.round(bb).astype(np.int32)
         y_min, x_min, y_max, x_max = bb
         if y_max > y_min and x_max > x_min:
             canvas_img[msk] = alpha * color + canvas_img[msk] * (1 - alpha)
+
+            if show_bbox:
+                ax.add_patch(
+                    plot.Rectangle(
+                        (x_min, y_min), x_max - x_min, y_max - y_min,
+                        fill=False, edgecolor=color / 255, linewidth=0.5,
+                        alpha=alpha))
 
         caption = []
         if label is not None and label_names is not None:
--- a/chainercv/visualizations/vis_instance_segmentation.py
+++ b/chainercv/visualizations/vis_instance_segmentation.py
@@ -78,15 +78,16 @@ def vis_instance_segmentation(
     colors = np.array(colors)
 
     _, H, W = mask.shape
-    canvas_img = np.zeros((H, W, 3), dtype=np.uint8)
+    canvas_img = np.zeros((H, W, 4), dtype=np.uint8)
     for i, (bb, msk) in enumerate(zip(bbox, mask)):
         # The length of `colors` can be smaller than the number of instances
         # if a non-default `colors` is used.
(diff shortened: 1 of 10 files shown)


Hakuyume:
Is `bbox` required when `show_bbox=False`?

knorth55:
`bbox` is used to determine where to print object label in the image.
But we can remove it by calculating the label location from object mask region.

yuyu2172:
@Hakuyume 
It is certainly not necessary to require `bbox` when `show_bbox=False`.
Should I make `bbox` optional by changing default values of arguments?

```
def vis_instance_segmentation(img, bbox=None, mask=None, label=None, ...)
```

> But we can remove it by calculating the label location from object mask region.

It feels wierd to change the interface of `vis_instance_segmentation` and `InstanceSegmentationDataset`.

Hakuyume:
> Should I make bbox optional by changing default values of arguments?

I think it is better.
If `bbox` is specified, calculate the location of label by `bbox` (this is the current behaviour).
Otherwise, calculate it by `mask` as @knorth55 suggested.
> But we can remove it by calculating the label location from object mask region.

> It feels wierd to change the interface of vis_instance_segmentation and InstanceSegmentationDataset.

From my understanding, it does not change the interface of `InstanceSegmentationDataset`.

yuyu2172:
> From my understanding, it does not change the interface of InstanceSegmentationDataset.

What I meant is to change `vis_instance_segmentation` to have the following interface, which is different from the one for `InstanceSegmentationDataset`. I think that this is not good.

```
def vis_instance_segmentation(img, mask=None, label=None, ...)
```

Anyway, as you mentioned, we can `show_bbox` even when `bbox` is not specified.

Aside from that, I start to feel that the current order of variables is not optimal.
I think `mask, bbox, label, score` is better. This is because `mask` contains strictly more information than `bbox` (in most cases), and more informative data should be placed first.
Any thoughts on this point? (@knorth55  @Hakuyume )

Hakuyume:
> I think mask, bbox, label, score is better. This is because mask contains strictly more information than bbox (in most cases), and more informative data should be placed first.

How about `mask`, `label`, `score (optional)`, `bbox (optional)`?
`label` is always required and `score` is always included in the prediction.

knorth55:
@yuyu2172 wants to standardize interface in this order `img, mask, bbox, label`.
But for visualization, `bbox` can be optional, so it can be `img, mask, label, bbox=None, score=None`.
I prefer this interface.

yuyu2172:
> I think bbox and show_bbox are confusing. The API shown above is simple (bbox argument determines whether to show the bbox or not).

I agree. Thanks.

> How about mask, label, score (optional), bbox (optional)?

I like this too. `bbox` is the least informative of four. 

> But for visualization, bbox can be optional, so it can be img, mask, label, bbox=None, score=None.
I prefer this interface.

For the reason above, I prefer `mask, label, score (optional), bbox (optional)`.

We are going to change the interface of `InstanceSegmentationDataset` (and other instance segmentation stuffs) if we are going to adapt this transition.

knorth55:
> We are going to change the interface of InstanceSegmentationDataset (and other instance segmentation stuffs) if we are going to adapt this transition.

I think we dont need to modify dataset interface.

yuyu2172:
Why?

knorth55:
why do we need to modify dataset interface?
we can use `img, mask, bbox, label` as dataset interface, and `img, mask, label` as visualization interface.
They can be different.

yuyu2172:
Why should we make them different when we can keep them consistent?

knorth55:
I think this discussion is not like that.
This discussion is that we can keep different interface.

yuyu2172:
I don’t understand why we should have different interfaces. You didn’t give a reason.
It becomes easier for users when the interfaces are consistent.

knorth55:
Dataset inteface is `img, mask, bbox, label`
visualization interface follow the dataset order, but `bbox` can be optional so that it moves to behind.
By the way, how about `img, mask, bbox=None, label=None, score=None` for visualization interface?
 `label` can be optional and it can follow dataset order.

yuyu2172:
> Dataset inteface is img, mask, bbox, label

No. That is wrong. The current interface returns `img, bbox, mask, label`.
https://github.com/chainer/chainercv/blob/master/chainercv/utils/testing/assertions/assert_is_instance_segmentation_dataset.py#L44

> label can be optional

True. Nice point.

> By the way, how about img, mask, bbox=None, label=None, score=None for visualization interface?
label can be optional and it can follow dataset order.

Yes. This order is not bad.  I will summarise three options.

+ `img, mask, bbox, label, score` for visualizers, `img, mask, bbox, label` for datasets https://github.com/chainer/chainercv/pull/546#issuecomment-377269762 https://github.com/chainer/chainercv/pull/546#issuecomment-377670783
+ `img, mask, label, score, bbox` for visualizers, `img, mask, label, bbox` for datasets https://github.com/chainer/chainercv/pull/546#issuecomment-377366859
+ This is a new proposal. `img, mask, label, bbox, score` for visualizers, `img, mask, label, bbox` for datasets

yuyu2172:
I totally forgot that `label` is optional. Taking that into consideration, the first option looks okay.
What do you think? @Hakuyume

knorth55:
oh, i misunderstood the dataset interface.
I'm for the first idea.

Hakuyume:
> label can be optional

I can't follow this point. When does `label` become `None`?

Anyway, I prefer the 2nd API in https://github.com/chainer/chainercv/pull/546#issuecomment-377687397.
I think the importance of `bbox` is low because `mask` contains similar information.

knorth55:
`bbox` can be optional as same as https://github.com/chainer/chainercv/blob/master/chainercv/visualizations/vis_bbox.py#L4

Hakuyume:
> label can be optional as same as https://github.com/chainer/chainercv/blob/master/chainercv/visualizations/vis_bbox.py#L4

I see.

yuyu2172:
Docstring does not make sense until merging #555  .
Please merge if you find it OK  @knorth55

chainercv/visualizations/vis_instance_segmentation.py
```diff
@@ -67,10 +74,15 @@ def vis_instance_segmentation(
         raise ValueError('The length of score must be same as that of bbox')
 
     n_inst = len(bbox)
-    colors = np.array([_default_cmap(l) for l in range(1, n_inst + 1)])
+    if colors is None:
+        colors = [voc_colormap(l) for l in range(1, n_inst + 1)]
+    colors = np.array(colors)
 
     canvas_img = img.transpose((1, 2, 0)).copy()
```
knorth55: For visualization, I think it is better to overlap instance segmentation images as `vis semantic segmentation`.
`canvas_img` makes this function less useful.

yuyu2172: Sorry. I do not follow why `vis semantic segmentation` comes up?
Can you elaborate on this point more?

knorth55: I mean how about reverting https://github.com/chainer/chainercv/commit/f7a41e4ddaa12a5cb19d0458431e88a16edd10b6?
the visualization function can be overlapped together with `img=None`.

knorth55: I mean use `vis_image` in `vis_instance_segmentation`, and visualize `canvas_img` on the image.
It will be useful when we support `img=None`.

yuyu2172: I will support `img=None` in this PR.

knorth55: Great! thank you :)

PR #548, opened 2018-03-27
Fix ResNets when initialW is specified

yuyu2172:
The program failed when `initialW is not None` because `conv_initialW` is unspecified.

### DIFF
--- a/chainercv/links/model/resnet/resnet.py
+++ b/chainercv/links/model/resnet/resnet.py
@@ -89,10 +89,10 @@ class ResNet(PickableSequentialChain):
             the mean value used to train the pretrained model is used.
             Otherwise, the mean value calculated from ILSVRC 2012 dataset
             is used.
-        conv_initialW (callable): Initializer for the weights of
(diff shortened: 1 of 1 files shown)


yuyu2172:
@Hakuyume
Can you check this?
This depends on other PRs.

Hakuyume:
Don't we need `fc_initial_bias`?

> This depends on other PRs.

Which PRs? I wiil check them first.

Hakuyume:
Perhaps, we can use `initialW`, `fc_initialW` and `fc_initial_bias`. Since most of layers in ResNet are `Convolution2D`, users can understand `initialW` means `initialW` for conv.

Another solution is adding `fc_kwargs` that is a `dict` including `initialW` and `initial_bias`.
This is consistent with `bn_kwargs` that we use in many other parts.

yuyu2172:
> Another solution is adding fc_kwargs that is a dict including initialW and initial_bias.
This is consistent with bn_kwargs that we use in many other parts.

Nice.

PR #550, opened 2018-03-27
Fix build scripts not to include binary files and .pyx files

### LINKED ISSUE #536: The source package includes Linux binary files
[`chainercv-0.8.0.tar.gz`](https://pypi.python.org/packages/6d/c1/ff40e5642f37fe15d1d987d3fa5834ac8f0630b1540ad74d4dac05708273/chainercv-0.8.0.tar.gz#md5=7ca069e95359829f0e1be7ffa72ebd8c) contains two Linux binary files:

```
$ find chainercv-0.8.0 -name '*.so'
chainercv-0.8.0/chainercv/utils/bbox/_nms_gpu_post.cpython-36m-x86_64-linux-gnu.so
chainercv-0.8.0/chainercv/utils/bbox/_nms_gpu_post.so
```

`chainercv-0.8.0.tar.gz` is a source package, which should not contain platform dependent files. When distributing compiled extensions, one has to distribute wheel files.

gwtnb:
Fix #536.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+recursive-include chainercv *.c
+recursive-exclude chainercv *.pyx
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,0 @@
-include README.md
-recursive-include chainercv *
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ for name, data in ext_data.items():
     extensions.append(
         Extension('chainercv.{}'.format(name),
                   sources=sources)
-        )
+    )
 
 
 class CheckingBuildExt(_build_ext):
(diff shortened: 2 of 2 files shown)


gwtnb:
The cause of #536 is `recursive-include chainercv *`. With this statement, `python setup.py sdist` may add a binary file to the source package if there is the binary file in the source tree.

Of course, it is preferable to build a source package from a clean source tree. This PR is intended to make the build scripts fail-safe.

gwtnb:
Update: Modified `setup.py` to distribute `.c` files with `python setup.py sdist`. 

Ref: https://github.com/cupy/cupy/pull/920/

yuyu2172:
Thank you.
I confirmed that from the clean source tree, I can build a tar file containing `*.c` file. 

Could you fix the travis error?
The error is related to the order of import.

gwtnb:
Fixed.

yuyu2172:
LGTM

PR #552, opened 2018-04-03
add mask_iou

knorth55:
add `mask_iou` function in `utils`

### DIFF
--- a/chainercv/utils/mask/mask_iou.py
+++ b/chainercv/utils/mask/mask_iou.py
@@ -1,0 +1,46 @@
+from __future__ import division
+
+
+from chainer import cuda
+
+
+def mask_iou(mask_a, mask_b):
+    """Calculate the Intersection of Unions (IoUs) between masks.
+
+    IoU is calculated as a ratio of area of the intersection
+    and area of the union.
+
+    This function accepts both :obj:`numpy.ndarray` and :obj:`cupy.ndarray` as
+    inputs. Please note that both :obj:`mask_a` and :obj:`mask_b` need to be
+    same type.
+    The output is same type as the type of the inputs.
+
+    Args:
+        mask_a (array): An array whose shape is :math:`(N, H, W)`.
+            :math:`N` is the number of masks.
+            The dtype should be :obj:`numpy.bool`.
+        mask_b (array): An array similar to :obj:`mask_a`,
+            whose shape is :math:`(K, H, W)`.
+            The dtype should be :obj:`numpy.bool`.
+
+    Returns:
+        array:
+        An array whose shape is :math:`(N, K)`. \
+        An element at index :math:`(n, k)` contains IoUs between \
+        :math:`n` th mask in :obj:`mask_a` and :math:`k` th mask \
+        in :obj:`mask_b`.
+
+    """
+    if mask_a.shape[1:] != mask_b.shape[1:]:
+        raise IndexError
+    xp = cuda.get_array_module(mask_a)
+
+    n_mask_a = len(mask_a)
+    n_mask_b = len(mask_b)
+    iou = xp.empty((n_mask_a, n_mask_b), dtype=xp.float32)
+    for n, m_a in enumerate(mask_a):
+        for k, m_b in enumerate(mask_b):
+            intersect = xp.bitwise_and(m_a, m_b).sum()
+            union = xp.bitwise_or(m_a, m_b).sum()
+            iou[n, k] = intersect / union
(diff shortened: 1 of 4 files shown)


tests/utils_tests/mask_tests/test_mask_iou.py
```diff
@@ -0,0 +1,89 @@
...
+         dtype=np.bool),
+     'expected': np.array(
+         [[1., 0.], [0., 1.], [1 / 4, 2 / 3], [2 / 3, 1 / 4]])
+     },
+    {'mask_a': np.zeros((0, 2, 2), dtype=np.bool),
+     'mask_b': np.array([[[False, False], [False, False]]], dtype=np.bool),
+     'expected': np.zeros((0, 1))
+     },
+)
+class TestmaskIou(unittest.TestCase):
```
Hakuyume: `TestMaskIou`

tests/utils_tests/mask_tests/test_mask_iou.py
```diff
@@ -0,0 +1,89 @@
...
+     'mask_b': np.array([[[False, False], [True, True]]], dtype=np.bool)
+     },
+    {'mask_a': np.array([[[False, False], [True, True]]], dtype=np.bool),
+     'mask_b': np.array([[[False], [True, True]]], dtype=np.bool)
+     },
+    {'mask_a': np.array([[[False, False], [True, True]]], dtype=np.bool),
+     'mask_b': np.array([[[False, False, True], [True, True]]], dtype=np.bool)
+     },
+)
+class TestmaskIouInvalidShape(unittest.TestCase):
```
Hakuyume: `TestMaskIou~`

knorth55:
I add doc, thx.

chainercv/utils/mask/mask_iou.py
```diff
@@ -0,0 +1,46 @@
...
+        in :obj:`mask_b`.
+
+    """
+    if mask_a.shape[1:] != mask_b.shape[1:]:
+        raise IndexError
+    xp = cuda.get_array_module(mask_a)
+
+    n_mask_a = len(mask_a)
+    n_mask_b = len(mask_b)
+    iou = xp.zeros((n_mask_a, n_mask_b), dtype=xp.float32)
```
Hakuyume: How about using `xp.empty`? Since all elements are updated in the following loop, `iou` does not need to be initialized here.

chainercv/utils/mask/mask_iou.py
```diff
@@ -0,0 +1,46 @@
...
+    for i, m_a in enumerate(mask_a):
+        for j, m_b in enumerate(mask_b):
```
Hakuyume: Using `n` and `k` is consistent with the docs.

chainercv/utils/mask/mask_iou.py
```diff
@@ -0,0 +1,46 @@
...
+
+    """
+    if mask_a.shape[1:] != mask_b.shape[1:]:
+        raise IndexError
+    xp = cuda.get_array_module(mask_a)
+
+    n_mask_a = len(mask_a)
+    n_mask_b = len(mask_b)
+    iou = xp.zeros((n_mask_a, n_mask_b), dtype=xp.float32)
+    for i, m_a in enumerate(mask_a):
```
Hakuyume: We can remove these loops by using `numpy tricks`.

```python
i = xp.bitwise_and(mask_a[:, None], mask_b).sum(axis=(2,3))
u = xp.bitwise_or(mask_a[:, None], mask_b).sum(axis=(2,3))
return i / u
```

How do you think of it?

Hakuyume: A drawback is that it requires an array `(N, K, H, W)`, where the original code requires only `(H, W)`.

knorth55: `(N, K, H, W)` requires too much memory, so I write the original one.

chainercv/utils/mask/mask_iou.py
knorth55: thx, i will updated.

chainercv/utils/mask/mask_iou.py
knorth55: thx, i will updated.

chainercv/utils/mask/mask_iou.py
Hakuyume: I see.

yuyu2172:
LGTM

PR #558, opened 2018-04-08
Add random_erasing

akitotakeki:
This PR is the implementation of Random Erasing (https://arxiv.org/abs/1708.04896).
Random Erasing is a new data augmentation method for training the convolutional neural network (CNN).

### DIFF
--- a/chainercv/transforms/image/random_erasing.py
+++ b/chainercv/transforms/image/random_erasing.py
@@ -1,0 +1,126 @@
+from __future__ import division
+
+import math
+import numpy as np
+import random
+
+
+def random_erasing(img, prob=0.5,
+                   scale_ratio_range=(0.02, 0.4),
+                   aspect_ratio_range=(0.3, 1 / 0.3),
+                   mean=[0.4914, 0.4822, 0.4465],
+                   return_param=False, copy=False):
+    """Select a rectangle region in an image and erase its pixels with
+    ImageNet mean values.
+
+    The size :math:`(H_{erase}, W_{erase})` and the left top coordinate
+    :math:`(y_{start}, x_{start})` of the region are calculated as follows:
+
+    + :math:`H_{erase} = \\lfloor{\\sqrt{s \\times H \\times W \
+        \\times a}}\\rfloor`
+    + :math:`W_{erase} = \\lfloor{\\sqrt{s \\times H \\times W \
+        \\div a}}\\rfloor`
+    + :math:`y_{start} \\sim Uniform\\{0, H - H_{erase}\\}`
+    + :math:`x_{start} \\sim Uniform\\{0, W - W_{erase}\\}`
+    + :math:`s \\sim Uniform(s_1, s_2)`
+    + :math:`b \\sim Uniform(a_1, a_2)` and \
+        :math:`a = b` or :math:`a = \\frac{1}{b}` in 50/50 probability.
+
+    Here, :math:`s_1, s_2` are the two floats in
+    :obj:`scale_ratio_interval` and :math:`a_1, a_2` are the two floats
+    in :obj:`aspect_ratio_interval`.
+    Also, :math:`H` and :math:`W` are the height and the width of the image.
+    Note that :math:`s \\approx \\frac{H_{erase} \\times
+    W_{erase}}{H \\times W}` and
+    :math:`a \\approx \\frac{H_{erase}}{W_{erase}}`.
+    The approximations come from flooring floats to integers.
+
+    .. note::
+
+        When it fails to sample a valid scale and aspect ratio for a hundred
+        times, it picks values in a non-uniform way.
+        If this happens, the selected scale ratio can be smaller
+        than :obj:`scale_ratio_interval[0]`.
+
+    Args:
+        img (~numpy.ndarray): An image array. This is in CHW format.
+        prob (float): Erasing probability.
+        scale_ratio_interval (tuple of two floats): Determines
+            the distribution from which a scale ratio is sampled.
+        aspect_ratio_interval (tuple of two floats): Determines
+            the distribution from which an aspect ratio is sampled.
+        return_param (bool): Returns parameters if :obj:`True`.
+
+    Returns:
+        ~numpy.ndarray or (~numpy.ndarray, dict):
+
+        If :obj:`return_param = False`,
+        returns only the cropped image.
+
+        If :obj:`return_param = True`,
+        returns a tuple of erased image and :obj:`param`.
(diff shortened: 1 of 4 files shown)


chainercv/transforms/image/random_erasing.py
```diff
@@ -0,0 +1,101 @@
...
+
+        When it fails to sample a valid scale and aspect ratio for a hundred
+        times, it picks values in a non-uniform way.
+        If this happens, the selected scale ratio can be smaller
+        than :obj:`scale_ratio_interval[0]`.
+
+    Args:
+        img (~numpy.ndarray): An image array. This is in CHW format.
+        prob (float): Erasing probability.
+        scale_ratio_interval (tuple of two floats): Determines
```
Hakuyume: In the arguments, `scale_ratio_range` is used.

chainercv/transforms/image/random_erasing.py
```diff
@@ -0,0 +1,101 @@
...
+            the distribution from which a scale ratio is sampled.
+        aspect_ratio_interval (tuple of two floats): Determines
```
Hakuyume: In the arguments, `aspect_ratio_range` is used.

chainercv/transforms/image/random_erasing.py
```diff
@@ -0,0 +1,101 @@
...
+            the distribution from which an aspect ratio is sampled.
```
Hakuyume: `mean` is not described.

chainercv/transforms/image/random_erasing.py
```diff
@@ -0,0 +1,103 @@
...
+        of the value.
+
+        * **y_slice** (*slice*): A slice used to erase a region in the input
+        image. The relation below holds together with :obj:`x_slice`.
+        * **x_slice** (*slice*): Similar to :obj:`y_slice`.
+        * **scale_ratio** (float): :math:`s` in the description (see above).
+        * **aspect_ratio** (float): :math:`a` in the description.
+
+    """
+    if random.randint(0, 1) > prob:
```
Hakuyume: `randint(0, 1)` returns `0` or `1`. When `0 < prob < 1`, this condition means 50% regardless of `prob`.

Hakuyume:
I think the following code is more simple.
```python
if np.random.unifrom() < prob:
    random_sized_crop(img, ...)[:] = mean
```

chainercv/transforms/image/random_erasing.py
```diff
@@ -0,0 +1,105 @@
...
+        of the value.
+
+        * **y_slice** (*slice*): A slice used to erase a region in the input
+        image. The relation below holds together with :obj:`x_slice`.
+        * **x_slice** (*slice*): Similar to :obj:`y_slice`.
+        * **scale_ratio** (float): :math:`s` in the description (see above).
+        * **aspect_ratio** (float): :math:`a` in the description.
+
+    """
+    if random.uniform(0.0, 1.0) > prob:
```
Hakuyume: `< prob`

chainercv/transforms/image/random_erasing.py
```diff
@@ -0,0 +1,105 @@
...
+        crop, params = random_sized_crop(img, scale_ratio_range,
+                                         aspect_ratio_range, return_param=True)
+        if random_value:
+            crop[:] = np.random.random(crop.shape) * scale
```
Hakuyume: This line modifies the input image.

chainercv/transforms/image/random_erasing.py
```diff
@@ -0,0 +1,107 @@
...
+    Args:
+        img (~numpy.ndarray): An image array. This is in CHW format.
+        prob (float): Erasing probability.
+        scale_ratio_range (tuple of two floats): Determines
+            the distribution from which a scale ratio is sampled.
+        aspect_ratio_range (tuple of two floats): Determines
+            the distribution from which an aspect ratio is sampled.
+        random_value (bool): Fill the rectangle region with random values.
+        scale (float): Pixel value scale.
+        fixed_value (~numpy.ndarray): Determines pixel values
```
Hakuyume: For the consistency with other functions, how about using `fill`?

chainercv/transforms/image/random_erasing.py
```diff
@@ -0,0 +1,107 @@
...
+import random
+
+from chainercv.transforms.image.random_sized_crop import random_sized_crop
+
+
+def random_erasing(img, prob=0.5,
+                   scale_ratio_range=(0.02, 0.4),
+                   aspect_ratio_range=(0.3, 1 / 0.3),
+                   random_value=True,
+                   scale=1.0,
```
Hakuyume: This should be 255 since we use `[0, 255]` as the default range of images.

Hakuyume: I think we can remove this option.

chainercv/transforms/image/random_erasing.py
Hakuyume: How about using random value when `fill='random'`? We can remove `scale` and `random_value`.

akitotakeki: Changing from `fixed_value' to 'fill' is fine.
However, if 'scale' is deleted, the range of random values can not be defined.

Hakuyume: Do we need to specify the range of random value? The range of pixel is always `[0, 255]` in ChainerCV

akitotakeki: I understood the situation. I'll fix the point.
However, the pixel range is always `[0, 255]` in ChainerCV, but in Chainer, it seems to be `[0, 1]`.
https://github.com/chainer/chainer/blob/master/chainer/datasets/mnist.py
https://github.com/chainer/chainer/blob/master/chainer/datasets/svhn.py
https://github.com/chainer/chainer/blob/master/chainer/datasets/fashion_mnist.py

The difference seems inconvenient when using Chainer and ChainerCV at the same time.

Hakuyume: > in Chainer, it seems to be [0, 1].

This is not true. Chainer sometimes uses `[0, 255]`. For example, `ResNet` in Chainer assumes the input range is `[0, 255]`.
https://github.com/chainer/chainer/blob/master/chainer/links/model/vision/resnet.py#L547-L548

akitotakeki: Mismatch of the pixel range in Chainer seems awkward.
However, this has nothing to do with ChainerCV, so I will follow your suggestion.

PR #562, opened 2018-04-10
Make Conv2DActiv support comm kwarg without ChainerMN

yuyu2172:
An error is raised when `bn_kwargs` includes `comm` when ChainerMN is not installed.


yuyu2172:
This feature should not be supported.

PR #566, opened 2018-04-11
fix test_assert_is_point

Hakuyume:


yuyu2172:
I prefer to allow mask to be all false and have arbitrary coordinate values.

Hakuyume:
If all elements of `mask` are false, the last condition become valid.

PR #560, opened 2018-04-09
Support using vis_bbox together with vis_instance_segmentation  

yuyu2172:
figure_1

PR #493, opened 2017-12-07
Add attr.disk

Hakuyume:
Add a new PyTest's decorator `attr.disk` to mark tests that consume a huge disk space.

Issue #159, opened 2017-05-09
Keyword consistency


Hakuyume:
- [ ] `n_noun` and `n_nouns`
    For the number of something, we use `n_noun`. For example, `n_class` and `n_bbox` (not `n_classes` and `n_bboxes`).
- [ ] `label` and `class`
    We use `label` for label/class information. Only for the number of class, we use `n_class`.
- [ ] `bbox` and `bboxes`
    We use `bbox` for an array of bounding boxes. `bboxes` for a list/set of arrays of bounding boxes. In the documents, we use `bounding box`.
- [ ] `pred_noun` and `gt_noun`
   In functions which require both prediction and ground truth, we mark them by adding prefixes, `pred_` and `gt_`. For  example, `pred_bboxes` and `gt_bboxes`.

Hakuyume:
I think it is better to unify keywords. I guessed current rules from the codes.

Hakuyume:
@yuyu2172 How do think of it? Do you know other implicit rules?

yuyu2172:
Thanks for summarizing this.
I have no complaint for everything, but the second one. I think we were discussing on the usage of `cls` inside `eval_detection`. I think this is under discussion.

Other than that, here are some that I came up.

- [ ] For a name of a function or a class, `image` is used. As a name of a variable inside a class, `img` is used.
- [ ] In extensions, models are named as `target`.

Hakuyume:
@yuyu2172 thank you. I merged the rules you suggested and change the state of usage of `class` to "under discussion".

yuyu2172:
Copied from #163, these are rules for bounding boxes.

+ `bb` is a `(R,)`
+ `bbox` is a `(R, 4)`
+ `bboxes` is a `(B, R, 4)` or list of `(R_i, 4)  i=1, ..., B`.
+ `rois` is a `(R', 4)` which consists of bounding boxes for multiple images. Assuming that there are B images each containing `R_i` bounding boxes, `R' = \sum R_i`. `rois` comes together with a `(R',)` array called `batch_indices`, which contains batch indices of images to which bounding boxes correspond to.


For images, a similar rule applies like below.

+ `img` is a `(C, H, W)`.
+ `imgs` is a `(B, C, H, W)` or a list of `(C, H, W)`.

Hakuyume:
@yuyu2172 I merged them.

yuyu2172:
I think there is another rule.

[...]

PR #568, opened 2018-04-13
add FCIS ResNet101

knorth55:
Merge after #560 

- add `mask_voting`
- add `FCIS`
- add `FCISResNet101`
- add tests
- add docs

```bash
cd chainercv/examples/fcis
wget https://raw.githubusercontent.com/knorth55/chainer-fcis/master/examples/voc/images/SBD_test2008_000090.jpg
python demo.py SBD_test2008_000090.jpg --gpu 0
```
fcis_voc_vis

### DIFF
--- a/chainercv/experimental/links/model/fcis/fcis.py
+++ b/chainercv/experimental/links/model/fcis/fcis.py
@@ -11,6 +11,58 @@ from chainercv.transforms.image.resize import resize
 
 class FCIS(chainer.Chain):
 
+    """Base class for FCIS.
+
+    This is a base class for FCIS links supporting instance segmentation
+    API [#]_. The following three stages constitute FCIS.
+
+    1. **Feature extraction**: Images are taken and their \
+        feature maps are calculated.
+    2. **Region Proposal Networks**: Given the feature maps calculated in \
+        the previous stage, produce set of RoIs around objects.
+    3. **Localization, Segmentation and Classification Heads**: Using feature \
+        maps that belong to the proposed RoIs, segment region of the objects, \
+        classify the categories of the objects in the RoIs and improve \
+        localizations.
+
+    Each stage is carried out by one of the callable
+    :class:`chainer.Chain` objects :obj:`feature`, :obj:`rpn` and :obj:`head`.
+    There are two functions :meth:`predict` and :meth:`__call__` to conduct
+    instance segmentation.
+    :meth:`predict` takes images and returns masks, object label and its score.
+    :meth:`__call__` is provided for a scnerario when intermediate outputs
+    are needed, for instance, for training and debugging.
+
+    Links that support instance segmentation API have method :meth:`predict`
+    with the same interface. Please refer to :meth:`predict` for further
+    details.
+
+    .. [#] Yi Li, Haozhi Qi, Jifeng Dai, Xiangyang Ji, Yichen Wei. \
+    Fully Convolutional Instance-aware Semantic Segmentation. CVPR 2017.
+
+    Args:
+        extractor (callable Chain): A callable that takes a BCHW image
+            array and returns feature maps.
+        rpn (callable Chain): A callable that has the same interface as
+            :class:`~chainercv.links.model.faster_rcnn.RegionProposalNetwork`.
+            Please refer to the documentation found there.
+        head (callable Chain): A callable that takes a BCHW array,
+        RoIs and batch indices for RoIs.
+        This returns class dependent segmentation scores, class-agnostic
+        localization parameters, class scores, improved RoIs and batch
+        indices for RoIs.
+        mean (numpy.ndarray): A value to be subtracted from an image
+            in :meth:`prepare`.
+        min_size (int): A preprocessing paramter for :meth:`prepare`. Please
+            refer to a docstring found for :meth:`prepare`.
+        max_size (int): A preprocessing paramter for :meth:`prepare`.
+        loc_normalize_mean (tuple of four floats): Mean values of
+            localization estimates.
+        loc_normalize_std (tupler of four floats): Standard deviation
+            of localization estimates.
+
+    """
+
     def __init__(
             self, extractor, rpn, head,
             mean, min_size, max_size,
@@ -36,6 +88,48 @@ class FCIS(chainer.Chain):
         return self.head.n_class
 
     def __call__(self, x, scale=1.):
+        """Forward FCIS.
+
+        Scaling paramter :obj:`scale` is used by RPN to determine the
+        threshold to select small objects, which are going to be
+        rejected irrespective of their confidence scores.
+
+        Here are notations used.
+
+        * :math:`N` is the number of batch size
+        * :math:`R'` is the total number of RoIs produced across batches. \
+            Given :math:`R_i` proposed RoIs from the :math:`i` th image, \
+            :math:`R' = \\sum _{i=1} ^ N R_i`.
+        * :math:`L` is the number of classes excluding the background.
+        * :math:`RH` is the height of pooled image by Position Sensitive \
+        ROI pooling.
+        * :math:`RW` is the height of pooled image by Position Sensitive \
+        ROI pooling.
+
+        Classes are ordered by the background, the first class, ..., and
+        the :math:`L` th class.
+
+        Args:
+            x (~chainer.Variable): 4D image variable.
+            scale (float): Amount of scaling applied to the raw image
+                during preprocessing.
+
+        Returns:
+            Variable, Variable, Variable, array, array:
+            Returns tuple of five values listed below.
+
+            * **roi_seg_scores**: Class-agnostic segmentation scores for \
+            the proposed ROIs. Its shape is :math:`(R', 2, RH, RW)`
+            * **roi_ag_locs**: Class-agnostic offsets and scalings for \
+            the proposed RoIs.  Its shape is :math:`(R', 2, 4)`.
+            * **roi_scores**: Class predictions for the proposed RoIs. \
+                Its shape is :math:`(R', L + 1)`.
+            * **rois**: RoIs proposed by RPN. Its shape is \
+                :math:`(R', 4)`.
+            * **roi_indices**: Batch indices of RoIs. Its shape is \
+                :math:`(R',)`.
+
+        """
         img_size = x.shape[2:]
 
         # Feature Extractor
@@ -80,11 +174,16 @@ class FCIS(chainer.Chain):
     def use_preset(self, preset):
         """Use the given preset during prediction.
 
-        This method changes values of :obj:`self.nms_thresh` and
-        :obj:`self.score_thresh`. These values are a threshold value
-        used for non maximum suppression and a threshold value
+        This method changes values of :obj:`self.nms_thresh`,
+        :obj:`self.score_thresh`, :obj:`self.mask_merge_thresh`,
+        :obj:`self.binary_thresh`, :obj:`self.binary_thresh` and
+        :obj:`self.min_drop_size`. These values are a threshold value
+        used for non maximum suppression, a threshold value
         to discard low confidence proposals in :meth:`predict`,
-        respectively.
+        a threshold value to merge mask in :meth:`predict`,
+        a threshold value to binalize segmentation scores in :meth:`predict`,
+        a limit number of predicted masks in one image and
+        a threshold value to discard small bounding boxes respectively.
 
         If the attributes need to be changed to something
         other than the values provided in the presets, please modify
@@ -113,6 +212,31 @@ class FCIS(chainer.Chain):
             raise ValueError('preset must be visualize or evaluate')
 
     def predict(self, imgs):
+        """Segment object instances from images.
+
+        This method predicts instance-aware object region for each image.
+
+        Args:
+            imgs (iterable of numpy.ndarray): Arrays holding images of shape
+                :math:`(B, C, H, W)`.  All images are in CHW and RGB format
+                and the range of their value is :math:`[0, 255]`.
+
+        Returns:
+           tuple of lists:
+           This method returns a tuple of three lists,
+           :obj:`(masks, labels, scores)`.
+
+           * **masks**: A list of float arrays of shape :math:`(R, H, W)`, \
+               where :math:`R` is the number of masks in a image. \
+               Each pixel holds value if it is inside the object inside or not.
+           * **labels** : A list of integer arrays of shape :math:`(R,)`. \
+               Each value indicates the class of the masks. \
+               Values are in range :math:`[0, L - 1]`, where :math:`L` is the \
+               number of the foreground classes.
+           * **scores** : A list of float arrays of shape :math:`(R,)`. \
+               Each value indicates how confident the prediction is.
+
+        """
 
         prepared_imgs = []
         sizes = []
--- a/chainercv/experimental/links/model/fcis/fcis.py
+++ b/chainercv/experimental/links/model/fcis/fcis.py
@@ -119,11 +119,11 @@ class FCIS(chainer.Chain):
             Variable, Variable, Variable, array, array:
             Returns tuple of five values listed below.
 
-            * **roi_seg_scores**: Class-agnostic segmentation scores for \
+            * **roi_mask_scores**: Class-agnostic segmentation scores for \
                 the proposed ROIs. Its shape is :math:`(R', 2, RH, RW)`
-            * **roi_ag_locs**: Class-agnostic offsets and scalings for \
+            * **ag_locs**: Class-agnostic offsets and scalings for \
                 the proposed RoIs.  Its shape is :math:`(R', 2, 4)`.
-            * **roi_scores**: Class predictions for the proposed RoIs. \
+            * **scores**: Class predictions for the proposed RoIs. \
                 Its shape is :math:`(R', L + 1)`.
             * **rois**: RoIs proposed by RPN. Its shape is \
                 :math:`(R', 4)`.
@@ -137,9 +137,9 @@ class FCIS(chainer.Chain):
         rpn_features, roi_features = self.extractor(x)
         rpn_locs, rpn_scores, rois, roi_indices, anchor = self.rpn(
             rpn_features, img_size, scale)
-        roi_seg_scores, roi_ag_locs, roi_scores, rois, roi_indices = \
+        roi_mask_scores, ag_locs, cls_scores, rois, roi_indices = \
             self.head(roi_features, rois, roi_indices, img_size)
-        return roi_seg_scores, roi_ag_locs, roi_scores, rois, roi_indices
+        return roi_mask_scores, ag_locs, cls_scores, rois, roi_indices
 
     def prepare(self, img):
         """Preprocess an image for feature extraction.
@@ -257,27 +257,27 @@ class FCIS(chainer.Chain):
                 # inference
                 img_var = chainer.Variable(self.xp.array(img[None]))
                 scale = img_var.shape[3] / size[1]
-                roi_seg_scores, _, roi_scores, rois, _ = \
+                roi_mask_scores, _, cls_scores, bboxes, _ = \
                     self.__call__(img_var, scale)
 
             # We are assuming that batch size is 1.
-            roi_seg_score = roi_seg_scores.array
-            roi_score = roi_scores.array
-            bbox = rois / scale
+            roi_mask_score = roi_mask_scores.array
+            cls_score = cls_scores.array
+            bbox = bboxes / scale
 
             # shape: (n_rois, 4)
             bbox[:, 0::2] = self.xp.clip(bbox[:, 0::2], 0, size[0])
             bbox[:, 1::2] = self.xp.clip(bbox[:, 1::2], 0, size[1])
 
-            roi_seg_prob = F.softmax(roi_seg_score).array
-            roi_prob = F.softmax(roi_score).array
+            roi_mask_prob = F.softmax(roi_mask_score).array[:, 1, :, :]
+            cls_prob = F.softmax(cls_score).array
 
-            roi_seg_prob = chainer.cuda.to_cpu(roi_seg_prob)
-            roi_prob = chainer.cuda.to_cpu(roi_prob)
+            roi_mask_prob = chainer.cuda.to_cpu(roi_mask_prob)
+            cls_prob = chainer.cuda.to_cpu(cls_prob)
             bbox = chainer.cuda.to_cpu(bbox)
 
-            roi_mask_score, bbox, label, score = mask_voting(
-                roi_seg_prob[:, 1, :, :], bbox, roi_prob, size,
+            roi_mask_prob, bbox, label, cls_prob = mask_voting(
+                roi_mask_prob, bbox, cls_prob, size,
                 self.score_thresh, self.nms_thresh,
                 self.mask_merge_thresh, self.binary_thresh,
                 limit=self.limit, bg_label=0)
@@ -287,24 +287,24 @@ class FCIS(chainer.Chain):
             keep_indices = np.where(
                 (height > self.min_drop_size) &
                 (width > self.min_drop_size))[0]
+            roi_mask_prob = roi_mask_prob[keep_indices]
             bbox = bbox[keep_indices]
-            roi_mask_score = roi_mask_score[keep_indices]
-            score = score[keep_indices]
             label = label[keep_indices]
+            cls_prob = cls_prob[keep_indices]
 
             mask = np.zeros(
-                (len(roi_mask_score), size[0], size[1]), dtype=np.bool)
-            for i, (roi_msk_sc, bb) in enumerate(zip(roi_mask_score, bbox)):
+                (len(roi_mask_prob), size[0], size[1]), dtype=np.bool)
+            for i, (roi_msk_pb, bb) in enumerate(zip(roi_mask_prob, bbox)):
                 bb = np.round(bb).astype(np.int32)
                 y_min, x_min, y_max, x_max = bb
-                roi_msk_sc = resize(
-                    roi_msk_sc.astype(np.float32)[None],
+                roi_msk_pb = resize(
+                    roi_msk_pb.astype(np.float32)[None],
                     (y_max - y_min, x_max - x_min))
-                roi_msk = (roi_msk_sc > self.binary_thresh)[0]
+                roi_msk = (roi_msk_pb > self.binary_thresh)[0]
                 mask[i, y_min:y_max, x_min:x_max] = roi_msk
 
             masks.append(mask)
             labels.append(label)
-            scores.append(score)
+            scores.append(cls_prob)
 
         return masks, labels, scores
--- a/chainercv/experimental/links/model/fcis/fcis.py
+++ b/chainercv/experimental/links/model/fcis/fcis.py
@@ -1,0 +1,187 @@
+from __future__ import division
+
+import chainer
+import chainer.functions as F
+import numpy as np
+
+from chainercv.experimental.links.model.fcis.utils.mask_voting \
+    import mask_voting
+from chainercv.transforms.image.resize import resize
+
+
+class FCIS(chainer.Chain):
+
+    def __init__(
+            self, extractor, rpn, head,
+            mean, min_size, max_size,
+            loc_normalize_mean, loc_normalize_std,
+    ):
+        super(FCIS, self).__init__()
+        with self.init_scope():
+            self.extractor = extractor
+            self.rpn = rpn
+            self.head = head
+
+        self.mean = mean
+        self.min_size = min_size
+        self.max_size = max_size
+        self.loc_normalize_mean = loc_normalize_mean
+        self.loc_normalize_std = loc_normalize_std
+
+        self.use_preset('visualize')
+
+    @property
+    def n_class(self):
+        # Total number of classes including the background.
+        return self.head.n_class
+
+    def __call__(self, x, scale=1.):
+        img_size = x.shape[2:]
+
+        # Feature Extractor
+        rpn_features, roi_features = self.extractor(x)
+        rpn_locs, rpn_scores, rois, roi_indices, anchor = self.rpn(
+            rpn_features, img_size, scale)
+        roi_seg_scores, roi_ag_locs, roi_scores, rois, roi_indices = \
+            self.head(roi_features, rois, roi_indices, img_size)
+        return roi_seg_scores, roi_ag_locs, roi_scores, rois, roi_indices
+
+    def prepare(self, img):
+        """Preprocess an image for feature extraction.
+
+        The length of the shorter edge is scaled to :obj:`self.min_size`.
+        After the scaling, if the length of the longer edge is longer than
+        :obj:`self.max_size`, the image is scaled to fit the longer edge
+        to :obj:`self.max_size`.
+
+        After resizing the image, the image is subtracted by a mean image value
+        :obj:`self.mean`.
+
+        Args:
+            img (~numpy.ndarray): An image. This is in CHW and RGB format.
+                The range of its value is :math:`[0, 255]`.
+
+        Returns:
+            ~numpy.ndarray:
+            A preprocessed image.
+
+        """
+        _, H, W = img.shape
+
+        scale = self.min_size / min(H, W)
+
+        if scale * max(H, W) > self.max_size:
+            scale = self.max_size / max(H, W)
+
+        img = resize(img, (int(H * scale), int(W * scale)))
+        img = (img - self.mean).astype(np.float32, copy=False)
+        return img
+
+    def use_preset(self, preset):
+        """Use the given preset during prediction.
+
+        This method changes values of :obj:`self.nms_thresh` and
+        :obj:`self.score_thresh`. These values are a threshold value
+        used for non maximum suppression and a threshold value
+        to discard low confidence proposals in :meth:`predict`,
+        respectively.
+
+        If the attributes need to be changed to something
+        other than the values provided in the presets, please modify
+        them by directly accessing the public attributes.
+
+        Args:
+            preset ({'visualize', 'evaluate'): A string to determine the
+                preset to use.
+
+        """
+        if preset == 'visualize':
+            self.nms_thresh = 0.3
+            self.score_thresh = 0.7
+            self.mask_merge_thresh = 0.5
+            self.binary_thresh = 0.4
+            self.limit = 100
+            self.min_drop_size = 16
+        elif preset == 'evaluate':
+            self.nms_thresh = 0.3
+            self.score_thresh = 1e-3
+            self.mask_merge_thresh = 0.5
+            self.binary_thresh = 0.4
+            self.limit = 100
+            self.min_drop_size = 16
+        else:
+            raise ValueError('preset must be visualize or evaluate')
+
+    def predict(self, imgs):
+
+        prepared_imgs = []
+        sizes = []
+        for img in imgs:
+            size = img.shape[1:]
+            img = self.prepare(img.astype(np.float32))
+            prepared_imgs.append(img)
+            sizes.append(size)
+
+        masks = []
+        labels = []
+        scores = []
+        bboxes = []
+
+        for img, size in zip(prepared_imgs, sizes):
+            with chainer.using_config('train', False), \
+                    chainer.function.no_backprop_mode():
+                # inference
+                img_var = chainer.Variable(self.xp.array(img[None]))
+                scale = img_var.shape[3] / size[1]
+                roi_seg_scores, roi_ag_locs, roi_scores, rois, _ = \
+                    self.__call__(img_var, scale)
+
+            roi_seg_score = roi_seg_scores.array
+            roi_score = roi_scores.array
+            bbox = rois / scale
+
+            # shape: (n_rois, 4)
+            bbox[:, 0::2] = self.xp.clip(bbox[:, 0::2], 0, size[0])
+            bbox[:, 1::2] = self.xp.clip(bbox[:, 1::2], 0, size[1])
+
+            roi_seg_prob = F.softmax(roi_seg_score).array
+            roi_prob = F.softmax(roi_score).array
+
+            roi_seg_prob = chainer.cuda.to_cpu(roi_seg_prob)
+            roi_prob = chainer.cuda.to_cpu(roi_prob)
+            bbox = chainer.cuda.to_cpu(bbox)
+
+            roi_mask_prob, label, score, bbox = mask_voting(
+                roi_seg_prob[:, 1, :, :], roi_prob, bbox,
+                size, self.n_class,
+                self.score_thresh, self.nms_thresh,
+                self.mask_merge_thresh, self.binary_thresh,
+                limit=self.limit, bg_label=0)
+
+            height = bbox[:, 2] - bbox[:, 0]
+            width = bbox[:, 3] - bbox[:, 1]
+            keep_indices = np.where(
+                (height > self.min_drop_size) &
+                (width > self.min_drop_size))[0]
+            bbox = bbox[keep_indices]
+            roi_mask_prob = roi_mask_prob[keep_indices]
+            score = score[keep_indices]
+            label = label[keep_indices]
+
+            mask = np.zeros(
+                (len(roi_mask_prob), size[0], size[1]), dtype=np.bool)
+            for i, (roi_msk_prb, bb) in enumerate(zip(roi_mask_prob, bbox)):
+                bb = np.round(bb).astype(np.int32)
+                y_min, x_min, y_max, x_max = bb
+                roi_msk_prb = resize(
+                    roi_msk_prb.astype(np.float32)[None],
+                    (y_max - y_min, x_max - x_min))
+                roi_msk = (roi_msk_prb > self.binary_thresh)[0]
+                mask[i, y_min:y_max, x_min:x_max] = roi_msk
+
+            masks.append(mask)
+            labels.append(label)
+            scores.append(score)
+            bboxes.append(bbox)
+
+        return masks, labels, scores, bboxes
--- a/chainercv/experimental/links/model/fcis/fcis.py
+++ b/chainercv/experimental/links/model/fcis/fcis.py
@@ -125,7 +125,6 @@ class FCIS(chainer.Chain):
         masks = []
         labels = []
         scores = []
-        bboxes = []
 
         for img, size in zip(prepared_imgs, sizes):
             with chainer.using_config('train', False), \
@@ -182,6 +181,5 @@ class FCIS(chainer.Chain):
             masks.append(mask)
             labels.append(label)
             scores.append(score)
-            bboxes.append(bbox)
 
-        return masks, labels, scores, bboxes
+        return masks, labels, scores
--- a/chainercv/experimental/links/model/fcis/fcis.py
+++ b/chainercv/experimental/links/model/fcis/fcis.py
@@ -260,6 +260,7 @@ class FCIS(chainer.Chain):
                 roi_seg_scores, _, roi_scores, rois, _ = \
                     self.__call__(img_var, scale)
 
+            # We are assuming that batch size is 1.
             roi_seg_score = roi_seg_scores.array
             roi_score = roi_scores.array
             bbox = rois / scale
--- a/chainercv/experimental/links/model/fcis/fcis.py
+++ b/chainercv/experimental/links/model/fcis/fcis.py
@@ -150,9 +150,8 @@ class FCIS(chainer.Chain):
             roi_prob = chainer.cuda.to_cpu(roi_prob)
             bbox = chainer.cuda.to_cpu(bbox)
 
-            roi_mask_prob, label, score, bbox = mask_voting(
-                roi_seg_prob[:, 1, :, :], roi_prob, bbox,
-                size, self.n_class,
+            roi_mask_score, bbox, label, score = mask_voting(
+                roi_seg_prob[:, 1, :, :], bbox, roi_prob, size,
                 self.score_thresh, self.nms_thresh,
                 self.mask_merge_thresh, self.binary_thresh,
                 limit=self.limit, bg_label=0)
@@ -163,19 +162,19 @@ class FCIS(chainer.Chain):
                 (height > self.min_drop_size) &
                 (width > self.min_drop_size))[0]
             bbox = bbox[keep_indices]
-            roi_mask_prob = roi_mask_prob[keep_indices]
+            roi_mask_score = roi_mask_score[keep_indices]
             score = score[keep_indices]
             label = label[keep_indices]
 
             mask = np.zeros(
-                (len(roi_mask_prob), size[0], size[1]), dtype=np.bool)
-            for i, (roi_msk_prb, bb) in enumerate(zip(roi_mask_prob, bbox)):
+                (len(roi_mask_score), size[0], size[1]), dtype=np.bool)
+            for i, (roi_msk_sc, bb) in enumerate(zip(roi_mask_score, bbox)):
                 bb = np.round(bb).astype(np.int32)
                 y_min, x_min, y_max, x_max = bb
-                roi_msk_prb = resize(
-                    roi_msk_prb.astype(np.float32)[None],
+                roi_msk_sc = resize(
+                    roi_msk_sc.astype(np.float32)[None],
                     (y_max - y_min, x_max - x_min))
-                roi_msk = (roi_msk_prb > self.binary_thresh)[0]
+                roi_msk = (roi_msk_sc > self.binary_thresh)[0]
                 mask[i, y_min:y_max, x_min:x_max] = roi_msk
 
             masks.append(mask)
--- a/chainercv/experimental/links/model/fcis/fcis.py
+++ b/chainercv/experimental/links/model/fcis/fcis.py
@@ -132,7 +132,7 @@ class FCIS(chainer.Chain):
                 # inference
                 img_var = chainer.Variable(self.xp.array(img[None]))
                 scale = img_var.shape[3] / size[1]
-                roi_seg_scores, roi_ag_locs, roi_scores, rois, _ = \
+                roi_seg_scores, _, roi_scores, rois, _ = \
                     self.__call__(img_var, scale)
 
             roi_seg_score = roi_seg_scores.array
--- a/chainercv/experimental/links/model/fcis/fcis.py
+++ b/chainercv/experimental/links/model/fcis/fcis.py
@@ -21,15 +21,16 @@ class FCIS(chainer.Chain):
     2. **Region Proposal Networks**: Given the feature maps calculated in \
         the previous stage, produce set of RoIs around objects.
     3. **Localization, Segmentation and Classification Heads**: Using feature \
-        maps that belong to the proposed RoIs, segment region of the objects, \
-        classify the categories of the objects in the RoIs and improve \
-        localizations.
+        maps that belong to the proposed RoIs, segment regions of the \
+        objects, classify the categories of the objects in the RoIs and \
+        improve localizations.
 
     Each stage is carried out by one of the callable
     :class:`chainer.Chain` objects :obj:`feature`, :obj:`rpn` and :obj:`head`.
     There are two functions :meth:`predict` and :meth:`__call__` to conduct
     instance segmentation.
-    :meth:`predict` takes images and returns masks, object label and its score.
+    :meth:`predict` takes images and returns masks, object labels
+    and their scores.
     :meth:`__call__` is provided for a scnerario when intermediate outputs
     are needed, for instance, for training and debugging.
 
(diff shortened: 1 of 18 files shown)


chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+            loc_normalize_mean, loc_normalize_std)
+
+        if pretrained_model in self._models:
+            path = download_model(self._models[pretrained_model]['url'])
+            chainer.serializers.load_npz(path, self)
+
+
+class ResNet101Extractor(chainer.Chain):
+
+    def __init__(self):
```
yuyu2172: Can you change the feature extractor to take `initialW` as an argument, and pass the zero-initializer when a pretrained model is specified?

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+                3, 64, 7, 2, 3, nobias=True, initialW=initialW)
+            self.pool1 = lambda x: F.max_pooling_2d(x, ksize=3, stride=2)
+            self.res2 = ResBlock(3, 64, 64, 256, 1, **kwargs)
+            self.res3 = ResBlock(4, 256, 128, 512, 2, **kwargs)
+            self.res4 = ResBlock(23, 512, 256, 1024, 2, **kwargs)
+            self.res5 = ResBlock(3, 1024, 512, 2048, 1, 2, **kwargs)
+
+    def __call__(self, x):
+        with chainer.using_config('train', False):
+            with chainer.function.no_backprop_mode():
```
yuyu2172: I prefer setting related backpropagation in the training code.
Can you remove `with chainer.function.no_backprop_mode` from here?

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        self.iter2 = iter2
+
+        initialW = chainer.initializers.Normal(0.01)
+        with self.init_scope():
+            self.psroi_conv1 = L.Convolution2D(
+                2048, 1024, 1, 1, 0, initialW=initialW)
+            self.psroi_conv2 = L.Convolution2D(
+                1024, group_size * group_size * n_class * 2,
+                1, 1, 0, initialW=initialW)
+            self.psroi_conv3 = L.Convolution2D(
```
yuyu2172: How about changing the name of the links to `conv1`, `cls_seg` and `ag_loc`?
This naming convention looks as if three convolutions are applied sequentially to the input.

Alternatively, `psroi_conv1`, `psroi_cls_seg`, `psroi_ag_loc` is OK (I prefer the first choice for simplicity).

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+                1024, group_size * group_size * 2 * 4,
+                1, 1, 0, initialW=initialW)
+
+    def __call__(self, x, rois, roi_indices, img_size):
+        h = F.relu(self.psroi_conv1(x))
+        h_cls_seg = self.psroi_conv2(h)
+        h_ag_locs = self.psroi_conv3(h)
```
yuyu2172: How about `h_ag_loc` instead of `h_ag_loc`.
We usually do not put `s` with variables starting with `h_` (i.e., not `hs`).
Also, this looks more consistent with `h_cls_seg`, which you did not put `s`.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        self.group_size = group_size
+        self.roi_size = roi_size
+        self.loc_normalize_mean = loc_normalize_mean
+        self.loc_normalize_std = loc_normalize_std
+        self.iter2 = iter2
+
+        initialW = chainer.initializers.Normal(0.01)
+        with self.init_scope():
+            self.psroi_conv1 = L.Convolution2D(
+                2048, 1024, 1, 1, 0, initialW=initialW)
```
yuyu2172: How about using `Conv2DActiv`?

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+            roi_ag_locs = F.concat(
+                (roi_ag_locs, roi_ag_locs2), axis=0)
+            roi_cls_scores = F.concat(
+                (roi_cls_scores, roi_cls_scores2), axis=0)
+        return roi_seg_scores, roi_ag_locs, roi_cls_scores, rois, roi_indices
+
+    def _pool(
+            self, h_cls_seg, h_ag_locs, rois, roi_indices):
+        # PSROI Pooling
+        # shape: (n_rois, n_class*2, roi_size, roi_size)
```
yuyu2172: `n_roi` instead of `n_rois`

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        pool_cls_seg = psroi_pooling_2d(
+            h_cls_seg, rois, roi_indices,
+            self.n_class * 2, self.roi_size, self.roi_size,
+            self.spatial_scale, self.group_size)
+        # shape: (n_rois, n_class, 2, roi_size, roi_size)
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        pool_cls_seg = pool_cls_seg.reshape(
+            (-1, self.n_class, 2, self.roi_size, self.roi_size))
+        # shape: (n_rois, 2*4, roi_size, roi_size)
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        pool_ag_locs = psroi_pooling_2d(
+            h_ag_locs, rois, roi_indices,
+            2 * 4, self.roi_size, self.roi_size,
+            self.spatial_scale, self.group_size)
+
+        # Classfication
+        # Group Max
+        # shape: (n_rois, n_class, roi_size, roi_size)
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        h_cls = pool_cls_seg.transpose((0, 1, 3, 4, 2))
+        h_cls = F.max(h_cls, axis=4)
+
+        # Global pooling (vote)
+        # shape: (n_rois, n_class)
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        roi_cls_scores = _global_average_pooling_2d(h_cls)
+
+        # Bbox Regression
+        # shape: (n_rois, 2*4)
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        roi_ag_locs = _global_average_pooling_2d(pool_ag_locs)
+        n_roi = roi_ag_locs.shape[0]
+        roi_ag_locs = roi_ag_locs.reshape((n_roi, 2, 4))
+
+        # Mask Regression
+        # shape: (n_rois, n_class, 2, roi_size, roi_size)
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        # Group Pick by Score
+        max_cls_indices = roi_cls_scores.array.argmax(axis=1)
+        # shape: (n_rois, 2, roi_size, roi_size)
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        roi_seg_scores = pool_cls_seg[
+            np.arange(len(max_cls_indices)), max_cls_indices]
+
+        return roi_seg_scores, roi_ag_locs, roi_cls_scores
+
+
+def _global_average_pooling_2d(x):
+    n_rois, n_channel, H, W = x.array.shape
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        pool_ag_locs = psroi_pooling_2d(
+            h_ag_locs, rois, roi_indices,
+            2 * 4, self.roi_size, self.roi_size,
+            self.spatial_scale, self.group_size)
+
+        # Classfication
+        # Group Max
+        # shape: (n_rois, n_class, roi_size, roi_size)
+        h_cls = pool_cls_seg.transpose((0, 1, 3, 4, 2))
+        h_cls = F.max(h_cls, axis=4)
```
yuyu2172: We can remvoe transpose by using the right axis when taking maximum.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+        n_roi = roi_ag_locs.shape[0]
+        roi_ag_locs = roi_ag_locs.reshape((n_roi, 2, 4))
+
+        # Mask Regression
+        # shape: (n_rois, n_class, 2, roi_size, roi_size)
+        # Group Pick by Score
+        max_cls_indices = roi_cls_scores.array.argmax(axis=1)
+        # shape: (n_rois, 2, roi_size, roi_size)
+        roi_seg_scores = pool_cls_seg[
+            np.arange(len(max_cls_indices)), max_cls_indices]
```
yuyu2172: Use `self.xp.arange`.
This is faster because `np.arange` need to transfer data from CPU to GPU.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+                (roi_ag_locs, roi_ag_locs2), axis=0)
+            roi_cls_scores = F.concat(
+                (roi_cls_scores, roi_cls_scores2), axis=0)
+        return roi_seg_scores, roi_ag_locs, roi_cls_scores, rois, roi_indices
+
+    def _pool(
+            self, h_cls_seg, h_ag_locs, rois, roi_indices):
+        # PSROI Pooling
+        # shape: (n_rois, n_class*2, roi_size, roi_size)
+        pool_cls_seg = psroi_pooling_2d(
```
yuyu2172: May I suggest to use `roi_seg_cls_scores` instead of `pool_cls_seg`?

`roi_seg_scores` is `(n_roi, 2, R, R)` and `roi_cls_scores` is`(n_roi, n_class)`.
For a variable with shape `(n_roi, n_class, 2, R, R)`, I think `roi_seg_cls_scores` is the appropriate name.
Basically, `_cls_` indicates that we have extra dimension that corresponds to categories.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+            h_cls_seg, rois, roi_indices,
+            self.n_class * 2, self.roi_size, self.roi_size,
+            self.spatial_scale, self.group_size)
+        # shape: (n_rois, n_class, 2, roi_size, roi_size)
+        pool_cls_seg = pool_cls_seg.reshape(
+            (-1, self.n_class, 2, self.roi_size, self.roi_size))
+        # shape: (n_rois, 2*4, roi_size, roi_size)
+        pool_ag_locs = psroi_pooling_2d(
```
yuyu2172: May I suggest to name `(n_roi, 2 * 4, R, R)` as `roi_seg_ag_locs` instead of `pool_ag_locs`?
This seems like an consistent usage of `_seg_`, which is used for all variables with `(n_roi, *, R, R)`.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+from chainercv.links.model.faster_rcnn.utils.loc2bbox import loc2bbox
+from chainercv.links.model.resnet.resblock import ResBlock
+from chainercv.utils import download_model
+
+
+class FCISResNet101(FCIS):
+
+    _models = {
+        'sbd': {
+            'n_fg_class': 20,
```
yuyu2172: Is setting the name of the pretrained weight `voc` inappropriate?

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,234 @@
...
+            'url': 'https://github.com/yuyu2172/share-weights/releases/'
+            'download/0.0.6/fcis_resnet101_2018_04_14.npz'
```
yuyu2172: Could you name the weight in the following protocol?
`fcis_resnet101_DATASETNAME_trained_DATE.npz`?
(e.g., `faster_rcnn_vgg16_voc07_trained_2017_08_06.npz`).

chainercv/experimental/links/model/fcis/fcis_resnet101.py
knorth55: Hmm, in that case, I cannot reuse this model in training code...

chainercv/experimental/links/model/fcis/fcis_resnet101.py
knorth55: hmmm, i'm don't see any merits of using Conv2DActiv...
It is easy to use default one for me.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
knorth55: VOC images with SBD annotations, so I prefer SBD.
Also, the model is trained with SBD training images, which differs from VOC train images.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
knorth55: OK, i will update

chainercv/experimental/links/model/fcis/fcis_resnet101.py
knorth55: OK

chainercv/experimental/links/model/fcis/fcis_resnet101.py
knorth55: OK

chainercv/experimental/links/model/fcis/fcis_resnet101.py
knorth55: OK

chainercv/experimental/links/model/fcis/fcis_resnet101.py
knorth55: So, you prefer setting `disable_update()` by hand in training code?
OK, i will do that.
In fact, I think `no_backprop_mode` and `disable_update` are confusing me many times.

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
...
+
+        c_bbox = np.array(
+            [y_min, x_min, y_max, x_max],
+            dtype=np.float32)
+        c_mask = mask[y_min:y_max, x_min:x_max]
+        return c_mask[None], c_bbox[None]
+
+
+def mask_voting(
+        mask_score, bbox, score, size,
```
yuyu2172: I have two comments on the names of variables.

First, if the range of the value is `[0, 1]`, use `_prob` instead of `_score`.

Second, the name `mask_scores` can be misunderstood as an array with shape `(R, H, W)`. How about naming a variable `(R, RH, RW)` as  `roi_mask_prob`.  (depending on the range of its values). This name is consistent with names in `fcis.py`.
Also, how about renaming `bbox` to `roi` so that it is clear that `roi_mask_prob` is defined in the region of `roi`.
When iterating over `roi`, either `for bb in roi:` or `for r in roi` is OK.
In addition to that, `score` can be renamed to `roi_prob` to make its shape more precise (this modification is only optional).

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
...
+        score_l = score[:, label]
+        keep_indices = non_maximum_suppression(
+            bbox, nms_thresh, score_l)
+        bbox_l = bbox[keep_indices]
+        score_l = score_l[keep_indices]
+        cls_bbox.append(bbox_l)
+        cls_score.append(score_l)
+
+    sorted_score = np.sort(np.concatenate(cls_score))[::-1]
+    keep_n = min(len(sorted_score), limit)
```
yuyu2172: `n_keep`

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+                    chainer.function.no_backprop_mode():
+                # inference
+                img_var = chainer.Variable(self.xp.array(img[None]))
+                scale = img_var.shape[3] / size[1]
+                roi_seg_scores, _, roi_scores, rois, _ = \
+                    self.__call__(img_var, scale)
+
+            roi_seg_score = roi_seg_scores.array
+            roi_score = roi_scores.array
+            bbox = rois / scale
```
yuyu2172: How about not introducing `bbox`.
I mean use `roi = rois / scale`.

The rationale to prefer `roi` is explained in the second suggestion of https://github.com/chainer/chainercv/pull/568#discussion_r183230729.

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+
+            # shape: (n_rois, 4)
+            bbox[:, 0::2] = self.xp.clip(bbox[:, 0::2], 0, size[0])
+            bbox[:, 1::2] = self.xp.clip(bbox[:, 1::2], 0, size[1])
+
+            roi_seg_prob = F.softmax(roi_seg_score).array
```
yuyu2172: How about `roi_seg_prob = F.softmax(roi_seg_score).array[:, 1, :, :]` so that the naming convention is consistent with `mask_voting.py`.

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+            roi_prob = F.softmax(roi_score).array
+
+            roi_seg_prob = chainer.cuda.to_cpu(roi_seg_prob)
+            roi_prob = chainer.cuda.to_cpu(roi_prob)
+            bbox = chainer.cuda.to_cpu(bbox)
+
+            roi_mask_score, bbox, label, score = mask_voting(
```
yuyu2172: Isn't the returned mask a probability?
In that case, can you rename `roi_mask_score` to `roi_mask_prob`.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,350 @@
...
+            rois2[:, 1::2] = self.xp.clip(rois2[:, 1::2], 0, img_size[1])
+
+            # PSROI pooling and regression
+            roi_seg_scores2, roi_ag_locs2, roi_cls_scores2 = self._pool(
+                h_cls_seg, h_ag_loc, rois2, roi_indices)
+
+            # concat 1st and 2nd iteration results
+            rois = self.xp.concatenate((rois, rois2))
+            roi_indices = self.xp.concatenate((roi_indices, roi_indices))
+            roi_seg_scores = F.concat(
```
yuyu2172: Do you have any particular reason to prefer using `*_seg_*` for names?
I think it is the same as `*_mask_*`, so how about using only one convention.
Since `mask` is used for the output, how about choosing `*_mask_*`.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,350 @@
...
+        # PSROI pooling and regression
+        roi_seg_scores, roi_ag_locs, roi_cls_scores = self._pool(
+            h_cls_seg, h_ag_loc, rois, roi_indices)
+        if self.iter2:
+            # 2nd Iteration
+            # get rois2 for more precise prediction
+            roi_ag_locs = roi_ag_locs.array
+            mean = self.xp.array(self.loc_normalize_mean)
+            std = self.xp.array(self.loc_normalize_std)
+            roi_loc = roi_ag_locs[:, 1, :]
```
yuyu2172: `roi_loc` --> `roi_locs`

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+
+        for img, size in zip(prepared_imgs, sizes):
+            with chainer.using_config('train', False), \
+                    chainer.function.no_backprop_mode():
+                # inference
+                img_var = chainer.Variable(self.xp.array(img[None]))
+                scale = img_var.shape[3] / size[1]
+                roi_seg_scores, _, roi_scores, rois, _ = \
+                    self.__call__(img_var, scale)
+
```
yuyu2172: Put a comment like this https://github.com/chainer/chainercv/blob/master/chainercv/links/model/faster_rcnn/faster_rcnn.py#L290

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+    This is a base class for FCIS links supporting instance segmentation
+    API [#]_. The following three stages constitute FCIS.
+
+    1. **Feature extraction**: Images are taken and their \
+        feature maps are calculated.
+    2. **Region Proposal Networks**: Given the feature maps calculated in \
+        the previous stage, produce set of RoIs around objects.
+    3. **Localization, Segmentation and Classification Heads**: Using feature \
+        maps that belong to the proposed RoIs, segment region of the objects, \
```
yuyu2172: segment region --> segment regions

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+        classify the categories of the objects in the RoIs and improve \
+        localizations.
+
+    Each stage is carried out by one of the callable
+    :class:`chainer.Chain` objects :obj:`feature`, :obj:`rpn` and :obj:`head`.
+    There are two functions :meth:`predict` and :meth:`__call__` to conduct
+    instance segmentation.
+    :meth:`predict` takes images and returns masks, object label and its score.
```
yuyu2172: object labels and their scores

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+
+    Args:
+        extractor (callable Chain): A callable that takes a BCHW image
+            array and returns feature maps.
+        rpn (callable Chain): A callable that has the same interface as
+            :class:`~chainercv.links.model.faster_rcnn.RegionProposalNetwork`.
+            Please refer to the documentation found there.
+        head (callable Chain): A callable that takes a BCHW array,
+        RoIs and batch indices for RoIs.
+        This returns class dependent segmentation scores, class-agnostic
```
yuyu2172: class agnostic segmentation scores ?

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+        localization parameters, class scores, improved RoIs and batch
+        indices for RoIs.
+        mean (numpy.ndarray): A value to be subtracted from an image
+            in :meth:`prepare`.
+        min_size (int): A preprocessing paramter for :meth:`prepare`. Please
```
yuyu2172: parameter

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+            refer to a docstring found for :meth:`prepare`.
+        max_size (int): A preprocessing paramter for :meth:`prepare`.
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+            self.binary_thresh = 0.4
+            self.limit = 100
+            self.min_drop_size = 16
+        else:
+            raise ValueError('preset must be visualize or evaluate')
+
+    def predict(self, imgs):
+        """Segment object instances from images.
+
+        This method predicts instance-aware object region for each image.
```
yuyu2172: object regions

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+            imgs (iterable of numpy.ndarray): Arrays holding images of shape
+                :math:`(B, C, H, W)`.  All images are in CHW and RGB format
+                and the range of their value is :math:`[0, 255]`.
+
+        Returns:
+           tuple of lists:
+           This method returns a tuple of three lists,
+           :obj:`(masks, labels, scores)`.
+
+           * **masks**: A list of float arrays of shape :math:`(R, H, W)`, \
```
yuyu2172: float --> boolean

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+    Fully Convolutional Instance-aware Semantic Segmentation. CVPR 2017.
+
+    Args:
+        extractor (callable Chain): A callable that takes a BCHW image
+            array and returns feature maps.
+        rpn (callable Chain): A callable that has the same interface as
+            :class:`~chainercv.links.model.faster_rcnn.RegionProposalNetwork`.
+            Please refer to the documentation found there.
+        head (callable Chain): A callable that takes a BCHW array,
+        RoIs and batch indices for RoIs.
```
yuyu2172: indentation

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+
+        Here are notations used.
+
+        * :math:`N` is the number of batch size
+        * :math:`R'` is the total number of RoIs produced across batches. \
+            Given :math:`R_i` proposed RoIs from the :math:`i` th image, \
+            :math:`R' = \\sum _{i=1} ^ N R_i`.
+        * :math:`L` is the number of classes excluding the background.
+        * :math:`RH` is the height of pooled image by Position Sensitive \
+        ROI pooling.
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+        * :math:`RW` is the height of pooled image by Position Sensitive \
+        ROI pooling.
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/fcis.py
```diff
@@ -0,0 +1,308 @@
...
+            x (~chainer.Variable): 4D image variable.
+            scale (float): Amount of scaling applied to the raw image
+                during preprocessing.
+
+        Returns:
+            Variable, Variable, Variable, array, array:
+            Returns tuple of five values listed below.
+
+            * **roi_seg_scores**: Class-agnostic segmentation scores for \
+            the proposed ROIs. Its shape is :math:`(R', 2, RH, RW)`
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
+import numpy as np
+
+from chainercv.transforms.image.resize import resize
+from chainercv.utils.bbox.bbox_iou import bbox_iou
+from chainercv.utils import non_maximum_suppression
+
+
+def mask_aggregation(
```
yuyu2172: make this a private function

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
...
+def mask_voting(
+        mask_score, bbox, score, size,
+        score_thresh, nms_thresh,
+        mask_merge_thresh, binary_thresh,
+        limit=100, bg_label=0
+):
+    """Refine mask probabilities by merging multiple masks.
+
+    First, this function discard invalid masks with non maximum suppression.
+    Then, it merge masks with weight calculated from class probabilities and
```
yuyu2172: merges

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
...
+    iou.
+    This function improve qualities by merging overlapped masks predicted as
```
yuyu2172: improves

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
...
+
+    Here are notations used.
+    * :math:`R'` is the total number of RoIs produced across batches.
+    * :math:`L` is the number of classes excluding the background.
+    * :math:`RH` is the height of pooled image.
+    * :math:`RW` is the height of pooled image.
+
+    Args:
+        mask_scores (array): A mask score array whose shape is
+            :math:`(R, RH, RW)`.
```
yuyu2172: forgetting the second argument

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
...
+    the same object class.
+
+    Here are notations used.
+    * :math:`R'` is the total number of RoIs produced across batches.
+    * :math:`L` is the number of classes excluding the background.
+    * :math:`RH` is the height of pooled image.
+    * :math:`RW` is the height of pooled image.
+
+    Args:
+        mask_scores (array): A mask score array whose shape is
```
yuyu2172: mask_scores --> mask_score

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
...
+            :math:`(R, RH, RW)`.
+        scores (array): A class score array whose shape is
```
yuyu2172: scores --> score

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
...
+            :math:`(R, L + 1)`.
+        size (tuple of int): Original image size.
+        score_thresh (float): A threshold value of the class score.
+        nms_thresh (float): A threshold value of non maximum suppression.
+        mask_merge_thresh (float): A threshold value of the bounding box iou
+            for mask merging.
+        binary_thresh (float): A threshold value of mask score
+            for mask merging.
+        limit (int): A limit number of outputs.
```
yuyu2172: A limit number of outputs --> The maximum number of outputs.

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,170 @@
...
+        bg_label (int): A background label.
```
yuyu2172: The id of the background label.

tests/experimental_tests/links_tests/model_tests/fcis_tests/test_fcis_resnet101.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @attr.gpu
+    @attr.slow
+    def test_call_gpu(self):
+        self.link.to_gpu()
+        self.check_call()
+
+
+class TestFCISResNet101Pretrained(unittest.TestCase):
+
```
yuyu2172: Please add @attr.disk like other dataset tests.

tests/experimental_tests/links_tests/model_tests/fcis_tests/test_fcis_resnet101.py
```diff
@@ -0,0 +1,99 @@
...
+    @attr.slow
+    def test_pretrained(self):
+        link = FCISResNet101(pretrained_model='sbd')
+        self.assertIsInstance(link, FCIS)
+
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/utils/mask_voting.py
yuyu2172: You actually do use something similar to what I suggested in `test_mask_voting.py`.

knorth55:
>You haven't followed the comments?
May I ask you to tell me why?

Because I found better solution.

>#568 (comment)
I asked roi_seg_ag_locs (in your code roi_ag_loc_scores)

it is nonsense to add `_seg_` in not segmentation variables.

> #568 (comment)
I asked roi_seg_cls_scores (in your code roi_cls_seg_scores)
`h_cls_seg` -> `roi_cls_seg_scores` -> roi_cls_scores`, `roi_seg_scores`
It is easier to understand.

chainercv/experimental/links/model/fcis/fcis.py
yuyu2172: By definition, `rois ` and `bboxes` are different (`bboxes` is a list of `bbox`).
But, as you said, in the case when `batchsize=1`, `rois = bboxes`.

Having said that, I think that continuing to use `rois` in `predict` is better because it explicitly indicates that `roi_mask_prob` is defined relative to `rois`.
This is the same reason why we use `rois` elsewhere, such as `__call__`.

yuyu2172:
> it is nonsense to add _seg_ in not segmentation variables.
This variable has shape `(n_roi, 2 *4, roi_size, roi_size)`, so the shape is mask. That is why adding `_seg_` or `_mask_` is needed. See https://github.com/chainer/chainercv/pull/568#discussion_r183230956

Also, this is not `scores`, but `locs` right?
Or, am I missing something?
https://github.com/knorth55/chainercv/blob/5899fddc6cef8e28aacfd08b8294ccf52f4cf801/chainercv/experimental/links/model/fcis/fcis_resnet101.py#L333

knorth55:
i still cannot understand why we need mask or seg when shape is (R, L, H, W).
in my opinion, shape does not matter because it will be pooled to (R, 2, 4).
seg and mask have different meanings so that i cannot use in this case.
This variable is a kind of scores.
Also, roi_ag_locs is calculated from the variable, so roi_ag_loc_scores sounds nice to me.

chainercv/experimental/links/model/fcis/fcis.py
yuyu2172: Good?

yuyu2172:
> This variable is a kind of scores.

OK. Maybe I am misunderstanding something.
The naming convention inside `_pool` is fine right now.
We can merge this for now because it is only an implementation detail.
I would appreciate if you explain this to me when we meet face to face.

Having said that, what do you think about stop using `_seg_` and always use `_mask_`?

knorth55:
>Having said that, what do you think about stop using _seg_ and always use _mask_? I don't see any difference between the two, and using both of them is confusing.

`_seg_` is multi-class including background and `_mask_` is class-agnostic, background and foreground only.
In FCIS, `_seg_` is quite important and unique point, and it propose to use not `(R, L, H, W)` but `(R, L, 2, H, W)` as segmentation.
In detailed, the paper proposes to predict foreground and background for all semantic labels.
That is why I use both `mask` and `seg`.
Maybe, in some part, my code is not consistent so that you got confused.

knorth55:
>OK. Maybe I am misunderstanding something.

At this point, I want to clarify the difference of `score` and `prob`.
I use `score` as network raw output `prob` as `[0,1]` stochastic values.
Is is correct?
I got confused, and for me, it looks `scores` in `predict` and `score` in `__call__` are different.

chainercv/experimental/links/model/fcis/fcis.py
knorth55: I don't see the difference of `bboxes` and `rois`.
In my understanding, `bboxes` and `rois` have same shape `(R, 4)`.
In FCIS, we don't use `roi_ag_locs` (because it is already applied in `__call__` with `iter2=True`) so that `roi` is same as `bbox`.
For the return output, the `predict` should return `bbox` as output not `roi`, so that I rename here.

knorth55:
And also, another question.
In recent PR, `chainer_experimental` is added in `chainercv`.
`FCIS` should be in `chainer_experimental`, `chainercv_experimental` or `experimental` dir?
Which one do you prefer?

tests/experimental_tests/links_tests/model_tests/fcis_tests/test_fcis_resnet101.py
knorth55: The same test of `FasterRCNNVGG16` and `SSDVGG16` does not have `@attr.disk`.
Is it needed? 
https://github.com/chainer/chainercv/blob/master/tests/links_tests/model_tests/faster_rcnn_tests/test_faster_rcnn_vgg.py#L17
https://github.com/chainer/chainercv/blob/master/tests/links_tests/model_tests/ssd_tests/test_ssd_vgg16.py#L56

tests/experimental_tests/links_tests/model_tests/fcis_tests/test_fcis_resnet101.py
knorth55: ditto

yuyu2172:
> FCIS should be in chainer_experimental, chainercv_experimental or experimental dir?

`experimental`

tests/experimental_tests/links_tests/model_tests/fcis_tests/test_fcis_resnet101.py
yuyu2172: That is because they don't use pretrained models.

yuyu2172:
> That is why I use both (R, 2, H, W) for mask and (R, L, 2, H, W) for seg.

Got it. This sounds convincing.

yuyu2172:
> I use score as network raw output prob as [0,1] stochastic values.

Yes. The range of score is `[-inf, inf]`. Whenever the range of the value is `[0, 1]`, we should always use `*_prob`.

chainercv/experimental/links/model/fcis/fcis.py
yuyu2172: After looking back at implementations of detection, we have no strict rule on when we can remove `roi_`. I understand that it is convenient to use shortend notation (e.g., `roi_prob -> prob`).

My main concern was that subset of variable has `roi_`, because it looks inconsistent (I would look for `roi` that corresponds to `roi_mask_score`).

https://github.com/knorth55/chainercv/blob/da631a105794f50e54e9cf991c788d74ed2b1463/chainercv/experimental/links/model/fcis/fcis.py#L278

To solve this problem, I want the code to transit from stop appending `roi_*` for all variables at same time.
How about using `roi`, `roi_seg_prob` and `roi_prob` until L278, and using `bbox`, `roimask_prob`, `label` and `prob` after L278 (the return vales of `mask_voting`)?
`roimask` is a name for data type of `(R, RH, RW)`.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,350 @@
...
+        # Bbox Regression
+        # shape: (n_roi, 2*4)
+        roi_ag_locs = _global_average_pooling_2d(roi_ag_loc_scores)
+        roi_ag_locs = roi_ag_locs.reshape((-1, 2, 4))
+
+        # Mask Regression
+        # shape: (n_roi, n_class, 2, roi_size, roi_size)
+        max_cls_indices = roi_cls_scores.array.argmax(axis=1)
+        # shape: (n_roi, 2, roi_size, roi_size)
+        roi_seg_scores = roi_cls_seg_scores[
```
yuyu2172: https://github.com/chainer/chainercv/pull/568#issuecomment-383481893

`roi_seg_scores --> roi_mask_scores`?

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,350 @@
...
+                (roi_ag_locs, roi_ag_locs2), axis=0)
+            roi_cls_scores = F.concat(
+                (roi_cls_scores, roi_cls_scores2), axis=0)
+        return roi_seg_scores, roi_ag_locs, roi_cls_scores, rois, roi_indices
+
+    def _pool(
+            self, h_cls_seg, h_ag_loc, rois, roi_indices):
+        # PSROI Pooling
+        # shape: (n_roi, n_class*2, roi_size, roi_size)
+        roi_cls_seg_scores = psroi_pooling_2d(
```
yuyu2172: How about not using `cls_seg` and using only `seg`.

These two seem to have no difference.
https://github.com/chainer/chainercv/pull/568#issuecomment-383481893

chainercv/experimental/links/model/fcis/utils/mask_voting.py
yuyu2172: > I look in some chainercv codes, and score and prob are mixed.
For example, object_detection returns scores, which is probability.
In order to clean up, we first set the rule first.

I missed this comment. sorry.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
knorth55: ok

tests/experimental_tests/links_tests/model_tests/fcis_tests/test_fcis_resnet101.py
yuyu2172: Right. That is a mistake.

chainercv/experimental/links/model/fcis/fcis.py
knorth55: > No. bboxes is a list of (R, 4), whereas rois is (R', 4).

You mean `bbox` is after non maximum suppression and other processing?
In that case, this `rois` are already `bboxes`.

knorth55: `roi_mask` are complete different from other `roi_*`.
What you want is removing other `roi_*` variables. Maybe i understand.
In that case, the rule below sounds good to me
```
roi_mask -> roi_mask (no change)
rois -> bboxes
roi_seg_scores -> seg_scores
roi_scores -> scores
```

chainercv/experimental/links/model/fcis/utils/mask_voting.py
knorth55: I totally agree that `prob` is easier to understand.
But, for API, `score` is consistent to object detection API, so I will use `scores` :(.

chainercv/experimental/links/model/fcis/fcis.py
yuyu2172: What I was saying is this.

```
roi_mask -> roimask
rois -> bboxes
roi_seg_scores -> seg_scores
roi_scores -> scores
```

yuyu2172: > You mean bbox is after non maximum suppression and other processing?
In that case, this rois are already bboxes.

No no. The types are different.

chainercv/experimental/links/model/fcis/utils/mask_voting.py
yuyu2172: You can consistently use `prob` up until the end of `prediction`.
In the end, you do `scores.append(prob)`.

tests/experimental_tests/links_tests/model_tests/fcis_tests/test_fcis_resnet101.py
knorth55: OK, i will update

chainercv/experimental/links/model/fcis/utils/mask_voting.py
knorth55: OK

chainercv/experimental/links/model/fcis/fcis.py
knorth55: > No no. The types are different.

Types are different? Can  you tell me more detailed information?

>roi_mask -> roimask

And I prefer `roi_mask` to `roimask` because it is easy to read.

knorth55:
@yuyu2172 
By the way, do we really need `@attr.disk` for pretrained model?
I read the PR #493 , and it is only added in `Dataset` test not `pretrained model`.

Hakuyume:
@knorth55 I think we don't need `attr.disk`. `attr.disk` is for tests that require a huge disk space.

yuyu2172:
https://github.com/chainer/chainercv/pull/568#discussion_r183338581

> Types are different? Can you tell me more detailed information?

Assume that we have two images, and two sets of bounding boxes are assigned to them.
For the first image, we have `bbox` of shape (R1, 4), and for the second image, we have `bbox` of shape (R2, 4).
`bboxes` would be `[(R1, 4), (R2, 4)]`, but `rois` would be `(R1 + R2, 4)`.

yuyu2172:
```
roi_mask -> roi_mask (no change)
roi_seg_scores -> seg_scores
roi_seg_prob -> seg_prob
roi_mask_scores -> mask_scores
roi_cls_scores -> cls_scores
roi_cls_prob -> cls_prob
```

`roi_*` contained quite important information in the case when the batch size is more than one.
For instance, with two images, `roi_cls_probs` is `(R1 + R2, L, 4)`, but `cls_probs` would be `(2, L, 4)`.
Since `__call__` can take `batchsize > 1`, can you *please* put back `roi_*` back to all variables at least in `__call__`?

For `predict`, we assume the batchsize to be 1, so it is really a matter of taste how we name variables.
Thus, it is ok to leave them as is.

Inside `predict`, I would prefer `roimask` to distinguish from `roi_*`. https://github.com/chainer/chainercv/pull/568#discussion_r183338581

Hakuyume:
> I consider pretrained weight to require a huge disk space (but the notion of huge disk is subjective).

Yes. The definition of `huge` is unclear. I think we should decide proper level considering the capacity of the CI system.

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,350 @@
...
+                Its shape is :math:`(R', 4)`. Given :math:`R_i` proposed
+                RoIs from the :math:`i` th image,
+                :math:`R' = \\sum _{i=1} ^ N R_i`.
+            roi_indices (array): An array containing indices of images to
+                which bounding boxes correspond to. Its shape is :math:`(R',)`.
+            img_size (tuple of int): A tuple containing image size.
+
+        """
+        h = F.relu(self.conv1(x))
+        h_cls_seg = self.cls_seg(h)
```
yuyu2172: `cls_seg --> seg_score` and `h_cls_seg --> h_seg_score`

chainercv/experimental/links/model/fcis/fcis_resnet101.py
```diff
@@ -0,0 +1,350 @@
...
+            roi_mask_scores = F.concat(
+                (roi_mask_scores, roi_mask_scores2), axis=0)
+            ag_locs = F.concat(
+                (ag_locs, ag_locs2), axis=0)
+            cls_scores = F.concat(
+                (cls_scores, cls_scores2), axis=0)
+        return roi_mask_scores, ag_locs, cls_scores, rois, roi_indices
+
+    def _pool(
+            self, h_cls_seg, h_ag_loc, rois, roi_indices):
```
yuyu2172: ditto

knorth55:
>roi_* contained quite important information in the case when the batch size is more than one.
For instance, with two images, roi_cls_probs is (R1 + R2, L, 4), but cls_probs would be (2, L, 4).
Since __call__ can take batchsize > 1, can you please put back roi_* back to all variables at least in __call__?
For predict, we assume the batchsize to be 1, so it is really a matter of taste how we name variables.
Thus, it is ok to leave them as is.
Inside predict, I would prefer roimask to distinguish from roi_*. #568 (comment)

Thinking of training code, we should define the variable name now.
`mask`: whole masks `(R, H, W)`
 <=>
`c_mask`: clipped mask `(R, RH, RW)`
`p_mask`, `pooled_mask`, `clipped_mask`, `roi_mask`, `ins_mask`, `obj_mask`?
I prefer `c_mask`.

yuyu2172:
> predict method is for batch_size=1, so rois = bboxes, which does not matter.

More precisely, `rois = bbox`. (`bboxes = [(R,4)] or (B, R, 4)`, `rois=(R,4)`).
That is why `bbox = rois` is correct, but `bboxes = rois` is not.

yuyu2172:
I see. How about `crop_mask`? The name improves readability.

```
crop_mask:  (R, BH, BW) 
roimask or roi_mask:   (R, RH, RW)
```

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,172 @@
...
+                msk_pb.astype(np.float32)[None],
+                (y_max - y_min, x_max - x_min))
+            msk_m = (msk_pb >= binary_thresh).astype(np.float32)[0]
+            mask[y_min:y_max, x_min:x_max] += msk_m * msk_w
+
+    y_indices, x_indices = np.where(mask >= binary_thresh)
+    if len(y_indices) == 0 or len(x_indices) == 0:
+        return None, None
+    else:
+        y_max = y_indices.max() + 1
```
yuyu2172: `+1` is not necessary

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def _mask_aggregation(
+        bbox, mask_prob, mask_weight,
+        size, binary_thresh
+):
+    assert bbox.shape[0] == len(mask_prob)
+    assert bbox.shape[0] == mask_weight.shape[0]
+
+    mask = np.zeros(size, dtype=np.float32)
```
yuyu2172: Isn't this `msk`?

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,172 @@
...
+    y_indices, x_indices = np.where(mask >= binary_thresh)
+    if len(y_indices) == 0 or len(x_indices) == 0:
+        return None, None
+    else:
+        y_max = y_indices.max() + 1
+        y_min = y_indices.min()
+        x_max = x_indices.max() + 1
+        x_min = x_indices.min()
+
+        c_bbox = np.array(
```
yuyu2172: `*_bb`?

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,172 @@
...
+        v_score_l = []
+
+        for i, bb in enumerate(bbox_l):
+            iou = bbox_iou(bbox, bb[np.newaxis, :])
+            keep_indices = np.where(iou >= mask_merge_thresh)[0]
+            mask_weight = cls_prob[keep_indices, label]
+            mask_weight = mask_weight / mask_weight.sum()
+            mask_prob_i = roi_mask_prob[keep_indices]
+            bbox_i = bbox[keep_indices]
+            c_mask, c_bbox = _mask_aggregation(
```
yuyu2172: ditto

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,172 @@
...
+    predicted as the same object class.
+
+    Here are notations used.
+    * :math:`R'` is the total number of RoIs produced across batches.
+    * :math:`L` is the number of classes excluding the background.
+    * :math:`RH` is the height of pooled image.
+    * :math:`RW` is the height of pooled image.
+
+    Args:
+        roi_mask_prob (array): A mask probability array whose shape is
```
yuyu2172: `roimask_prob`

chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -0,0 +1,172 @@
...
+        * **v_bbox**: Bounding boxes for the merged masks. Its shape is \
+            :math:`(N, 4)`.
+        * **v_label**: Class labels for the merged masks. Its shape is \
+            :math:`(N, )`.
+        * **v_score**: Class probabilities for the merged masks. Its shape \
+            is :math:`(N, )`.
+
+    """
+
+    roi_mask_size = roi_mask_prob.shape[1:]
```
yuyu2172: `roimask_size`

chainercv/experimental/links/model/fcis/utils/mask_voting.py
yuyu2172: Oh, wait. It was ok

yuyu2172:
This PR is pretty big, and implementation is good except for some naming conventions.
Since naming convention of ChainerCV has not been included in the documentation yet, I would like to postpone a discussion regarding names in another PR.
For now, I will merge this PR.

Related:
https://github.com/chainer/chainercv/issues/159

PR #457, opened 2017-10-17
Use SliceableDataset

Hakuyume:
merge after #571

PR #571, opened 2018-04-17
Support keys assign in GetterDataset

Hakuyume:
cherry-picked from #457

### DIFF
--- a/chainercv/chainer_experimental/datasets/sliceable/getter_dataset.py
+++ b/chainercv/chainer_experimental/datasets/sliceable/getter_dataset.py
@@ -1,6 +1,6 @@
-from chainercv.chainer_experimental.datasets.sliceable import SliceableDataset
 from chainercv.chainer_experimental.datasets.sliceable.sliceable_dataset \
     import _as_indices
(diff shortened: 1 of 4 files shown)


chainercv/chainer_experimental/datasets/sliceable/getter_dataset.py
```diff
@@ -49,13 +49,40 @@ class GetterDataset(SliceableDataset):
...
+        if isinstance(keys, (list, tuple)):
+            self._return_tuple = True
+        else:
+            keys, self._return_tuple = (keys,), False
+
+        new_keys = []
+        for key in keys:
+            if isinstance(key, int):
+                key_index = key
+                if key_index >= len(self._keys):
```
yuyu2172: How about 

```
if key_index < 0:
    key_index += len(self._keys)
if key_index >= len(self._keys) or key_index < 0:
    raise IndexError('index {} is out of bounds for keys with size {}'.format(key, len(self._keys))
```

yuyu2172: The error message would be similar to NumPy.

yuyu2172:
LGTM

PR #575, opened 2018-04-18
Support list in sliceable datasets

Hakuyume:


yuyu2172:
https://github.com/Hakuyume/chainercv/blob/c82dd4b5ea4460918d7c6164b9e71a55788ad714/chainercv/chainer_experimental/datasets/sliceable/getter_dataset.py#L69

Please fix the doc.

Hakuyume:
I want to users to use `tuple` officially even if `list` is accepted. This PR is only for careless users.

Hakuyume:
I'm using some `list` styles in #457. Perhaps, it is better to rewrite them with `tuple` style.

yuyu2172:
LGTM

PR #581, opened 2018-04-20
add more badges in README

knorth55:
Adding badges as `chainer-chemistry`
chainercv


yuyu2172:
`chainer/chainer` looks like what you propose. 
LGTM

PR #583, opened 2018-04-24
str --> string

yuyu2172:
This change is for consistency in the documentation.


Hakuyume:
https://github.com/yuyu2172/chainercv/blob/str-string/chainercv/datasets/directory_parsing_label_dataset.py#L20

yuyu2172:
Thank you

PR #584, opened 2018-04-24
Use `prob` more strictly inside SSD and Faster R-CNN

yuyu2172:
We should use `prob` whenever possible.


Hakuyume:
We cannot assume that all models return a value in [0, 1] as a confidential score.

yuyu2172:
None of them return `probs`. The internal naming is changed to be more explicit.

Hakuyume:
> None of them return probs.

Sorry, it's my misunderstanding.

> The internal naming is changed to be more explicit.
Using both `score` and `prob` is confusing. I don't think this modification improves readability.

yuyu2172:
> Using both score and prob is confusing. I don't think this modification improves readability.

I agree. It is not necessary to enforce this rule.

PR #585, opened 2018-04-25
Add naming convention doc

yuyu2172:
Naming convention is important, but it is currently not explicitly documented.
Most of the contents are taken from https://github.com/chainer/chainercv/issues/159

# sample
screenshot from 2018-04-25 14-22-02


yuyu2172:
@Hakuyume 

Please review

PR #586, opened 2018-04-25
Add YOLOv3

Hakuyume:
figure


yuyu2172:
Can you make the error message more user-friendly in the case when `n_fg_class` is not supplied?

Hakuyume:
Are you talking about this message? https://github.com/chainer/chainercv/blob/master/chainercv/links/model/ssd/ssd_vgg16.py#L235

yuyu2172:
No, but I found that there is the same issue for SSD.
We can do this in another PR.

PR #590, opened 2018-05-02
Add YOLO to README

Hakuyume:


yuyu2172:
The visualization result looks crowded. For instance, people in the third row are not visible because of labels.
Perhaps, the size of the canvas need to be adjusted for better visualization.

The original code may have not produced the desirable image.

Hakuyume:
> The visualization result looks crowded. For instance, people in the third row are not visible because of labels.
Perhaps, the size of the canvas need to be adjusted for better visualization.
>
> The original code may have not produced the desirable image.

I will try. Perhaps, we should reduce the number of images.

Hakuyume:
@yuyu2172 I updated the figure.

yuyu2172:
LGTM

PR #597, opened 2018-05-10
Fix hacking version

yuyu2172:
Travis is failing due to the recent change made to `hacking`.
This fix follows Chainer.

https://github.com/chainer/chainer/pull/4731/files


Hakuyume:
Don't we need to modify this line?
https://github.com/chainer/chainercv/blob/master/.travis.yml#L45

Issue #416, opened 2017-09-07
Why can we not load images from VOC retrain (SSD)?


conner-starsky:
Why can we not use the weights learned in the VOC object detection? It seems like this would allow for much better accuracy than using weights that start from ImageNet. Would love to hear if this is a correct line of thought and what other people think.

Hakuyume:
Let me confirm.
You want to create a SSD model that has different number of classes from VOC (more/less than 20) and initialize its weights by VOC trained model as much as possible. Is this right?

I think this feature is helpful and not difficult to implement. However, I'm wondering what is the best API.

Here are some candidates.
1. Do not change the arguments. If the `n_fg_class` is different from the number of class of `pretrained_model`, the initialization of final layers is skipped (initialized by random).
2. Add a new pretrained_model option, `voc0712_without_conf`.
3. Add a boolean argument. If it is `True`, the initialization of final layers is skipped. (`finetune_mode`?)

@conner-starsky @yuyu2172 How do you think it?

yuyu2172:
Actually, I am working on this.
I think that the cleanest API choice is 1.

However, it gets complicated to load a pretrained weight in the current Chainer.
I am planning to add support for this feature after this PR is merged to the main Chainer.
https://github.com/chainer/chainer/pull/3176

Hakuyume:
> I think that the cleanest API choice is 1.

I agree that 1 is most simple. One concern is that users cannot ignore the final layers when the number of classes is 20 but the classes are different from those of VOC. Perhaps, we can change the condition from `the n_fg_class is different from the number of class of pretrained_model`  to `when n_fg_class is specified`?

> However, it gets complicated to load a pretrained weight in the current Chainer.
I am planning to add support for this feature after this PR is merged to the main Chainer.
chainer/chainer#3176

Nice job.

conner-starsky:
I like 1 as well.

conner-starsky:
Currently, I am having a problem with the SSD demo, where the training script just freezes and doesn't seem to do anything (even with a Titan X selected).

Hakuyume:
> Currently, I am having a problem with the SSD demo, where the training script just freezes and doesn't seem to do anything (even with a Titan X selected).

Please try `MultiprocessIterator` -> `SerialIterator`.
https://github.com/chainer/chainercv/blob/master/examples/ssd/train.py#L154

conner-starsky:
It works now, thank you. Still seems a little bit slow at `0.58801 iters/sec`

[...]

PR #601, opened 2018-05-10
Add fine-tuning tutorial

yuyu2172:
Related
https://github.com/chainer/chainercv/issues/416

### DIFF
--- a/docs/source/tutorial/link.rst
+++ b/docs/source/tutorial/link.rst
@@ -57,7 +57,7 @@ Finally, this is a complete example using SSD300.
     # the number of classes in VOC is different from 50
     dst = SSD300(n_fg_class=50)
     # initialized weights
-    dst(np.zeros((1, 3, 300, 300)).astype(np.float32))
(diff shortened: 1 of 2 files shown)


docs/source/tutorial/link.rst
```diff
@@ -0,0 +1,74 @@
...
+.. code-block:: python
+
+    from chainercv.links import SSD300
+    import numpy as np
+
+    src = SSD300(pretrained_model='voc0712')
+    # the number of classes in VOC is different from 50
+    dst = SSD300(n_fg_class=50)
+    # initialized weights
+    dst(np.zeros((1, 3, 300, 300)).astype(np.float32))
```
Hakuyume: How about using `model.insize` instead of `300`?

Hakuyume: You can write `np.zeros((1, 3, model.insize, insize), dtype=np.float32)`.

yuyu2172: sure

PR #604, opened 2018-05-10
Delete unnecessary assignment in ProposalCreator

t2kasa:
This PR is to delete unnecessary assignment to score with slicing. 
Since the score is not used after getting its sorted order in line 132.


yuyu2172:
LGTM.
Thank you!

PR #608, opened 2018-05-22
update fcis variable names

knorth55:
- `(R', 2, L, RH, RW)`: roi_cls_ag_seg
- `(R', L, RH, RW)`: roi_cls_seg
- `(R', 2, RH, RW)`: roi_ag_seg
- `(R', RH, RW)`: roi_seg

### DIFF
--- a/chainercv/experimental/links/model/fcis/utils/mask_voting.py
+++ b/chainercv/experimental/links/model/fcis/utils/mask_voting.py
@@ -6,22 +6,22 @@ from chainercv.utils import non_maximum_suppression
 
 
 def _mask_aggregation(
-        bbox, cmask_prob, cmask_weight,
+        bbox, seg_prob, seg_weight,
         size, binary_thresh
 ):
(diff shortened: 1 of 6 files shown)


chainercv/experimental/links/model/fcis/utils/mask_voting.py
```diff
@@ -40,7 +40,7 @@ def _mask_aggregation(
 
 
 def mask_voting(
-        roi_cmask_prob, bbox, roi_cls_prob, size,
```
yuyu2172: How about `roi_seg_prob` --> `seg_prob` and `roi_cls_prob` -> `cls_prob`.

Also, there is an inconsistency in the documentation and the variable names.

yuyu2172:
Can you rename `test_mask_voting.py` too?

knorth55:
I rename variables in `test/` too.

yuyu2172:
LGTM

PR #615, opened 2018-05-25
fix typo in examples/instance_segmentation/eval_sbd.py

knorth55:


examples/instance_segmentation/eval_sbd.py
```diff
@@ -14,9 +14,9 @@
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        '--model', choices=('fcis_resnet101'),
+        '--model', choices=['fcis_resnet101'],
         default='fcis_resnet101')
-    parser.add_argument('--pretrained_model')
```
yuyu2172: This was correct because it is consistent with other examples.

knorth55: how about changing all others?

examples/instance_segmentation/eval_sbd.py
```diff
@@ -14,9 +14,9 @@
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        '--model', choices=('fcis_resnet101'),
+        '--model', choices=['fcis_resnet101'],
```
Hakuyume: For the consistency with others, `('fcis_resnet101',)`. Actually, both list and tuple are acceptable. I prefer using tuple when mutability is not required.

examples/instance_segmentation/eval_sbd.py
Hakuyume: I agree with @knorth55. `--foo-bar` style is more common.

yuyu2172: OK

examples/instance_segmentation/eval_sbd.py
knorth55: I used to use `tuple` but it shows weird output `  --model {f,c,i,s,_,r,e,s,n,e,t,1,0,1}` as below;
```bash
$ python eval_sbd.py --help
usage: eval_sbd.py [-h] [--model {f,c,i,s,_,r,e,s,n,e,t,1,0,1}]
                   [--pretrained_model PRETRAINED_MODEL]
                   [--iou-thresh IOU_THRESH] [--gpu GPU]

optional arguments:
  -h, --help            show this help message and exit
  --model {f,c,i,s,_,r,e,s,n,e,t,1,0,1}
  --pretrained_model PRETRAINED_MODEL
  --iou-thresh IOU_THRESH
  --gpu GPU
```
When I use `list`, it does not occur.

examples/instance_segmentation/eval_sbd.py
knorth55: move to #616

examples/instance_segmentation/eval_sbd.py
Hakuyume: Perhaps, you forgot `,`.
OK: `('fcis_resnet101',)` (this is a tuple)
NG: `('fcis_resnet101')` (this is a string)

knorth55: I see, I will modify to use `tuple`

PR #616, opened 2018-05-25
args --pretrained_model -> --pretrained-model

knorth55:
From https://github.com/chainer/chainercv/pull/615#discussion_r190784052


Hakuyume:
We also need to fix READMEs.

knorth55:
@Hakuyume Thank you. I updated Readme, too.

PR #621, opened 2018-05-26
Use bias only for He's ResNet50

yuyu2172:
Related: https://github.com/chainer/chainercv/pull/321#issuecomment-392237672


chainercv/links/model/resnet/resnet.py
```diff
@@ -151,7 +151,8 @@ def __init__(self, n_layer,
             conv1_no_bias = True
         elif arch == 'he':
             stride_first = True
-            conv1_no_bias = False
+            # Kaiming He uses bias only for ResNet50
+            conv1_no_bias = n_layer == 50
```
Hakuyume: `not`

PR #631, opened 2018-06-01
Note pretrained models' license

Hakuyume:


yuyu2172:
PSPNet  is missing

Hakuyume:
I'm just looking for the URL of resource of PSPNet.

PR #632, opened 2018-06-01
Add FCIS to README

yuyu2172:


Hakuyume:
How about moving `classification` first?
I mean
Image Classification,
Object Detecion,
Semantic Segmentation,
Instance Segmentation

PR #638, opened 2018-06-06
Support bool in SliceableDataset

### LINKED ISSUE #637: Make SlicableHelper accept boolean array
Now, we need to convert boolean arrays to integer arrays for indexing.

Here is a possible usage I had problem with.
```
import chainercv
import numpy as np                                             
                                                               
dataset= chainercv.datasets.COCOBboxDataset(split='val')       
bboxes = dataset.slice[:, 'bbox']                              
exist_bbox_mask = np.array([len(bbox) > 0 for bbox in bboxes]) 
dataset = dataset.slice[np.where(exist_bbox_mask)[0]]          
```

Ideally

```
import chainercv
import numpy as np                                             
                                                               
dataset= chainercv.datasets.COCOBboxDataset(split='val')       
bboxes = dataset.slice[:, 'bbox']                              
exist_bbox_mask = np.array([len(bbox) > 0 for bbox in bboxes]) 
dataset = dataset.slice[exist_bbox_mask]          
```

Hakuyume:
fix #637

### DIFF
--- a/chainercv/chainer_experimental/datasets/sliceable/getter_dataset.py
+++ b/chainercv/chainer_experimental/datasets/sliceable/getter_dataset.py
@@ -1,5 +1,7 @@
 from chainercv.chainer_experimental.datasets.sliceable.sliceable_dataset \
     import _as_key_indices
+from chainercv.chainer_experimental.datasets.sliceable.sliceable_dataset \
+    import _is_iterable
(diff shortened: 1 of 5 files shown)


yuyu2172:
We should raise error when the sizes of a boolean array and a dataset are different.

yuyu2172:
LGTM

Issue #646, opened 2018-06-10
eval_detection_coco should force areas to be supplied


yuyu2172:
Currently, when `gt_areas` is not supplied, the areas are computed as the area of bounding boxes.
However, the COCO metric uses the area of masks instead for categorizing scale of objects (small, medium , large).
Thus, it seems confusing to allow users to not supply areas.

If users really want object-scale-wise scores for datasets without gt masks, they can substitute areas of bounding boxes by themselves.

PR #648, opened 2018-06-10
Add DetectionCOCOEvaluator

yuyu2172:
~~Merge after #647~~

Currently, the evaluator assumes the dataset to return `areas` and `crowdeds`.
The evaluator can accept wider range of datasets, but I want to discuss #646 first.


chainercv/extensions/evaluator/detection_coco_evaluator.py
```diff
@@ -0,0 +1,176 @@
...
+    default_name = 'validation'
+    priority = chainer.training.PRIORITY_WRITER
+
+    def __init__(
+            self, iterator, target,
+            label_names=None):
+        if not _available:
+            raise ValueError(
+                'Please install pycocotools \n'
+                'pip install -e \'git+https://github.com/pdollar/coco.git'
```
Hakuyume: Why are you using pdollar's repository? https://github.com/cocodataset/cocoapi looks official one. I know `eval_detection_coco` is also using it.

PR #665, opened 2018-07-03
Add COCO Instance segmentation dataset

knorth55:
Move from #547

### DIFF
--- a/chainercv/datasets/coco/coco_instance_segmentation_dataset.py
+++ b/chainercv/datasets/coco/coco_instance_segmentation_dataset.py
@@ -12,9 +12,9 @@ from chainercv import utils
 
 try:
     from pycocotools import mask as coco_mask
-    _availabel = True
+    _available = True
 except ImportError:
-    _availabel = False
+    _available = False
 
 
 class COCOInstanceSegmentationDataset(GetterDataset):
@@ -71,7 +71,7 @@ class COCOInstanceSegmentationDataset(GetterDataset):
             use_crowded=False, return_crowded=False,
             return_area=False
     ):
-        if not _availabel:
+        if not _available:
             raise ValueError(
                 'Please install pycocotools \n'
                 'pip install -e \'git+https://github.com/pdollar/coco.git'
--- a/chainercv/datasets/coco/coco_instance_segmentation_dataset.py
+++ b/chainercv/datasets/coco/coco_instance_segmentation_dataset.py
@@ -1,0 +1,194 @@
+from collections import defaultdict
+import json
+import numpy as np
+import os
+
+import chainer
+
+from chainercv import utils
+
+from chainercv.datasets.coco.coco_utils import get_coco
+
+try:
+    from pycocotools import mask as coco_mask
+    _availabel = True
+except ImportError:
+    _availabel = False
+
+
+class COCOInstanceSegmentationDataset(chainer.dataset.DatasetMixin):
+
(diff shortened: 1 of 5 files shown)


knorth55:
ready for review

chainercv/datasets/coco/coco_instance_segmentation_dataset.py
```diff
@@ -0,0 +1,183 @@
...
+import PIL.ImageDraw
+
+from chainercv.chainer_experimental.datasets.sliceable import GetterDataset
+from chainercv.datasets.coco.coco_utils import get_coco
+from chainercv import utils
+
+
+try:
+    from pycocotools import mask as coco_mask
+    _availabel = True
```
Hakuyume: `available` ?

chainercv/datasets/coco/coco_instance_segmentation_dataset.py
```diff
@@ -0,0 +1,183 @@
...
+
+        self.cat_ids = [cat['id'] for cat in annos['categories']]
+
+        self.id_to_anno = defaultdict(list)
+        for anno in annos['annotations']:
+            self.id_to_anno[anno['image_id']].append(anno)
+
+        self.add_getter('img', self._get_image)
+        self.add_getter(['mask', 'label', 'area', 'crowded'],
+                        self._get_annotations)
```
Hakuyume: If we can retrieve 'label', 'area', 'crowded' without loading 'mask', it is better to separate getter function.

knorth55: Can you tell me why?
`self.use_crowded` is set as `True`, `crowded` is loaded for every annotation.

chainercv/datasets/coco/coco_instance_segmentation_dataset.py
knorth55: Fixed, thank you.

chainercv/datasets/coco/coco_instance_segmentation_dataset.py
Hakuyume: I mean making `get_mask` and `get_label_area_crowded`.

Hakuyume: I know `mask` requires `crowded` in some cases. However, `label` or `area` can be retrieved without loading `mask`.

yuyu2172: So, in `get_mask`, `crowded` is loaded, but not returned?

knorth55: @Hakuyume can you give me more advice? spliting into two function is efficient for data load?

yuyu2172: He is saying that if `_get_annotation` is split into two, we can access `labels` without `masks` using `dataset.slice[:, 'slice']`.
We are assuming that loading `masks` takes substantially longer time than loading `labels`.

Hakuyume: > So, in _get_mask, crowded is loaded, but not returned?

Yes, if use_crowded=False, `_get_mask` calls `_get_label_area_crowded` internally to get `crowded`. If use_crowded=True, `_get_mask` doesn't need to call `_get_label_area_crowded`.

PR #667, opened 2018-07-04
Add Imagenet detection dataset

yuyu2172:
I deleted the test for now because this requires manual downloading of the entire dataset.

The test can be found below.
https://github.com/chainer/chainercv/blob/bde8e36406babd375f2b80aa48e5242118b2dc0f/tests/datasets_tests/imagenet_tests/test_imagenet_det_bbox_dataset.py


yuyu2172:
Please wait until I add tests.

PR #681, opened 2018-08-28
Assert input type for transform functions

knorth55:
Merge after #686 
- add `assert_is_image` in `transforms/image`
- add `assert_is_bbox` in `transforms/bbox`
- refine `transforms_tests/bbox_tests`


knorth55:
Updated to pass with `nosetests -a '!gpu,!slow'`

PR #686, opened 2018-09-08
use kwargs for ResBlock in pspnet

knorth55:
because `ResBlock` has `group` args after stride,
it collapses positional arguments.


yuyu2172:
Thanks!

PR #687, opened 2018-09-08
Fix coco instance dataset test

knorth55:
I haven't tested with `slow` tag, this is my fault :(
I now tested and got errors with `coco_instance_segmentation`.


yuyu2172:
Thanks!

PR #688, opened 2018-09-08
assert flip image ndim == 3

knorth55:


yuyu2172:
Why do we need to add assertion for this function only?

knorth55:
because flip process like `img[:, ::-1, :]` can be done with `ndim==4`.
The other transform function should also check input shape and type, but it is other problems.

yuyu2172:
OK

PR #690, opened 2018-09-11
add rotate transforms for image

knorth55:
add `rotate` for `img`.
`random_rotate` exists but `rotate` does not exists.

### DIFF
--- a/chainercv/transforms/image/rotate.py
+++ b/chainercv/transforms/image/rotate.py
@@ -1,0 +1,37 @@
+import warnings
+
+
+try:
+    import scipy.ndimage
+    _available = True
+except ImportError:
+    _available = False
+
+
+def _check_available():
+    if not _available:
+        warnings.warn(
+            'SciPy is not installed in your environment,'
+            'so rotate cannot be loaded.'
+            'Please install SciPy to load dataset.\n\n'
+            '$ pip install scipy')
+
+
+def rotate(img, angle, expand=True):
+    """Rotate images by degrees.
+
+    Args:
+        img (~numpy.ndarray): An arrays that get flipped. This is in
+            CHW format.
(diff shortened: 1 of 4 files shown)


tests/transforms_tests/image_tests/test_rotate.py
```diff
@@ -0,0 +1,26 @@
...
+from chainercv.transforms import rotate
+
+
+@testing.parameterize(*testing.product({
+    'k': [0, 1, 2, 3]
+}))
+class TestRotate(unittest.TestCase):
+
+    def test_rotate(self):
+        img = np.random.uniform(size=(3, 24, 24))
```
Hakuyume: How about using non-square image for test?

knorth55: updated

yuyu2172:
Could you change argument `k` to `angle`?
If possible, could you support continuous values between `-180,180`?

knorth55:
i think continuous rotate support is different problem.
continuous rotate do not change `img.shape`, intepolate pixel value and do zero-padding in outside area.
this function is `rotate90` introduced in `random_rotate`.

knorth55:
Updated to use `scipy.ndimage.rotate`

chainercv/transforms/image/rotate.py
```diff
@@ -0,0 +1,37 @@
...
+            '$ pip install scipy')
+
+
+def rotate(img, angle, expand=True):
+    """Rotate images by degrees.
+
+    Args:
+        img (~numpy.ndarray): An arrays that get flipped. This is in
+            CHW format.
+        angle (float): Clock-wise rotation angle (degree) in [-180, 180].
```
yuyu2172: https://github.com/chainer/chainercv/pull/692/files#r218293211

PR #693, opened 2018-09-13
deprecate random_rotate

yuyu2172:
This is no longer recommended.
Instead, we recommend to use a random function combined with `rotate` (to be added).


knorth55:
continuous rotate and `rotate90` is different augmentation.
we should rename this to `rotate90` and add `rotate` as continuous rotate.

Issue #139, opened 2017-04-28
Add KITTI to datasets


mitmul:
http://www.cvlibs.net/datasets/kitti/

Sirokujira:
I have a few questions.
· Items to be extracted as data
    (Image data only, label, area as well?)

· Whether external library usage is ok or not.
　(About use of pykitti)

· About description of test code
  (Items for determining implementation completion)
  (If you can confirm that you can retrieve data, you decide that implementation is complete?)

cbd2015:
how to  add myself dataset to  finetune a   new model

PR #697, opened 2018-09-21
Add kitti dataset

Sirokujira:
Issue #139 

use package : 
pykitti package(0.2.4/0.3.0)
parseTrackletXML.py
(Methods for parsing tracklets (e.g. dataset labels), 
 originally created by Christian Herdtweck.)

Function :
Display labels and boxes

not implemented
Grayscale processing


yuyu2172:
Hi! Thank you for your PR! Some comments.

1. Could you change default data directory from `KITTI` to `kitti`? (similarly for COCO, we name the directory as `coco`)
2. Could you raise more friendly error when `date` or `drive_num` are not specified? (`label_resolution` option of `CityscapesSemanticSegmentationDataset` is a good example for situation like this)
3. Could you change the shape of `label` to `(0,)` when there is only one instance?
https://github.com/chainer/chainercv/blob/master/chainercv/utils/testing/assertions/assert_is_bbox_dataset.py#L51
4. Is it possible to drop support for gray scale image? If we drop support, can we use `pykitti==0.3.0` for all settings? I personally prefer to set `0.3.0` as the recommended pykitti version.

yuyu2172:
Do you know any experiments in academic papers that can be supported by this dataset?

Sirokujira:
1. current data folder name Lower case(and download destination folder)

2. add check logic (check date, drivenum)

3. fixed code(label shape 2 dim -> 1 dim)

4. fixed use pykitti >= 0.3.0 (dropped function grayscale image process.)

>Do you know any experiments in academic papers that can be supported by this dataset?

source of a quote?
It is the item "raw dataset" of "Citation" at http://www.cvlibs.net/datasets/kitti/.
(add comment kitti_bbox_dataset.py)

I do not know about the research citing the data set.

chainercv/datasets/kitti/kitti_bbox_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+        :obj:`bbox` [#kitti_bbox_1]_, ":math:`(R, 4)`", :obj:`float32`, \
+        ":math:`(y_{min}, x_{min}, y_{max}, x_{max})`"
+        :obj:`label`, scalar, :obj:`int32`, ":math:`[0, \#class - 1]`"
+
+    .. [#kitti_bbox_1] If :obj:`tracklet = True`, \
+        :obj:`bbox` and :obj:`label` contain crowded instances.
+
+    When queried by an index, if :obj:`tracklet == True`,
+    this dataset returns a corresponding
+    :obj:`img, bbox, label`, a tuple of an image, bounding boxes, labels.
```
yuyu2172: Could you make two classes: `KITTIBboxDataset` and `KITTIImageDataset`?
It is confusing that when `tracklet=False`, the dataset never returns unempty bounding box or label even though it is named as `BboxDataset`.
`KITTIBboxDataset` should only support `2011_09_26`, and `KITTIImageDataset` should support all the other dates.
Also, if you don't want to, you do not need to implement `KITTIImageDataset`.

We have two classes for Cityscapes as well because subset of the dataset does not contain annotation.

chainercv/datasets/__init__.py
```diff
@@ -19,6 +19,13 @@
 from chainercv.datasets.cub.cub_utils import cub_label_names  # NOQA
 from chainercv.datasets.directory_parsing_label_dataset import directory_parsing_label_names  # NOQA
 from chainercv.datasets.directory_parsing_label_dataset import DirectoryParsingLabelDataset  # NOQA
+from chainercv.datasets.kitti.kitti_bbox_dataset import KITTIBboxDataset  # NOQA
+from chainercv.datasets.kitti.kitti_utils import kitti_bbox_label_colors  # NOQA
+from chainercv.datasets.kitti.kitti_utils import kitti_bbox_label_names  # NOQA
+from chainercv.datasets.kitti.kitti_utils import kitti_date_lists  # NOQA
```
yuyu2172: Do we need to expose this object at this namespace?
We only expose `_label_names` and `_colors` here.

chainercv/datasets/__init__.py
```diff
@@ -19,6 +19,13 @@
...
+from chainercv.datasets.kitti.kitti_utils import kitti_date_num_dicts  # NOQA
```
yuyu2172: ditto

PR #699, opened 2018-09-25
Remove angle range in `rotate_bbox` and `rotate`

knorth55:
- remove angle range in `rotate`
- remove angle range in `rotate_bbox`
- fix `test_rotate.py`

### DIFF
--- a/chainercv/transforms/bbox/rotate_bbox.py
+++ b/chainercv/transforms/bbox/rotate_bbox.py
@@ -21,7 +21,6 @@ def rotate_bbox(bbox, angle, size):
         raise ValueError(
             'angle which satisfies angle % 90 == 0 is only supported: {}'
             .format(angle))
-        return bbox
     H, W = size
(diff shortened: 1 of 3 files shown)


chainercv/transforms/bbox/rotate_bbox.py
```diff
@@ -20,22 +19,22 @@ def rotate_bbox(bbox, angle, size):
     """
     if angle % 90 != 0:
         raise ValueError(
-            'angle in [180, 90, 0, -90, -180] is only supported: {}'
+            'angle which satisfies angle % 90 == 0 is only supported: {}'
             .format(angle))
         return bbox
```
yuyu2172: this line is unnecessary

yuyu2172: yes

knorth55: OK

PR #701, opened 2018-09-27
make mask_to_bbox support all False mask 

yuyu2172:


yuyu2172:
I will merge this on behalf of knorth55

PR #703, opened 2018-10-11
Release objects in tests

Hakuyume:
NOTE: Most of changes in this PR can be reverted after we switch from nosetests to pytest.


yuyu2172:
LGTM

PR #709, opened 2018-10-15
Fix installation of pycocotools

Hakuyume:


Hakuyume:
Since the pypi package is not official, I prefer installing from github repository.

PR #716, opened 2018-10-25
Fix failing CUB tests

yuyu2172:


yuyu2172:
Merge after https://github.com/chainer/chainercv/pull/717

chainercv/datasets/cub/cub_point_dataset.py
```diff
@@ -75,7 +77,15 @@ def __init__(self, data_dir='auto', return_bb=False,
...
+            os.path.join(self.data_dir, 'images', self.paths[i]),
+            color=True)
+
         point = np.array(self._point_dict[i], dtype=np.float32)
         mask = np.array(self._mask_dict[i], dtype=np.bool)
-        return point, mask
+
+        _, H, W = img.shape
+        point[:, 0] = np.minimum(point[:, 0], H)
+        point[:, 1] = np.minimum(point[:, 1], W)
```
Hakuyume: Is it OK to clip the coordinates? Isn't it better to set `mask` to `False`?

yuyu2172: Thanks. That is the correct way.

PR #725, opened 2018-11-02
raise error when year=2017 and split in [minival, valminusminival]

knorth55:
for COCOBboxDataset and COCOInstanceSegmentationDataset


chainercv/datasets/coco/coco_bbox_dataset.py
```diff
@@ -71,6 +71,11 @@ class COCOBboxDataset(GetterDataset):
 
     def __init__(self, data_dir='auto', split='train', year='2017',
                  use_crowded=False, return_area=False, return_crowded=False):
+        if year == '2017' and split in ['minival', 'valminusminival']:
+            raise ValueError(
+                'coco2017 dataset does not support given split: {1}'
+                .format(year, split))
```
yuyu2172: Isn't his simpler?
`'coco2017 dataset does not support given split: {}'.format(split)`

PR #728, opened 2018-11-06
Remove version specification of OpenCV

Hakuyume:


yuyu2172:
Context:
`opencv==3.2` required Python 3.4.
This dependency is removed with this change

Hakuyume:
> opencv==3.2 required Python 3.4.
Maybe Python 3.5

PR #729, opened 2018-11-06
use scales (tuple of floats) in region_proposal_network

knorth55:
`x` is `(N, C, H, W)` and `scale` is scalar value.
I modify to use `scales: (N, )`.

### DIFF
--- a/chainercv/links/model/faster_rcnn/faster_rcnn.py
+++ b/chainercv/links/model/faster_rcnn/faster_rcnn.py
@@ -112,7 +112,7 @@ class FasterRCNN(chainer.Chain):
         # Total number of classes including the background.
         return self.head.n_class
 
-    def __call__(self, x, scales):
+    def __call__(self, x, scales=None):
         """Forward Faster R-CNN.
 
         Scaling paramter :obj:`scales` is used by RPN to determine the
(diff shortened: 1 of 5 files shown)


chainercv/links/model/faster_rcnn/faster_rcnn.py
```diff
@@ -112,10 +112,10 @@ def n_class(self):
         # Total number of classes including the background.
         return self.head.n_class
 
-    def __call__(self, x, scale=1.):
+    def __call__(self, x, scales):
```
yuyu2172: The default value should be set in order to prevent this from becoming `no-compat`

chainercv/links/model/faster_rcnn/region_proposal_network.py
```diff
@@ -67,7 +67,7 @@ def __init__(
             self.loc = L.Convolution2D(
                 mid_channels, n_anchor * 4, 1, 1, 0, initialW=initialW)
 
-    def __call__(self, x, img_size, scale=1.):
+    def __call__(self, x, img_size, scales):
```
yuyu2172: ditto

knorth55:
Thanks, updated :)

PR #734, opened 2018-11-18
Return rgb when rgba is loaded

yuyu2172:
Sometimes, `cv2.imread` returns `(H, W, 4)` array (e.g., the image below).
It seems that this behavior is dependent on the build of cv2.
The bug occurred with the `cv2` installed using anaconda, but the same problem did not occur for `cv2` installed with apt.

EDIT:
Alpha channel extracted by `img = cv2.imread(path)[:, :, 3]`
figure_1

RGB extracted by `img = cv2.imread(path)[:, :, :3][:. :. ::-1]`
figure_1-2

PIL
`f = Image.open(path); img = f.convert('RGB');`
figure_1-2


chainercv/utils/image/read_image.py
```diff
@@ -22,6 +22,11 @@ def _read_image_cv2(path, dtype, color):
         # reshape (H, W) -> (1, H, W)
         return img[np.newaxis].astype(dtype)
     else:
+        # alpha channel is inclued
```
Hakuyume: `included`?

chainercv/utils/image/read_image.py
```diff
@@ -22,6 +22,11 @@ def _read_image_cv2(path, dtype, color):
...
+        if img.shape[-1] == 4:
+            bgr = img[:, :, :3]
+            alpha = img[:, :, 3:] / 255
+            img = bgr * alpha + 255 * np.ones_like(bgr) * (1 - alpha)
```
Hakuyume: I think there are three options.
1. just ignore alpha channel
2. blend with black image
3. blend with white image

Why do you choose 3.? Personally, I prefer 1.

yuyu2172: This is because the alpha channel contains meaningful information (for some part, the value is in `(0, 1)`).

Hakuyume: 2. is not acceptable?

yuyu2172: Is 2 better?
I thought 3 was more natural

yuyu2172:
@Hakuyume 
Please review this again.

chainercv/utils/image/read_image.py
```diff
@@ -10,9 +12,35 @@
     _cv2_available = False
 
 
-def _read_image_cv2(path, dtype, color):
+def _handle_four_channel_image(img, alpha):
+    if alpha is None:
+        raise ValueError(
+            'An RGBA image is read by chainercv.utils.read_image, '
+            'but the `alphe` option is not set. Please set the option so that '
```
Hakuyume: `alphe` -> `alpha`

PR #749, opened 2018-12-05
Unify evaluation for semantic segmentation

yuyu2172:
Previously, batch statistics is recalculated inside `eval_camvid.py` for SegNet.
This logic is now incorporated inside the training script.


yuyu2172:
Updated

yuyu2172:
@Hakuyume 
Please merge

PR #758, opened 2019-01-11
Fix docstring of `ProposalTargetCreator`

nai62:
Fixed an easy mistake in the docstring of `ProposalTargetCreator`.


yuyu2172:
Thank you!

PR #762, opened 2019-01-31
Perform batch size check before converting scales

23pointsNorth:
If batch size 2 is used, it fails with
```
scale = np.asscalar(cuda.to_cpu(scale))
  File "...numpy/lib/type_check.py", line 482, in asscalar
    return a.item()
ValueError: can only convert an array of size 1 to a Python scalar
```
Proposed outcome:
By reordering the checks, the `ValueError` would be raised before the conversion, avoiding this error.


yuyu2172:
Thank for sending the PR!

PR #764, opened 2019-02-02
 Fix PickableSequentialChain.copy()

ktns:
On master, it seems that modification on copied object by `PickableSequentialChain.copy()` affects original object, due to shallow-copied `layer_names` and `_pick`.

This PR adds tests about it and fixes it.


yuyu2172:
Could you copy `_return_tuple` too?

ktns:
Because it seems that `_return_tuple` is supposed to be a `bool` value, isn't it better to explicitly sanitize it like `copied._return_tuple = bool(self._return_tuple)` ?

ktns:
Sorry, I've changed my mind. Now I think it's better to copy `_return_tuple` also in case the usage of `_return_tuple` changes in future.

PR #770, opened 2019-02-06
Update to v0.12

Hakuyume:


yuyu2172:
I think we should wait for the "slow" tests to pass before merging this PR.

PR #774, opened 2019-02-10
Fix ConstantStubLink for `chainer==6.0.0b2`

ktns:
In `chainer==6.0.0b2`, `chainer.Chain.to_gpu`/`to_cpu` no longer invokes `to_gpu`/`to_cpu` of child links, but directly enumerates parameters and persistent values and convert arrays accordingly.
This change affects `ConstantStubLink` because its overridden `to_gpu`/`to_cpu` is ignored by parent `Chain`s.

This PR addresses the problem using `add_persistent` to register output values of `ConstantStubLink`.


Hakuyume:
A hotfix is added in #823

PR #776, opened 2019-02-11
set n_gpu * 0.0005 as fcis lr default

knorth55:


examples/fcis/train_sbd_multi.py
```diff
@@ -36,11 +37,10 @@ def main():
                         help='Output directory')
     parser.add_argument('--seed', '-s', type=int, default=0)
     parser.add_argument(
-        '--lr', '-l', type=float, default=0.0005,
-        help='Default value is for 1 GPU.\n'
+        '--lr', '-l', type=float, default=None,
+        help='This value is for 1 GPU.\n'
              'The learning rate should be multiplied by the number of gpu')
```
yuyu2172: With the new implementation, the description is wrong.
When `lr` is specified, the learning rate is not for 1GPU.

examples/fcis/train_coco_multi.py
```diff
@@ -42,9 +42,10 @@ def main():
                         help='Output directory')
     parser.add_argument('--seed', '-s', type=int, default=0)
     parser.add_argument(
-        '--lr', '-l', type=float, default=0.0005,
-        help='Default value is for 1 GPU.\n'
+        '--lr', '-l', type=float, default=None,
+        help='This value is for 1 GPU.\n'
              'The learning rate should be multiplied by the number of gpu')
```
yuyu2172: ditto

examples/fcis/train_sbd_multi.py
knorth55: thx, updated

PR #777, opened 2019-02-12
add sort_by_score option to vis_instance_segmentation

yuyu2172:
It is usually more informative to visualize a highly scored instance in front of less scored ones.
Here is an example (yellow instance in the right corner of the left image is with very low score. Ideally, this should not hide the more highly scored "car" instance, which is in orange).
The left image is with `sort_by_score=False` and the right image is with `sort_by_score=True`.

figure_1-5


yuyu2172:
@knorth55 
What do you think about setting the default value of `sort_by_score` to `True`?

knorth55:
Good!

PR #779, opened 2019-02-12
Prevent mutating input of FeaturePredictor.prepare

yuyu2172:
When `scale_size` is `None`, a bug occurs.


yuyu2172:
When `scale_size` is `None`, a bug occurs.

PR #780, opened 2019-02-12
fix eval_instance_segmentation_coco when `pred_masks` is empty

yuyu2172:
# Why
When `pred_mask.shape == (0, H, W)`, the function fails.

# Note

The test data did not follow the current data format, so I fixed it with the following conversion script.

```python
import numpy as np

data = np.load('eval_instance_segmentation_coco_result_2018_07_06.npz', encoding='latin1')

labels = data['labels']
masks = data['masks']
scores = data['scores']
masks = [np.array(mask) for mask in masks]

np.savez(
    'eval_instance_segmentation_coco_result_2019_02_12.npz',
    labels=labels,
    scores=scores,
    masks=masks)
```


yuyu2172:
In the original data, `masks` are stored as `[[(H, W)]]`.
This contradicts with the format of `masks`, which is `[(R, H, W)]`.

PR #787, opened 2019-02-14
Add COCOKeypointDataset and vis_keypoint_coco

yuyu2172:
Merge after https://github.com/chainer/chainercv/pull/806

# Demo
```python
from chainercv.datasets import COCOKeypointDataset
from chainercv.visualizations import vis_keypoint_coco
data = COCOKeypointDataset(split='val')
img, point, valid = data[10][:3]
vis_keypoint_coco(img, point, valid, markersize=10)
plt.show()

```

# Memo
Point is represented as `[(K, 3)]` in COCO dataset.
The second axis contains `y, x, v`, where `v` is :
```
0: not labeled; 1: labeled, not inside mask;
2: labeled and inside mask
```

The distinction between `1` and `2` are not made in evaluation code.
From what I understand, the distinction is also not made in Detectron.


yuyu2172:
pfnCI, test this please

mitmul:
How's the progress on this PR? Is it ready for merge or the API design might be still changed?

yuyu2172:
This PR has been pending because I cannot allocate time to upload test data to GCP so that CI can test the implementation of this PR.

yuyu2172:
@Hakuyume

PR #798, opened 2019-02-18
fix resize when 0 length image is given

yuyu2172:
This PR fixes an unintended behavior of `resize` when img of lenght 0 is given.
The example that demonstrates the behavior is below.
```python
import numpy as np
import PIL.Image

from chainercv.transforms import resize


x = np.zeros((0, 512, 400), dtype=np.float32)
out = resize(x, (400, 500), interpolation=PIL.Image.NEAREST)
print(out.shape)  # (512, 400, 500)
```

### DIFF
--- a/chainercv/transforms/image/resize.py
+++ b/chainercv/transforms/image/resize.py
@@ -61,6 +61,9 @@ def resize(img, size, interpolation=PIL.Image.BILINEAR):
         ~numpy.ndarray: A resize array in CHW format.
 
     """
+    if len(img) == 0:
+        return np.zeros((0,) + size, dtype=img.dtype)
+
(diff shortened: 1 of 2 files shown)


chainercv/transforms/image/resize.py
```diff
@@ -61,6 +61,10 @@ def resize(img, size, interpolation=PIL.Image.BILINEAR):
         ~numpy.ndarray: A resize array in CHW format.
 
     """
+    if len(img) == 0:
+        assert len(size) == 2
+        return np.zeros((0,) + size, dtype=img.dtype)
```
knorth55: `np.empty` ?

yuyu2172: nice

PR #800, opened 2019-02-19
change dtype before resize in fpn

knorth55:


yuyu2172:
Input image is expected to be `float32`.
I don't think this change necessary.

PR #801, opened 2019-02-19
add sort_by_score in vis_bbox

knorth55:
same as `vis_instance_segmentation`
for instance segmentation visualization, I own use both visualize functions.


yuyu2172:
Please add tests.

PR #808, opened 2019-02-21
change interface of point/transforms

yuyu2172:
Assume `point` as `[(K, 2)]` instead of `(K, 2)`.

PR #809, opened 2019-02-21
Change interface of vis_point

yuyu2172:
Assume `point` as `[(K, 2)]` instead of `(K, 2)`.

### DIFF
--- a/chainercv/visualizations/vis_point.py
+++ b/chainercv/visualizations/vis_point.py
@@ -21,8 +21,8 @@ def vis_point(img, point, visible=None, ax=None):
     Args:
         img (~numpy.ndarray): See the table below.
             If this is :obj:`None`, no image is displayed.
-        point (~numpy.ndarray): See the table below.
-        visible (~numpy.ndarray): See the table below.
+        point (~numpy.ndarray or list): See the table below.
+        visible (~numpy.ndarray or list): See the table below.
         ax (matplotlib.axes.Axes): If provided, plot on this axis.
 
     .. csv-table::
--- a/chainercv/visualizations/vis_point.py
+++ b/chainercv/visualizations/vis_point.py
@@ -19,7 +19,7 @@ def vis_point(img, point, visible=None, ax=None):
         >>> plt.show()
 
     Args:
-        img (~numpy.ndarray): See the table below. 
+        img (~numpy.ndarray): See the table below.
             If this is :obj:`None`, no image is displayed.
         point (~numpy.ndarray): See the table below.
         visible (~numpy.ndarray): See the table below.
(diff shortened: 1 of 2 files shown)


chainercv/visualizations/vis_point.py
```diff
@@ -6,34 +6,33 @@
...
+        point (~numpy.ndarray): See the table below.
+        visible (~numpy.ndarray): See the table below.
         ax (matplotlib.axes.Axes): If provided, plot on this axis.
 
+    .. csv-table::
+        :header: name, shape, dtype, format
+
+        :obj:`img`, ":math:`(3, H, W)`", :obj:`float32`, \
+        "RGB, :math:`[0, 255]`"
+        :obj:`point`, ":math:`[(K, 2)]`", :obj:`float32`, ":math:`(y, x)`"
```
Hakuyume: If we assumes all instances have the same `K`, it should be `(N, K, 2)` (`N` is the number of instances).

Hakuyume: In #808, `R` is used for the number of instances. It's better to make them consistent.

yuyu2172: We don't assume that all instances have the same `K` for this function.
That is why I left the documentation to accept lists.

Hakuyume: So, these two lines should be like `(~numpy.ndarray or list of arrays)`.
https://github.com/chainer/chainercv/blob/07f136e1d797ffe8024e95e9b2c74cb2cb602fe2/chainercv/visualizations/vis_point.py#L24-L25

yuyu2172:
fixed

chainercv/visualizations/vis_point.py
```diff
@@ -6,34 +6,34 @@
...
+        point (~numpy.ndarray or list of arrays): See the table below.
+        visible (~numpy.ndarray or list of arrays): See the table below.
         ax (matplotlib.axes.Axes): If provided, plot on this axis.
 
+    .. csv-table::
+        :header: name, shape, dtype, format
+
+        :obj:`img`, ":math:`(3, H, W)`", :obj:`float32`, \
+        "RGB, :math:`[0, 255]`"
+        :obj:`point`, ":math:`[(K, 2)]` or :math:`(R, K, 2)`", \
```
Hakuyume: https://github.com/chainer/chainercv/pull/808#discussion_r260124931

yuyu2172: Ah. Thanks

chainercv/visualizations/vis_point.py
```diff
@@ -6,34 +6,34 @@
...
         ax (matplotlib.axes.Axes): If provided, plot on this axis.
 
+    .. csv-table::
+        :header: name, shape, dtype, format
+
+        :obj:`img`, ":math:`(3, H, W)`", :obj:`float32`, \
+        "RGB, :math:`[0, 255]`"
+        :obj:`point`, ":math:`(R, K, 2)` or :math:`[(K, 2)]`", \
+        :obj:`float32`, ":math:`(y, x)`"
+        :obj:`visible`, ":math:`[(K,)]` or :math:`(R, K)`", :obj:`bool`, --
```
Hakuyume:
```
":math:`(R, K)` or :math:`[(K,)]`"
```

PR #815, opened 2019-02-25
Set names of output Variables from PickableSequentialChain

ktns:
Set names of output `Variable`s from `PickableSequentialChain`, so that they match with names in `pick` and `layer_names`.

This is for more readable computational graphs.


yuyu2172:
There is a possibility that a user of `PickableSequentialChain` uses a function that outputs a variable with `name`...

yuyu2172:
How about setting the name only when `var.name` is None?

ktns:
If an instance of `PicakbleSequentialChain` has a nested `PickableSequentialChain` as its child, I think its better `var.name` is rewritten by the parent because the parent tends to have more information about the whole structure of the NN. I thought a scheme like `parent_name/child_name` is too complicated, but it might require reconsideration.

PR #816, opened 2019-02-25
fix resume in train.py for ssd

zhangwenwen:


Hakuyume:
Why do you want to change that line?
We should resume `trainer` because it contains the number of proceeded iterations, status of the optimizer and learning rate.

PR #817, opened 2019-02-25
add left time in progress hook

knorth55:


chainercv/utils/iterator/progress_hook.py
```diff
@@ -23,9 +23,12 @@ def __call__(self, in_values, out_values, rest_values):
         self.n_processed += len(in_values[0])
         fps = self.n_processed / (time.time() - self.start)
         if self.n_total is not None:
+            eta = int((self.n_total - self.n_processed) / fps)
             sys.stdout.write(
-                '\r{:d} of {:d} samples, {:.2f} samples/sec'.format(
-                    self.n_processed, self.n_total, fps))
+                '\r{:d} of {:d} samples, {:.2f} samples/sec,'
+                ' {:4d}:{:02d}:{:02d}'.format(
```
Hakuyume: How about `ETA {:4d}:{:02d}:{:02d` ?

yuyu2172:
Can you check the case when `fps=0` (This could technically happen when  `len(in_values[0])`)?

chainercv/utils/iterator/progress_hook.py
knorth55: updated

knorth55:
updated

yuyu2172:
LGTM!

PR #820, opened 2019-03-02
Fix docstring of DirectoryParsingLabelDataset

iwiwi:
This is a small fix to the docstring of class `DirectoryParsingLabelDataset`. I think it does not have property `paths` and `img_paths` is appropriate here.


yuyu2172:
Thank you!

PR #830, opened 2019-03-10
FPN - Add min/max size params for pre-processing.

23pointsNorth:
Allow for the `min_size`, `max_size` parameter to be set when creating an FPN network. 
It is following the Faster-RCNN template.

### DIFF
--- a/chainercv/links/model/fpn/faster_rcnn.py
+++ b/chainercv/links/model/fpn/faster_rcnn.py
@@ -46,7 +46,7 @@ class FasterRCNN(chainer.Chain):
     _stride = 32
 
     def __init__(self, extractor, rpn, head,
-                 min_size, max_size):
+                 min_size=800, max_size=1333):
         super(FasterRCNN, self).__init__()
         with self.init_scope():
             self.extractor = extractor
--- a/chainercv/links/model/fpn/faster_rcnn.py
+++ b/chainercv/links/model/fpn/faster_rcnn.py
@@ -26,6 +26,9 @@ class FasterRCNN(chainer.Chain):
         head (Link): A link that has the same interface as
             :class:`~chainercv.links.model.fpn.Head`.
             Please refer to the documentation found there.
+        min_size (int): A preprocessing paramter for :meth:`prepare`. Please
+            refer to a docstring found for :meth:`prepare`.
+        max_size (int): A preprocessing paramter for :meth:`prepare`.
 
     Parameters:
         nms_thresh (float): The threshold value
@@ -40,16 +43,18 @@ class FasterRCNN(chainer.Chain):
 
     """
 
-    _min_size = 800
-    _max_size = 1333
     _stride = 32
 
-    def __init__(self, extractor, rpn, head):
+    def __init__(self, extractor, rpn, head,
+                 min_size, max_size):
         super(FasterRCNN, self).__init__()
         with self.init_scope():
             self.extractor = extractor
             self.rpn = rpn
             self.head = head
+
+        self._min_size = min_size
+        self._max_size = max_size
 
         self.use_preset('visualize')
 
--- a/chainercv/links/model/fpn/faster_rcnn.py
+++ b/chainercv/links/model/fpn/faster_rcnn.py
@@ -28,7 +28,9 @@ class FasterRCNN(chainer.Chain):
             Please refer to the documentation found there.
         min_size (int): A preprocessing paramter for :meth:`prepare`. Please
             refer to a docstring found for :meth:`prepare`.
-        max_size (int): A preprocessing paramter for :meth:`prepare`.
+        max_size (int): A preprocessing paramter for :meth:`prepare`. Note
+            that the result of :meth:`prepare` can exceed this size due to
+            alignment with stride.
 
     Parameters:
         nms_thresh (float): The threshold value
--- a/chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
+++ b/chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
@@ -60,8 +60,8 @@ class FasterRCNNFPNResNet50(FasterRCNNFPNResNet):
     Args:
         n_fg_class (int): The number of classes excluding the background.
         pretrained_model (string): The weight file to be loaded.
-           This can take :obj:`'coco'`, `filepath` or :obj:`None`.
-           The default value is :obj:`None`.
+            This can take :obj:`'coco'`, `filepath` or :obj:`None`.
+            The default value is :obj:`None`.
 
             * :obj:`'coco'`: Load weights trained on train split of \
                 MS COCO 2017. \
--- a/chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
+++ b/chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
@@ -19,7 +19,8 @@ class FasterRCNNFPNResNet(FasterRCNN):
     A subclass of this class should have :obj:`_base` and :obj:`_models`.
     """
 
-    def __init__(self, n_fg_class=None, pretrained_model=None):
+    def __init__(self, n_fg_class=None, pretrained_model=None,
+                 min_size=800, max_size=1333):
         param, path = utils.prepare_pretrained_model(
             {'n_fg_class': n_fg_class}, pretrained_model, self._models)
 
@@ -56,8 +57,8 @@ class FasterRCNNFPNResNet50(FasterRCNNFPNResNet):
        Feature Pyramid Networks for Object Detection. CVPR 2017
 
     Args:
-       n_fg_class (int): The number of classes excluding the background.
-       pretrained_model (string): The weight file to be loaded.
+        n_fg_class (int): The number of classes excluding the background.
+        pretrained_model (string): The weight file to be loaded.
            This can take :obj:`'coco'`, `filepath` or :obj:`None`.
            The default value is :obj:`None`.
 
@@ -74,6 +75,9 @@ class FasterRCNNFPNResNet50(FasterRCNNFPNResNet):
             * `filepath`: A path of npz file. In this case, :obj:`n_fg_class` \
                 must be specified properly.
             * :obj:`None`: Do not load weights.
+        min_size (int): A preprocessing paramter for :meth:`prepare`. Please
+            refer to a docstring found for :meth:`prepare`.
+        max_size (int): A preprocessing paramter for :meth:`prepare`.
 
     """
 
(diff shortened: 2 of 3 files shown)


yuyu2172:
Thanks. Can you set the default values for `FasterRCNN` so that the tests won't fail?

chainercv/links/model/fpn/faster_rcnn.py
```diff
@@ -26,6 +26,9 @@ class FasterRCNN(chainer.Chain):
         head (Link): A link that has the same interface as
             :class:`~chainercv.links.model.fpn.Head`.
             Please refer to the documentation found there.
+        min_size (int): A preprocessing paramter for :meth:`prepare`. Please
+            refer to a docstring found for :meth:`prepare`.
```
Hakuyume: How about this?
```
Please refer to :meth:`prepare`.
```
`docstring` will be confusing when it rendered by sphinx.

chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
```diff
@@ -56,8 +57,8 @@ class FasterRCNNFPNResNet50(FasterRCNNFPNResNet):
        Feature Pyramid Networks for Object Detection. CVPR 2017
 
     Args:
-       n_fg_class (int): The number of classes excluding the background.
-       pretrained_model (string): The weight file to be loaded.
+        n_fg_class (int): The number of classes excluding the background.
+        pretrained_model (string): The weight file to be loaded.
```
Hakuyume: Why did you change the indentation?

chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
```diff
@@ -99,10 +103,10 @@ class FasterRCNNFPNResNet101(FasterRCNNFPNResNet):
...
 
     Args:
-       n_fg_class (int): The number of classes excluding the background.
-       pretrained_model (string): The weight file to be loaded.
-           This can take :obj:`'coco'`, `filepath` or :obj:`None`.
-           The default value is :obj:`None`.
+        n_fg_class (int): The number of classes excluding the background.
+        pretrained_model (string): The weight file to be loaded.
+            This can take :obj:`'coco'`, `filepath` or :obj:`None`.
+            The default value is :obj:`None`.
```
Hakuyume: ditto.

chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
```diff
@@ -19,7 +19,8 @@ class FasterRCNNFPNResNet(FasterRCNN):
     A subclass of this class should have :obj:`_base` and :obj:`_models`.
     """
 
-    def __init__(self, n_fg_class=None, pretrained_model=None):
+    def __init__(self, n_fg_class=None, pretrained_model=None,
+                 min_size=800, max_size=1333):
```
Hakuyume: `min_size` and `max_size` are not used.

Hakuyume:
> Thanks. Can you set the default values for FasterRCNN so that the tests won't fail?

Perhaps, we can modify the tests to pass those parameters explicitly.  In that case, we can remove the default value from `FasterRCNN`.

Hakuyume:
> Thanks. Can you set the default values for FasterRCNN so that the tests won't fail?

Perhaps, we can modify the tests to pass those parameters explicitly.  In that case, we can remove the default value from `FasterRCNN`.

chainercv/links/model/fpn/faster_rcnn.py
23pointsNorth: This was done in the same style as the faster-rcnn docs with regards to the param.

Hakuyume:
@23pointsNorth Can you add tests that change `min_size` or `max_size`?
https://github.com/chainer/chainercv/blob/master/tests/links_tests/model_tests/fpn_tests/test_faster_rcnn.py#L120-L126

chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
23pointsNorth: The rest of the code is indented with 4 space. 
These lines were with 3. I am happy to revert if comments don't follow this requirement.

chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
23pointsNorth: That was a versioning mistake. It will be added on L39.

23pointsNorth:
I am not sure how you expect the testing to look like. If we inspect the other place `preprare` is used, i.e. https://github.com/chainer/chainercv/blob/master/tests/links_tests/model_tests/faster_rcnn_tests/test_faster_rcnn.py#L81 only the input param is checked against. 

I can alter the tests to include a value for min/max, such that we can remove the default in the class (but I think it may be unnecessary).

chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
Hakuyume: That was my mistake. Thank you for fixing!

chainercv/links/model/fpn/faster_rcnn_fpn_resnet.py
Hakuyume: This was my mistake. Thank you for fixing!

chainercv/links/model/fpn/faster_rcnn.py
Hakuyume: I see.

Hakuyume:
> I am not sure how you expect the testing to look like. If we inspect the other place preprare is used, i.e. https://github.com/chainer/chainercv/blob/master/tests/links_tests/model_tests/faster_rcnn_tests/test_faster_rcnn.py#L81 only the input param is checked against.

The test in `faster_rcnn_tests` checks the effect of `min_size/max_size` because it uses values that are different from the default values (`(200, 400)` vs `(600, 1000)`). On the other hand, the test in `fpn_tests` uses the default value. So the test cannot detect an error even if `FasterRCNN` ignores those parameters. At least we should use different values for the test. It would be better to check both default params and other params (with `testing.parameterize`).

> I can alter the tests to include a value for min/max, such that we can remove the default in the class (but I think it may be unnecessary, Edit based on:

I'm sorry for my confusing comment. It is OK to keep the default values. I just mentioned the side-effect.

23pointsNorth:
@Hakuyume I attempted to add extra tests. I changed them to follow the ones in [test_faster_rcnn](https://github.com/chainer/chainercv/blob/master/tests/links_tests/model_tests/faster_rcnn_tests/test_faster_rcnn.py#L74) in regards to `in_shape`, `expected_shape`. 

The `paramterize` tag becomes a bit clunky, because in our tests we want to make the permutations of 3*`n_fg_class`, 2*(`min_size`, `max_size`, `in_shape`, `expected shape`). I don't know if I can give 2 dict and the testing can do the permutations of them.

tests/links_tests/model_tests/fpn_tests/test_faster_rcnn.py
```diff
@@ -31,24 +31,35 @@ def __call__(self, x):
...
         )
 
 
 @testing.parameterize(
-    {'n_fg_class': 1},
-    {'n_fg_class': 5},
-    {'n_fg_class': 20},
+    {'n_fg_class': 1, 'min_size': 200, 'max_size': 400, 
+     'in_shape': (3, 200, 50), 'expected_shape': (3, 400, 100)},
+    {'n_fg_class': 1, 'min_size': 800, 'max_size': 133,
```
Hakuyume: `133` is typo of `1333`?

tests/links_tests/model_tests/fpn_tests/test_faster_rcnn.py
```diff
@@ -117,13 +128,19 @@ def test_predict_gpu(self):
...
-        x, scales = self.link.prepare(imgs)
-        self.assertEqual(x.shape, (2, 3, 800, 1088))
+    def check_prepare(self):
+        x = _random_array(np, self.in_shape)
+        out, scales = self.link.prepare(x)
+        self.assertIsInstance(out, np.ndarray)
+        self.assertEqual(out.shape, self.expected_shape)
+
+    def test_prepare_cpu(self):
+        self.check_prepare()
```
Hakuyume: We do not need to make `_cpu/_gpu` because `prepare` is not related `cupy`.

Hakuyume:
You can use `testing.product_dict`.

```python
@testing.parameterize(*testing.product_dict(
    [
        {'n_fg_class': 1},
        {'n_fg_class': 5},
        {'n_fg_class': 20},
    ],
    [
        {
            'in_shape': (3, 480, 640),
            'min_size': 800, 'max_size': 1333,
            'expected_shape': (3, 800, 1088),
        },
        {
            'in_shape': (3, 200, 50),
            'min_size': 200, 'max_size': 400,
            'expected_shape': (3, 400, 100),
        },
    ],
))
class TestFasterRCNN(unittest.TestCase):
```

tests/links_tests/model_tests/fpn_tests/test_faster_rcnn.py
```diff
@@ -31,24 +31,35 @@ def __call__(self, x):
...
 
 
 @testing.parameterize(
-    {'n_fg_class': 1},
-    {'n_fg_class': 5},
-    {'n_fg_class': 20},
+    {'n_fg_class': 1, 'min_size': 200, 'max_size': 400, 
+     'in_shape': (3, 200, 50), 'expected_shape': (3, 400, 100)},
+    {'n_fg_class': 1, 'min_size': 800, 'max_size': 133, 
+     'in_shape': (3, 480, 640), 'expected_shape': (3, 800, 1088)},
```
Hakuyume: `prepare` can accept a batch of images of different shapes. 
How about `'in_sizes': [(480, 640), (320, 320)], 'expected_size': (800, 1088)` ? (I omit `3` because it is not parameterized)

tests/links_tests/model_tests/fpn_tests/test_faster_rcnn.py
23pointsNorth: Yes, a typo.

23pointsNorth:
I think I've incorporated all of the requests. A general **unexpected behaviour** I found based on the tests i.e. 
https://travis-ci.org/chainer/chainercv/jobs/505234966#L847
If the `max_size` is the parameter based on which the size is calculated, but it is not divisible by `stride`, based on line 
https://github.com/chainer/chainercv/blob/master/chainercv/links/model/fpn/faster_rcnn.py#L165
we'll get a resulting image size that is larger than `max_size` (i.e. max_size of 400 -> final image size 416).

Hakuyume:
> If the max_size is the parameter based on which the size is calculated, and if it is not divisible by stride, based on line
https://github.com/chainer/chainercv/blob/master/chainercv/links/model/fpn/faster_rcnn.py#L165
we'll get a resulting image size that is larger than max_size (i.e. max_size of 400 -> final image size 416).

Actually, this is an expected behavior (same as the Detecton's implementation).
But I agree with you that it is confusing because the output can exceed `max_size`.

> This I think is ok in general, but should be documented somewhere.

How about adding a note here?
https://github.com/chainer/chainercv/blob/f1ad965c005954c9b19c49c5e5741712b070feb5/chainercv/links/model/fpn/faster_rcnn.py#L31
For example,
```
max_size (int): A preprocessing paramter for :meth:`prepare`. 
        Note that the result of :meth:`prepare` can exceed this size
        because of alignment.
```

PR #831, opened 2019-03-11
Fix ImageNet example

keisukefukuda:
The ImageNet example (`examples/classification/train_imagenet_multi.py`) is broken because it uses `slice` method of `TransformDataset`, which is from chainercv's `chainer_experimental`.

This PR fixed the issue by importing chainercv's TransformDataset.


yuyu2172:
Thank you so much!
Could you fix failing Travis tests?

keisukefukuda:
Travis CI passed!

PR #836, opened 2019-03-17
Add argument of `alpha` to `MixUpSoftLabelDataset`

### LINKED ISSUE #821: Add `alpha` argument to `MixUpSoftLabelDataset`
In the original paper of mixup (https://arxiv.org/pdf/1710.09412.pdf), λ (`mix_ratio` in the code) is sampled from Beta(α, α), where α is a hyperparameter. On the other hand, in ChainerCV's `MixUpSoftLabelDataset`, `mix_ratio` is sampled from uniform distribution (which is the same as α=1 case). It might be more useful if we could specify the value of α. 

https://github.com/chainer/chainercv/blob/6ce2d47f4b1992a0e257392f75312547b268eb5f/chainercv/datasets/mixup_soft_label_dataset.py#L67

crcrpar:
Resolves #821.

### DIFF
--- a/chainercv/datasets/mixup_soft_label_dataset.py
+++ b/chainercv/datasets/mixup_soft_label_dataset.py
@@ -39,6 +39,11 @@ class MixUpSoftLabelDataset(dataset_mixin.DatasetMixin):
 
             The shapes of images and labels should be constant.
         n_class (int): The number of classes in the base dataset.
+        alpha (float): A hyperparameter of Beta distribution.
+            ``mix_ratio`` is sampled from :math:`B(\\alpha,\\alpha)`.
+            The default value is 1.0 meaning that the distribution is
+            equivalent to Uniform distribution with lower boundary of
+            :math:`0` and upper boundary of :math:`1`.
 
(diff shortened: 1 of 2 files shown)


yuyu2172:
Thank you so much.
Could you add test that uses `alpha` option?

chainercv/datasets/mixup_soft_label_dataset.py
```diff
@@ -55,16 +60,21 @@ class MixUpSoftLabelDataset(dataset_mixin.DatasetMixin):
     .. [#mixup_1] Same as :obj:`dataset`.
     """
 
-    def __init__(self, dataset, n_class):
+    def __init__(self, dataset, n_class, alpha=1.0):
+        if alpha <= 0.0:
```
crcrpar: note: NumPy documentation says `alpha` is non-negative but it rejects alpha of 0 raising Value Error of `a <= 0`.

yuyu2172:
LGTM!

yuyu2172:
pfnCI, test this please

PR #844, opened 2019-04-05
Fix SSD train_multi.py

Hakuyume:


Hakuyume:
pfnCI, test this please

PR #853, opened 2019-04-22
Set allow_pickle=True

Hakuyume:
```
Unpickling while loading requires explicit opt-in

The functions np.load, and np.lib.format.read_array take an
allow_pickle keyword which now defaults to False in response to
CVE-2019-6446 <https://nvd.nist.gov/vuln/detail/CVE-2019-6446>_.
```
https://github.com/numpy/numpy/releases/tag/v1.16.3


Hakuyume:
pfnCI, test this please

PR #856, opened 2019-04-26
use pycocotools released in pypi

knorth55:
I switch to use `pycocotools` via pypi.
now `pycocotools` is released via pypi.
https://pypi.org/project/pycocotools/


Hakuyume:
Is `https://pypi.org/project/pycocotools/` the official package? It looks unofficial to me.

knorth55:
probably, you are right

PR #860, opened 2019-05-02
Change the condition in `mask_to_segm()` to eliminate invalid bboxes

### LINKED ISSUE #859: Invalid bboxes can pass a check in mask_to_segm()
During the training of Mask R-CNN model, I sometimes see an error like

```
File "train_multi.py", line 104, in forward
 head_gt_labels, self.model.mask_head.segm_size)
File "/home/shunta/.pyenv/versions/miniconda3-4.3.30/lib/python3.6/site-packages/chainercv-0.12.0-py3.6-linux-x86_64.egg/chainercv/links/model/fpn/mask_head.py", line 219, in mask_head_loss_pre
 mask_to_segm(gt_mask, mask_roi, segm_size, gt_index))
File "/home/shunta/.pyenv/versions/miniconda3-4.3.30/lib/python3.6/site-packages/chainercv-0.12.0-py3.6-linux-x86_64.egg/chainercv/links/model/fpn/mask_utils.py", line 75, in mask_to_segm
 chainer.backends.cuda.to_cpu(mask[i, y_min:y_max, x_min:x_max])
ValueError: could not broadcast input array from shape (224,0) into shape (224,64)
```

and it means either of `y_min >= y_max` or `x_min >= x_max` (or both) is actually happening in that line. These four variables are the elements of `bb` which represents an roi and `bb` is an element of the `bbox` variable, so I modified that part and caught this error and checked the actual value of `bbox`, `cropped_m.shape`, and `mask.shape` right before the error line (L:75), then got

```python
ipdb> bbox
array([[ -6, 520, 178, 810],
       [268, 811, 492, 886],
       [584, 422, 726, 767],
       [-18,  -6, 516, 161],
       [239,   8, 755, 165],
       [-30,  -6, 856, 180],
       [245,  -5, 801, 135],
       [ 31, 424, 202, 803],
       [ 91,  -6, 759, 168],
       [171,  28, 791, 797],
       [ 74, 157, 619, 719],
       [ 43, 128, 791, 666],
       [ -3,  80, 708, 732],
       [-24, 192, 704, 653],
       [ 28, -12, 583, 751],
       [ 12, 119, 745, 910]], 
dtype=int32)
ipdb> cropped_m.shape
(224, 75)
ipdb> mask.shape
(5, 1344, 800)
```

From these values, I'd say `bbox` can have many rois located outside of the mask, and it seems to be unexpected in the current implementation. The current code is trying to skip invalid bboxes by clipping the values first here:
https://github.com/chainer/chainercv/blob/ef784bb8d691a3ed7085da5ed12c9c9e87717b93/chainercv/links/model/fpn/mask_utils.py#L59-L62

and then check the values with this if statement:
https://github.com/chainer/chainercv/blob/ef784bb8d691a3ed7085da5ed12c9c9e87717b93/chainercv/links/model/fpn/mask_utils.py#L75

But, that way to clip values allows `x_min` to be larger than  `x_max` when both `bb[1]` and `bb[3]` are larger than `W` (and same for `y_*`, `bb[0]`, `bb[2]`, and `H`). I'd say the problem is that such a situation can pass the value checks at L75.

So how about changing the condition in L75 to the following?

```python
if y_max <= y_min or x_max <= x_min:
```

mitmul:
Resolves #859


yuyu2172:
pfnCI, test this please

yuyu2172:
LGTM

PR #879, opened 2019-05-20
add test_xception.py

knorth55:
add `test_xception`


yuyu2172:
pfnCI, test this please

PR #880, opened 2019-05-20
[WIP] add coverage test in pfnci

knorth55:
add `coverage` test in pfnci


knorth55:
@Hakuyume we cannot calculate correct coverage because pfnci do `CPU` and `GPU` test separetely.

knorth55:
move to #881

PR #885, opened 2019-05-31
fix doc build

knorth55:
fix doc build
- we need to build `chainerx` because of these lines https://github.com/chainer/chainer/blob/54966403f2d4ceca91c041adc272b7372d773dd1/setup.py#L188-L190
- in order to build `chainerx`, we need `python=3.5` `cmake`
- I switch to use conda to build readthedoc environment
- build succeeded
https://readthedocs.org/projects/knorth55-chainercv/builds/9157251/


knorth55:
this PR is for `0.13.1`

PR #889, opened 2019-06-05
`example/ssd/train_multi.py` MultiboxTrainChain is not set communicator?

higumachan:
I think that it have to set `comm`.


Hakuyume:
pfnCI, test this please

Hakuyume:
Thank you for fixing a bug.

I'm running the training code to confirm it works well. After the training finished, I'll merge this PR.

PR #901, opened 2019-06-25
Update to Chainer 6.1.0

Hakuyume:

PR #902, opened 2019-06-26
Simplify SEBlock by broadcast of binary op

shinh:
instead of explicit broadcast_to. The main motivation of this
change is to simplify the exported ONNX, but this would also
improve performance.


yuyu2172:
pfnCI, test this please

shinh:
Rebased to pull #901 

pfnCI, test this please

PR #904, opened 2019-06-26
[WIP] Support Weight Standardization

yuyu2172:


yuyu2172:
Noticed a more general approach taken to support this in Chainer.

PR #906, opened 2019-06-26
Set allow_pickle = True

Hakuyume:
Same as #853


Hakuyume:
pfnCI, test this please

PR #911, opened 2019-07-16
Fix argument document of dataset selection

take-cheeze:


Hakuyume:
pfnCI, test this please

PR #923, opened 2019-09-02
Add type of args.lr in train_multi.py of PSPNet example

soskek:
The script set `1e-2` as `args.lr` in argparse but didn't set its type (`float`). If we change from command line like `--lr 5e-3`, it is interpreted as `str` and error happens. This fixed it.


Hakuyume:
LGTM

PR #931, opened 2019-10-04
read_image's alpha is 'blend_with_white' as default

okdshin:
For back compatibility, `read_image` 's `alpha` parameter should be 'blend_with_white'.
Current `None` fails classifier train example, because imagenet has images with alpha.


Hakuyume:
I think that issue should be fixed in the ImageNet example.

PR #933, opened 2019-10-08
Accept pathlib.Path type for read_image.

sergeant-wizard:
This patch enables users to pass a `pathlib.Path` object to `read_image`, which is currently not possible due to the limitations `cv2.imread` ( `cv2.imread(pathlib.Path('hoge.png'))` throws a cryptic error )

### DIFF
--- a/chainercv/utils/image/read_image.py
+++ b/chainercv/utils/image/read_image.py
@@ -1,5 +1,7 @@
 from __future__ import division
 
+import pathlib
+
 import chainer
 import numpy as np
 from PIL import Image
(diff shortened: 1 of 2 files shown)


sergeant-wizard:
Looks like we need to wait for the drop of python 2 support :sob:

Hakuyume:
FYI: Our policy about py2 support is [here](https://chainer.org/announcement/2019/08/21/python2.html).

sergeant-wizard:
Will ChainerCV update to Chainer 7 after its stable release?

Hakuyume:
After Chainer v7 becomes stable, we can drop py2 support from ChainerCV.

sergeant-wizard:
Ok the test should pass once PY2 test is removed. I'll rebase once that
happens.

PR #935, opened 2019-10-18
Match the remote version to the local in CI

beam2d:
In the current CI, when testing a pull request to Chainer (CuPy) stable branch, the script installs the master version of CuPy (Chainer). This PR fixes the script to use the stable version of CuPy (Chainer) in such a case. It will hopefully resolve some troubles after dropping Python2 in master branches.

### DIFF
--- a/.pfnci/common.sh
+++ b/.pfnci/common.sh
@@ -12,8 +12,10 @@ gcloud auth configure-docker
 TEMP=$(mktemp -d)
 mount -t tmpfs tmpfs ${TEMP}/ -o size=100%
 
-git merge-base --is-ancestor HEAD v6 && LOCAL_VERSION=stable
-git merge-base --is-ancestor HEAD master && LOCAL_VERSION=master
+function get_local_version() {
+    git merge-base --is-ancestor HEAD v6 && LOCAL_VERSION=stable 
+    git merge-base --is-ancestor HEAD master && LOCAL_VERSION=master
+}
 
 REPOSITORY=${REPOSITORY:-chainercv}
 case ${REPOSITORY} in
@@ -22,12 +24,14 @@ case ${REPOSITORY} in
         cp -a . ${TEMP}/chainercv
         ;;
     chainer)
+        get_local_version
         CHAINER=local
         CUPY=${LOCAL_VERSION}
         cp -a . ${TEMP}/chainer
         mv ${TEMP}/chainer/chainercv/ ${TEMP}/
         ;;
     cupy)
+        get_local_version
         CHAINER=${LOCAL_VERSION}
         CUPY=local
         cp -a . ${TEMP}/cupy
--- a/.pfnci/common.sh
+++ b/.pfnci/common.sh
@@ -13,8 +13,8 @@ TEMP=$(mktemp -d)
 mount -t tmpfs tmpfs ${TEMP}/ -o size=100%
 
 get_local_version() {
-    git merge-base --is-ancestor HEAD v6 && LOCAL_VERSION=stable 
-    git merge-base --is-ancestor HEAD master && LOCAL_VERSION=master
+    git merge-base --is-ancestor v6 HEAD && LOCAL_VERSION=stable 
+    git merge-base --is-ancestor master HEAD && LOCAL_VERSION=master
 }
 
 REPOSITORY=${REPOSITORY:-chainercv}
--- a/.pfnci/common.sh
+++ b/.pfnci/common.sh
@@ -1,8 +1,8 @@
 #! /usr/bin/env sh
 set -eux
 
-STABLE=6.3.0
-LATEST=7.0.0b3
+STABLE=6.4.0
+LATEST=7.0.0b4
 
 systemctl stop docker.service
 mount -t tmpfs tmpfs /var/lib/docker/ -o size=100%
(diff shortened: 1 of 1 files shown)


.pfnci/common.sh
```diff
@@ -1,8 +1,8 @@
 #! /usr/bin/env sh
 set -eux
 
-STABLE=6.3.0
-LATEST=7.0.0b3
+STABLE=6.4.0
+LATEST=7.0.0b4
```
beam2d: I found slightly older versions are used, so I just updated them. I can split them into another PR if requested.

.pfnci/common.sh
```diff
@@ -12,6 +12,9 @@ gcloud auth configure-docker
 TEMP=$(mktemp -d)
 mount -t tmpfs tmpfs ${TEMP}/ -o size=100%
 
+git merge-base --is-ancestor HEAD v6 && LOCAL_VERSION=stable
+git merge-base --is-ancestor HEAD master && LOCAL_VERSION=master
```
Hakuyume: FlexCI does not provide `.git`. So these `git` commands always fail.

beam2d: Hmm, I see. What can be used, then...?

beam2d: @kmaehashi Oh, that's nice! I put the setting into the configs.
@Hakuyume OK, I updated to avoid searching ancestors for chainercv case.

.pfnci/common.sh
```diff
@@ -12,20 +12,27 @@ gcloud auth configure-docker
 TEMP=$(mktemp -d)
 mount -t tmpfs tmpfs ${TEMP}/ -o size=100%
 
+function get_local_version() {
```
kmaehashi:
```suggestion
get_local_version() {
```

kmaehashi: This succeeds with `bash` but fails with `sh`.

beam2d: Oops, thanks for the catch!

.pfnci/common.sh
Hakuyume: Updating ChainerCV's `config.pbtxt` doesn't affect Chainer/CuPy's CI.
Please fix the following files instead.
https://github.com/chainer/chainer/blob/master/.pfnci/config.pbtxt
https://github.com/cupy/cupy/blob/master/.pfnci/config.pbtxt

beam2d: Oh, OK, I'll revert the change in this PR and fix each of Chainer/CuPy's config.

beam2d:
Reverted the change on `gen_config.py` and `config.pbtxt`. I will send PRs to Chainer and CuPy to change configs.

kmaehashi:
https://github.com/cupy/cupy/pull/2577 and https://github.com/chainer/chainer/pull/8331 has merged.

Hakuyume:
> cupy/cupy#2577 and chainer/chainer#8331 has merged.

Those PRs should be back-ported on v6 branches.

kmaehashi:
> Those PRs should be back-ported on v6 branches.
How about using `include_dot_git` in the default config of each job? Currently nothing is configured in v6 branch and using the default config. https://github.com/chainer/chainer/blob/v6/.pfnci/config.pbtxt

Hakuyume:
> How about using include_dot_git in the default config of each job? Currently nothing is configured in v6 branch and using the default config. https://github.com/chainer/chainer/blob/v6/.pfnci/config.pbtxt

That is another option. However, I recommend to update v6's config.pbtxt since it is hard to maintain the default configs.

kmaehashi:
> That is another option. However, I recommend to update v6's config.pbtxt since it is hard to maintain the default configs.

OK, I agree with that idea, will do.

.pfnci/common.sh
```diff
@@ -12,20 +12,27 @@ gcloud auth configure-docker
 TEMP=$(mktemp -d)
 mount -t tmpfs tmpfs ${TEMP}/ -o size=100%
 
+get_local_version() {
+    git merge-base --is-ancestor HEAD v6 && LOCAL_VERSION=stable 
+    git merge-base --is-ancestor HEAD master && LOCAL_VERSION=master
```
kmaehashi:
```suggestion
    git merge-base --is-ancestor master HEAD && LOCAL_VERSION=master
```

.pfnci/common.sh
```diff
@@ -12,20 +12,27 @@ gcloud auth configure-docker
 TEMP=$(mktemp -d)
 mount -t tmpfs tmpfs ${TEMP}/ -o size=100%
 
+get_local_version() {
+    git merge-base --is-ancestor HEAD v6 && LOCAL_VERSION=stable
```
kmaehashi: https://git-scm.com/docs/git-merge-base says:

> --is-ancestor
> Check if the first <commit> is an ancestor of the second <commit>, and exit with status 0 if true, or with status 1 if not. Errors are signaled by a non-zero status that is not 1.


```suggestion
    git merge-base --is-ancestor v6 HEAD && LOCAL_VERSION=stable 
```

kmaehashi:
Backport to v6 branch have been merged (both chainer/cupy).
@beam2d Could you take a look on my comment, or should I takeover?

beam2d:
Sorry for being late, I applied your comments. Thank you!

